PR #4, opened 2014-12-08
New more configurable version of JMX exporter, using simpleclient.

brian-brazil:
It supports:
- Scraping the local jvm
- Scraping a remote jvm
- The existing output format by default (mostly, escaping is a bit safer, and it doesn't assume 'type' is the first beanProperty)
- Customising the name, help and labels as needed
- Exporting only what you want (only whitelist right now, blacklist would be easy to add)
- The java simpleclient

A config of 
  { "hostpost": "localhost:1234",
    "rules": [
      {"^java"}
    ]}
would export everything under java, with the default (i.e. existing) format.

I haven't written tests or the servlet bit yet, but wanted to get initial feedback on the approach.
It also puts out a metric per sample too rather than combining metrics with the same name, which isn't correct.


brian-brazil:
I've changed the scrape time to scrape duration, per discussion on a prometheus server PR.

brian-brazil:
Have you have any further comments on this?

run_sample_httpserver.sh
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 # Script to run a java application for testing jmx4prometheus.
-dctry=$(dirname $0)
-builddir=${dctry}/target/
 
-java -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=5555 -cp ${builddir}/jmx4prometheus-0.2-tests.jar com.prometheus.jmx4prometheus.Main
+java -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=5555 -jar httpserver/target/httpserver-0.3-SNAPSHOT-jar-with-dependencies.jar 5556 sample_config.json
```
typingduck: I notice the exposed port is on the commandline but the target port is in the config. Any reason for the two ways of configuring?

brian-brazil: The webserver is separate from the collector, so they're configured separately. It is possible to use the collector without the webserver (and in places I'll be using that configuration), so it seemed a layering violation to mix them.

README.md
```diff
@@ -1,29 +1,75 @@
...
+     "labels": {},
+     "help": "Cassandra metric $1 $2",
+     "type": "GAUGE",
+     "attrNameSnakeCase": false},
+  ]
+}
+```
+Name     | Description
+---------|------------
+hostPort | The host and port to connect to via remote JMX. If not specified, will talk to the local JVM.
```
typingduck: I guess it'll  be picked up once prometheus has alerting and people setup non-generic alerts, but I remain concerned that configuring typos such as "hostport" will be silently ignored and the operator won't know they're not getting metrics (and if they know, hard to know why).

README.md
```diff
@@ -1,29 +1,75 @@
...
-
-# About
-A prometheus server makes a JSON request to a the jmx4prometheus server which then
-scrapes all beans of a target JMX port on another process. The target can
-either be statically configured in the configuration or passed in by
-prometheus server as a query string parameter.
-Jmx4prometheus can return all mBeans (that are numeric) or be configured with
-whitelist/blacklist behaviour.
+A Collector that can configurably scrape and expose mBeans of a JMX target.
+You can optionally run this as a HTTP server and/or scrape via remote JMX.
```
typingduck: Not sure what this line means as the http server scraps via remote JMX.

README.md
```diff
@@ -1,29 +1,75 @@
...
+```
+{ 
+  "hostPort": "127.0.0.1:1234",
+  "rules": [
+    {"pattern": "^org.apache.cassandra.metrics<type=(\w+), name=(\w+)><>Value:",
+     "name": "cassandra_$1_$2",
+     "labels": {},
+     "help": "Cassandra metric $1 $2",
+     "type": "GAUGE",
+     "attrNameSnakeCase": false},
```
typingduck: This is beautiful! Just tried this on cassandra.

README.md
```diff
@@ -1,29 +1,75 @@
...
 `mvn package` to build.
-See `run_jmx4prometheus_server.sh` for a sample script that runs the webserver.
+See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
+
+## Configuration
+The configuration is in JSON, an example with all possible options:
+```
+{ 
+  "hostPort": "127.0.0.1:1234",
+  "rules": [
```
typingduck: I think a little intro to rules before jumping into the details would help. e.g. they can be used for filtering/renaming variables, what's the reasoning behind them.

README.md
brian-brazil: Yeah, that line is awkward. How's this?

README.md
brian-brazil: Can I borrow your config? It'll save me a small amount of time ;)

I'd like sample configs for as many common Java applications as practical.

README.md
brian-brazil: It's a problem with using unstructured json for configuration. I'd say if it becomes a problem that we add some checks for unknown attributes.

README.md
```diff
@@ -1,29 +1,75 @@
...
+     "help": "Cassandra metric $1 $2",
+     "type": "GAUGE",
+     "attrNameSnakeCase": false},
+  ]
+}
+```
+Name     | Description
+---------|------------
+hostPort | The host and port to connect to via remote JMX. If not specified, will talk to the local JVM.
+rules    | A list of rules to apply in order, processing stops at the first matching rule. Attributes that aren't matched aren't collected. If not specified, defaults to collection everything in the default format.
```
typingduck: _If not specified, defaults to collection everything_

Is it correct that not specifying "rules" is different specifying an empty list?

README.md
```diff
@@ -1,29 +1,75 @@
...
+pattern  | Pattern to match against each bean attribute. The pattern is not anchored. Capture groups can be used in other options. Defaults to matching everything.
```
typingduck: pattern = regex of some kind?

README.md
brian-brazil: This is very much reference documentation. I plan to write a blog post a few days after launch on how to use the JMX Exporter Java Agent with Cassandra that's more a tutorial/codelab, and ultimately move that to be part of the central Prometheus docs.

README.md
brian-brazil: Yes, I figured it'd make your transition easier and it makes the minimal useful config quite small.

README.md
typingduck: It might help if it didn't default to local jvm, but required the operator to explicitly set it to be local.

README.md
brian-brazil: Tweaked

README.md
typingduck: Our config is empty :)

There was an intention to do filtering of the variables but there has not been a need yet.

README.md
```diff
@@ -1,29 +1,75 @@
...
+pattern  | Regex pattern to match against each bean attribute. The pattern is not anchored. Capture groups can be used in other options. Defaults to matching everything.
+attrNameSnakeCase | Converts the attribute name to snake case, this is seen in the named matched by the pattern and the default format. For example, anAttrName to an\_attr\_name.
+name     | The metric name to set. Capture groups from the `pattern` can be used. If not specified, the default format will be used.
+labels   | A map of label name to label value pairs. Capture groups from `pattern` can be used in each. `name` must be set to use this. If not specified and the default format is not being used, no labels are set.
+help     | Help text for the metric. Capture groups from `pattern` can be used. `name` must be set to use this. Defaults to the mBean attribute decription and the full name of the attribute.
+type     | The type of the metric, can be `GAUGE` or `COUNTER`. `name` must be set to use this. Defaults to `GAUGE`.
```
typingduck: We have a lot of devs who don't necessarily know much about prometheus internals, they just one to set one up and point it at their hadoop/cassandra. I think a few of my above comments are along the theme about making it easier for people who just one to get something running to know what to do.

e.g. Words 'labels' and 'COUNTER' vs 'GAUGE' aren't something they'll know straight off so including a blurb or links to external docs would help alot.

httpserver/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.prometheus.jmx</groupId>
+    <artifactId>parent</artifactId>
+    <version>0.3-SNAPSHOT</version>
+  </parent>
+
+  <groupId>io.prometheus.jmx</groupId>
+  <artifactId>httpserver</artifactId>
```
typingduck: Is "httpserver.jar" is a bit too generic?

README.md
brian-brazil: I considered that, having one hostname be 'special' didn't feel right though.

README.md
brian-brazil: This is all reference documentation, I'll have general user documentation up in about 2 weeks with the public launch

httpserver/pom.xml
brian-brazil: What would you suggest? jmx_httpserver?

typingduck: something with prometheus and jmx?

typingduck:
Ran this against a cassandra node which is our worst exporter. Scrape time increased from 2 to 4 seconds (due to the scrape then report) but I guess this it not much of a concern. I was surprised that when I curl that the new 'simple text' is slightly larger than the previous json response (again just an fyi more than a concern)

typingduck:
:+1: Really looking forward to using the renaming ability! Sorry it took so long to get a code review.

brian-brazil:
> new 'simple text' is slightly larger than the previous json response

That's likely due to exporting a metric per timeseries, I went for something simple to start with and will fix it up later.

httpserver/pom.xml
brian-brazil: Changed to jmx_prometheus_httpserver

README.md
```diff
@@ -1,29 +1,75 @@
...
-
-# About
-A prometheus server makes a JSON request to a the jmx4prometheus server which then
-scrapes all beans of a target JMX port on another process. The target can
-either be statically configured in the configuration or passed in by
-prometheus server as a query string parameter.
-Jmx4prometheus can return all mBeans (that are numeric) or be configured with
-whitelist/blacklist behaviour.
+A Collector that can configurably scrape and expose mBeans of a JMX target.
+This can be run as an independant HTTP server, and scrape remote JMX targets.
```
juliusv: "independent"

"HTTP server," -> actually no comma after this one for a change ;) (it's neither a list of three or more items, nor is the thing after the comma a whole sentence by itself)

README.md
```diff
@@ -1,29 +1,75 @@
...
 
 ## Building and Running
 
 `mvn package` to build.
-See `run_jmx4prometheus_server.sh` for a sample script that runs the webserver.
+See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
+
+## Configuration
+The configuration is in JSON, An example with all possible options:
```
juliusv: s/JSON,/JSON./

README.md
juliusv: s/collection/collecting/

README.md
```diff
@@ -1,29 +1,75 @@
...
+     "attrNameSnakeCase": false},
+  ]
+}
+```
+Name     | Description
+---------|------------
+hostPort | The host and port to connect to via remote JMX. If not specified, will talk to the local JVM.
+rules    | A list of rules to apply in order, processing stops at the first matching rule. Attributes that aren't matched aren't collected. If not specified, defaults to collection everything in the default format.
+pattern  | Regex pattern to match against each bean attribute. The pattern is not anchored. Capture groups can be used in other options. Defaults to matching everything.
+attrNameSnakeCase | Converts the attribute name to snake case, this is seen in the named matched by the pattern and the default format. For example, anAttrName to an\_attr\_name.
```
juliusv: s/snake case, this/snake case. This/

juliusv: s/named/name/?

README.md
```diff
@@ -1,29 +1,75 @@
...
+name     | The metric name to set. Capture groups from the `pattern` can be used. If not specified, the default format will be used.
+labels   | A map of label name to label value pairs. Capture groups from `pattern` can be used in each. `name` must be set to use this. If not specified and the default format is not being used, no labels are set.
+help     | Help text for the metric. Capture groups from `pattern` can be used. `name` must be set to use this. Defaults to the mBean attribute decription and the full name of the attribute.
+type     | The type of the metric, can be `GAUGE` or `COUNTER`. `name` must be set to use this. Defaults to `GAUGE`.
+
+Metric names and label names are sanatized. All characters other than `[a-zA-Z0-9:_]` are replaced with underscores,
```
juliusv: s/sanatized/sanitized/

README.md
```diff
@@ -1,29 +1,75 @@
...
+
+### Pattern input
+The format of the input matches against the pattern is
+```
+domain<beanpropertyName1=beanPropertyValue1, beanpropertyName2=beanPropertyValue2, ...><key1, key2, ...>attrName: value
+```
+
+Part     | Description
+---------|------------
+domain   | Bean name, this is the part before the colon in the JMX object name.
```
juliusv: , -> .

README.md
```diff
@@ -1,29 +1,75 @@
...
+beanProperyName/Value | Bean properties, this are the key/values after the colon in the JMX object name.
```
juliusv: dito

juliusv: s/this/these/

juliusv:
I just did an extra quick scan for any obvious spelling/grammar/typo issues. I'll trust Conor's :+1:  on the content review.

brian-brazil:
Fixed all those, thanks.

PR #6, opened 2015-01-15
Add Java Agent, to allow running inside the jvm and exposing standard exports.

brian-brazil:
This depends on #4


johnkilf:
LGTM

PR #9, opened 2015-01-22
License cleanup

beorn7:
@brian-brazil


brian-brazil:
:+1:

Due to the license change, we should get an explicit ack from @typingduck

beorn7:
@typingduck Heads-up for you about the license change.

(I guess this was done on work time, so SoundCloud policies apply and this actually _has_ to use the approved license... ;-> )

typingduck:
"explicit ack"  :+1:

PR #12, opened 2015-02-09
Add options for full JMX URL, Username and Password

cetra3:
This pull request includes a few more optional configuration items to be able to properly set the JMX connection URL and also specify a username & password.  This should be compatible with existing configurations, and if anyone is using the `JmxScraper` class directly, this should also work in the exact same way.

The 3 new options are:
  * **hostUrl**: specify the full JMX URL.  If this isn't defined, it'll create this from hostPort, or set it to nothing if that's not defined (current behaviour).
  * **username**: specify the username when connecting to JMX  Both username and password have to be set for this to be used.
  * **password**: specify the password when connecting to JMX.  Both username and password have to be set for this to be used.

`JmxScraper` has a new overloaded constructor for the `env` variable, which defaults to `null` if username or password is not set (current behaviour).

We also check to see whether hostUrl is in `host:port` form or full URL form by checking for the presence of a`/` character, which is not a valid character for either the hostname or the port number.

The net effect of this change is if you're using either of these classes, their behaviour should not change, unless you're using the 3 above options.

Also added the full stack trace to the logger, which prints out a bit more useful information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,13 @@ java -javaagent:target/jmx_prometheus_javaagent-0.3-SNAPSHOT.jar=1234:config.jso
 
 ## Configuration
 The configuration is in JSON. An example with all possible options:
+
 ```
 { 
   "hostPort": "127.0.0.1:1234",
+  "hostUrl": "service:jmx:rmi:///jndi/rmi://hostname:port/jmxrmi"
+  "username" : "admin"
+  "password" : "admin"
   "lowercaseOutputName": false,
   "lowercaseOutputLabelNames": false,
   "rules": [
@@ -35,9 +39,13 @@ The configuration is in JSON. An example with all possible options:
   ]
 }
 ```
+
 Name     | Description
 ---------|------------
 hostPort | The host and port to connect to via remote JMX. If not specified, will talk to the local JVM.
+hostUrl | The full JMX URL.  If not specified, will use hostPort.
+username | The JMX Connection username.  If not specified, no username/password is used.
+password | The JMX Connection password.  If not specified, no username/passowrd is used.
 lowercaseOutputName | Lowercase the output metric name. Applies to default format and `name`. Defaults to false.
 lowercaseOutputLabelNames | Lowercase the output metric label names. Applies to default format and `labels`. Defaults to false.
 rules    | A list of rules to apply in order, processing stops at the first matching rule. Attributes that aren't matched aren't collected. If not specified, defaults to collecting everything in the default format.
@@ -58,6 +66,7 @@ Example configurations for javaagents can be found at  https://github.com/promet
 
 ### Pattern input
 The format of the input matches against the pattern is
+
 ```
 domain<beanpropertyName1=beanPropertyValue1, beanpropertyName2=beanPropertyValue2, ...><key1, key2, ...>attrName: value
 ```
@@ -75,6 +84,7 @@ The default help includes this string, except for the value.
 
 ### Default format
 The default format will transform beans in a way that should produce sane metrics in most cases. It is
+
 ```
 domain_beanPropertyValue1_key1_key2_...keyN_attrName{beanpropertyName2="beanPropertyValue2", ...}: value
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,9 +20,13 @@ java -javaagent:target/jmx_prometheus_javaagent-0.3-SNAPSHOT.jar=1234:config.jso
 
 ## Configuration
 The configuration is in JSON. An example with all possible options:
+
 ```
 { 
   "hostPort": "127.0.0.1:1234",
+  "hostUrl": "service:jmx:rmi:///jndi/rmi://hostname:port/jmxrmi"
```
brian-brazil: Providing both hostUrl and hostPort should be an error

README.md
```diff
@@ -35,9 +39,13 @@ The configuration is in JSON. An example with all possible options:
   ]
 }
 ```
+
 Name     | Description
 ---------|------------
 hostPort | The host and port to connect to via remote JMX. If not specified, will talk to the local JVM.
+hostUrl | The full JMX URL.  If not specified, will use hostPort.
```
brian-brazil: We should probably split out a separate "Connection" section, as this is a bit hard to understand now.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -59,6 +58,26 @@ private JmxCollector(JSONObject config) throws ParseException {
...
+        } else {
+          //Use hostPort variable if it's set.
+          if(!hostPort.isEmpty()) {
+              hostUrl = "service:jmx:rmi:///jndi/rmi://" + hostPort + "/jmxrmi";
+          } else {
+              hostUrl = "";
+          }
+        }
+
+        if(config.containsKey("username") && config.containsKey("password")) {
```
brian-brazil: Could someone just have a username?

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -59,6 +58,26 @@ private JmxCollector(JSONObject config) throws ParseException {
           hostPort = "";
         }
 
+        if(config.containsKey("hostUrl")) {
+          hostUrl = (String)config.get("hostUrl");
+        } else {
+          //Use hostPort variable if it's set.
```
brian-brazil: Space after //

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -50,11 +54,12 @@ public JmxScraper(String hostPort, MBeanReceiver receiver) {
...
         MBeanServerConnection beanConn;
         JMXConnector jmxc = null;
-        if (hostPort.isEmpty()) {
+        if (hostUrl.isEmpty()) {
           beanConn = ManagementFactory.getPlatformMBeanServer();
         } else {
-          String url = "service:jmx:rmi:///jndi/rmi://" + hostPort + "/jmxrmi";
-          jmxc = JMXConnectorFactory.connect(new JMXServiceURL(url), null);
+          //Accepts either the full URL or just host:port
+          String url = hostUrl.contains("/") ? hostUrl : "service:jmx:rmi:///jndi/rmi://" + hostUrl + "/jmxrmi";
```
brian-brazil: Let's keep this logic in one place, so do this up in the Collector.

brian-brazil:
Also, you should add yourself to the AUTHORS.md.

cetra3:
Should I close this commit and open a new one with your suggested changes? Or did you want me to append this one.

brian-brazil:
You can add to this one.

juliusv:
@cetra3 Any updates / status here?

cetra3:
@juliusv Not currently. I haven't had a spare moment to make those changes as suggested.

fsimmend:
Normal jmx remote urls look like this 
`service:jmx:rmi:/<HOST>:<RMI_SERVER_PORT>/jndi/rmi://<HOST>:<RMI_REGISTRY_PORT>/jmxrmi`
Only if its a JVM >=7 you can prevent both ports are different. Even though tomcat throws a warning if both ports are identical, but thats a bug of tomcat. 

In case RMI is not used and for example mx4j is used the url might look completely different.

I would recommend a single property for the url.

brian-brazil:
> I would recommend a single property for the url.

@fsimmend Sounds good to me.

cetra3:
The only consideration is backwards compatibility. If we remove the other options this will break existing configurations. Is there a way of maybe flagging them as deprecated?

brian-brazil:
We have version releases now, so the release notes should be sufficient.

brian-brazil:
Superseded by #54.

PR #13, opened 2015-02-09
Change output filename on java agent to prevent double shading of the jar

cetra3:
This is a strange one and I'm not sure whether it's my version of maven.

If you've already run `mvn package` at least once, then subsequent `mvn package` commands without `clean` causes the Maven Shade plugin to use the combined jar file, rather than the original.
I.e, It doubles up the shading in the jar which causes some weird issues.  For instance, running on tomcat, I get this error message with the bad jar:

```
Exception in thread "main" java.lang.ClassNotFoundException: io.prometheus.jmx.shaded.io.prometheus.jmx.JavaAgent
```

`mvn clean package` works fine.

This fix also changes the output file to be more in-line with the http server project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ See `./run_sample_httpserver.sh` for a sample script that runs the httpserver ag
 To run as a javaagent:
 
 ```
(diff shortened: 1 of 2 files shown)


brian-brazil:
I've noticed that, thanks for the fix!

PR #15, opened 2015-03-04
use daemon threads for jetty

jimlar:
This makes sure Jetty uses daemon threads for clean exit when the JVM stops. It prevents Cassandra hanging on startup when using the agent with the standard Cassandra start scripts. For more background see https://issues.apache.org/jira/browse/CASSANDRA-7254

Before this PR running "java -javaagent:jmx_prometheus_javaagent-0.3-SNAPSHOT.jar=7070:jmx_exporter.json" results in the JVM showing usage text and then not exiting properly. After PR it exits cleanly. (and this is exactly what the cassandra start scripts do)


brian-brazil:
Thanks!

PR #16, opened 2015-03-11
add some fixes after test jmx_exporter with tomcat 8

rossbachp:
Signed-off-by: Peter Rossbach <peterrossbach@me.com>

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -247,7 +247,7 @@ public class JmxCollector extends Collector {
             }
           }
           // Add to samples.
-          LOGGER.fine("add metric sample: " + name + " " + labelNames + " "+labelValues + " " + ((Number)value).doubleValue());
+          LOGGER.fine("add metric sample: " + name + " " + labelNames + " " + labelValues + " " + ((Number)value).doubleValue());
           addSample(new MetricFamilySamples.Sample(name, labelNames, labelValues, ((Number)value).doubleValue()),
               rule.type, help);
           return;
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -92,7 +92,7 @@ public class JmxCollector extends Collector {
               rule.labelNames = new ArrayList<String>();
               rule.labelValues = new ArrayList<String>();
               for (Map.Entry<String, Object> entry : (Set<Map.Entry<String, Object>>)labels.entrySet()) {
-                rule.labelNames.add(entry.getKey());                                
+                rule.labelNames.add(entry.getKey());
                 rule.labelValues.add((String)entry.getValue());
               }
             }
@@ -111,7 +111,7 @@ public class JmxCollector extends Collector {
         }
 
     }
-    
+
     class Receiver implements JmxScraper.MBeanReceiver {
       Map<String, MetricFamilySamples> metricFamilySamplesMap =
         new HashMap<String, MetricFamilySamples>();
@@ -247,6 +247,7 @@ public class JmxCollector extends Collector {
             }
           }
           // Add to samples.
+          LOGGER.fine("add metric sample: " + name + " " + labelNames + " "+labelValues + " " + ((Number)value).doubleValue());
           addSample(new MetricFamilySamples.Sample(name, labelNames, labelValues, ((Number)value).doubleValue()),
               rule.type, help);
           return;
@@ -296,4 +297,3 @@ public class JmxCollector extends Collector {
       }
     }
 }
-
--- a/example_configs/tomcat.json
+++ b/example_configs/tomcat.json
@@ -8,7 +8,7 @@
       "labels": {
         "protocol":"$1",
         "port":"$2" },
-      "help": "tomcat global $3",
+      "help": "Tomcat global $3",
       "type": "COUNTER",
     },
     {
@@ -17,7 +17,7 @@
       "labels": {
         "module":"$1",
         "servlet":"$2" },
-      "help": "tomcat servlet $3 total",
+      "help": "Tomcat servlet $3 total",
       "type": "COUNTER",
     },
     {
@@ -26,7 +26,7 @@
       "labels": {
         "protocol":"$1",
         "port":"$2" },
-      "help": "tomcat threadpool $3",
+      "help": "Tomcat threadpool $3",
       "type": "GAUGE",
     },
     {
@@ -35,7 +35,7 @@
       "labels": {
         "host":"$1",
         "context":"$2" },
-      "help": "tomcat session $3 total",
+      "help": "Tomcat session $3 total",
       "type": "COUNTER",
     },
   ]
--- a/example_configs/tomcat.json
+++ b/example_configs/tomcat.json
@@ -1,4 +1,6 @@
 {
+  "lowercaseOutputName": true,
+  "lowercaseOutputLabelNames": true,
   "rules": [
     {
       "pattern": "Catalina<type=GlobalRequestProcessor, name=\"(\w+-\w+)-(\d+)\"><>(\w+):",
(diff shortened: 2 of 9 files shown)


example_configs/tomcat.json
```diff
@@ -0,0 +1,34 @@
+{
+  "rules": [
+    {
+      "pattern": "Catalina<type=GlobalRequestProcessor, name=\"(\w+-\w+)-(\d+)\"><>requestCount:",
+      "name": "tomcat_request_count",
```
brian-brazil: Do you have an example /metrics that this produces?

This should probably be tomcat_requests_total

example_configs/tomcat.json
```diff
@@ -0,0 +1,34 @@
...
+      "name": "tomcat_request_send_bytes",
+      "labels": {
+        "impl":"$1",
+        "port":"$2" },
+      "help": "Tomcat request send bytes",
+      "type": "COUNTER",
+      "attrNameSnakeCase": false
+    },
+    {
+      "pattern": "Catalina<j2eeType=Servlet, WebModule=//([-a-zA-Z0-9+&@#/%?=~_|!:.,;]*[-a-zA-Z0-9+&@#/%=~_|]), name=([-a-zA-Z0-9+/$%~_-|!.]*), J2EEApplication=none, J2EEServer=none><>requestCount:",
```
brian-brazil: I'm not very familiar with tomcat, how does this differ from the first item?

example_configs/tomcat.json
```diff
@@ -0,0 +1,34 @@
...
+  "rules": [
+    {
+      "pattern": "Catalina<type=GlobalRequestProcessor, name=\"(\w+-\w+)-(\d+)\"><>requestCount:",
+      "name": "tomcat_request_count",
+      "labels": {
+        "impl":"$1",
+        "port":"$2" },
+      "help": "Tomcat request count",
+      "type": "COUNTER",
+      "attrNameSnakeCase": false
```
brian-brazil: This is the default, no need to mention it.

example_configs/tomcat.json
```diff
@@ -0,0 +1,34 @@
+{
+  "rules": [
+    {
+      "pattern": "Catalina<type=GlobalRequestProcessor, name=\"(\w+-\w+)-(\d+)\"><>requestCount:",
```
brian-brazil: This will only produce 3 metrics, I presume tomcat has far more of interest?


You can do something like:
```
 {
      "pattern": "Catalina<type=GlobalRequestProcessor, name=\"(\w+-\w+)-(\d+)\"><>(\w+):",
      "name": "tomcat_$3_total",
      "labels": {
        "impl":"$1",
        "port":"$2" },
      "help": "Tomcat global $3",
      "type": "COUNTER",
      "attrNameSnakeCase": true
    },
```
to pull out all the metrics

rossbachp: Nice, works great...

brian-brazil:
I presume you've already seen http://www.boxever.com/cassandra-consoles-with-jmx-and-prometheus

Getting good useful metrics out of an existing system is unfortunately a bit of a dark art, with tradeoffs between perfect metrics and having to maintain extensive configs.

example_configs/tomcat.json
rossbachp: Yes, I have made a setup with a working prom tomcat 8 example:) I have implemented a ServletContextListener to start the jmxCollector and other java default registries.

example_configs/tomcat.json
rossbachp: I fix that in some moments:)

example_configs/tomcat.json
brian-brazil: If the java agent isn't practical for TomCat, then it'd be good to get that code into an official prometheus repo somewhere.

rossbachp:
If we use Java 7 we can use catch (javax.management.RuntimeMBeanException | javax.management.RuntimeErrorException | javax.management.RuntimeOperationsException | javax.management.AttributeNotFoundException e) {
                logScrape(mbeanName, attr, "Fail: " + e);
                continue;
}

:)

rossbachp:
> Brian: If the java agent isn't practical for TomCat, then it'd be good to get that code into an official prometheus repo somewhere.

peter: No, it exist some classloader issues with bundled jetty container. I am Tomcat committer and I don't want use two servlet engine products at same process...

brian-brazil:
> If we use Java 7

We're limited to Java 6 I'm afraid.

brian-brazil:
> No, it exist some classloader issues with bundled jetty container. I am Tomcat committer and I don't want use two servlet engine products at same process...

Hmm, what are those issues? It's all shaded so it shouldn't get in the way.

example_configs/tomcat.json
```diff
@@ -0,0 +1,44 @@
+{
+  "rules": [
+    {
+      "pattern": "Catalina<type=GlobalRequestProcessor, name=\"(\w+-\w+)-(\d+)\"><>(\w+):",
+      "name": "tomcat_$3_total",
+      "labels": {
+        "protocol":"$1",
+        "port":"$2" },
+      "help": "Tomcat global $3",
+      "type": "COUNTER",
```
brian-brazil: I take it that all of these are counters?

Where there are a mix, I've left things as all Gauge.

example_configs/tomcat.json
```diff
@@ -0,0 +1,44 @@
...
+      "attrNameSnakeCase": false
```
brian-brazil: You'll want to use one of the options to lower case everything, as snake case is the standard Prometheus way to do things

rossbachp:
After I install my exception fix the agent also works fine, with the tomcat.json example config :)

example_configs/tomcat.json
rossbachp: All attributes of GlobalRequestProcessor can only increase. The only decrease after reboot the system, overflow or you manually reset the counters.

example_configs/tomcat.json
rossbachp: I fix that:) Thank you.

example_configs/tomcat.json
brian-brazil: Sounds good. Things are rarely that clean :)

rossbachp:
I think we need a jmx_exporter webapp to easy deployment at every standard web container. We can use servlet api 2.5. My test webapp works fine.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -247,6 +247,7 @@ public void recordBean(
             }
           }
           // Add to samples.
+          LOGGER.fine("add metric sample: " + name + " " + labelNames + " "+labelValues + " " + ((Number)value).doubleValue());
```
brian-brazil: spacing around + isn't consistent here

brian-brazil:
Okay, I think the only thing left is to ensure all the metric names come out as lowercase. You can use attrNameSnakeCase or lowercaseOutputName depending on what you think is more readable.

example_configs/tomcat.json
```diff
@@ -0,0 +1,42 @@
...
+  "lowercaseOutputName": true,
+  "lowercaseOutputLabelNames": true,
+  "rules": [
+    {
+      "pattern": "Catalina<type=GlobalRequestProcessor, name=\"(\w+-\w+)-(\d+)\"><>(\w+):",
+      "name": "tomcat_$3_total",
+      "labels": {
+        "protocol":"$1",
+        "port":"$2" },
+      "help": "tomcat global $3",
```
juliusv: Ah, I think @brian-brazil didn't mean to also lower-case all help strings, only metric names (anything in the "name" field).

juliusv:
:+1: besides the last small comment about the lower-casing miscommunication. Thanks a lot Peter!

juliusv:
@rossbachp Oh, you sent an update! @brian-brazil does everything look good to you now?

rossbachp:
Many Thanks!

juliusv:
Likewise!

Issue #17, opened 2015-04-17
UnmarshalException when connecting to Bamboo


levilovelock:
When I tried to set up the sample server to hook into my Bamboo server, I get the following:

```SEVERE: JMX scrape failed: java.rmi.UnmarshalException: error unmarshalling return; nested exception is:
	java.io.WriteAbortedException: writing aborted; java.io.NotSerializableException: org.apache.catalina.loader.WebappLoader```

It seems like it connects fine and when I hit the /metrics endpoint I do see a list of metrics, but it's definitely truncated, presumably at the point of the unmarshalling exception.

I can connect to the bamboo server via jconsole no problem.. I tried to set up some rules to ignore that bean, but it seems like the rules are applied after the unmarshalling stage anyway. Any ideas of what I can or should do here?

**Comment by brian-brazil** [comment]
Hmm, this is likely a bug on the bamboo side. We should be resilient to such things, can you try adding the UnmarshalException to the list in doScrape()? If you turn on FINE logging, then we can figure out which bean and attribute is the problem.

**Comment by levilovelock** [comment]
Okay. I didn't know how to enable the FINE logging, so I just changed to log SEVERE. The output looked like this:

```bash
Apr 17, 2015 6:06:12 AM io.prometheus.jmx.JmxScraper logScrape
SEVERE: scrape: 'Catalina{j2eeType=WebModule, name=//localhost/, J2EEApplication=none, J2EEServer=none}aliases': null
java.rmi.UnmarshalException: error unmarshalling return; nested exception is:
...
Apr 17, 2015 6:06:16 AM io.prometheus.jmx.JmxScraper logScrape
SEVERE: scrape: 'Catalina:j2eeType=Filter,name=accessLogFilter,WebModule=//localhost/,J2EEApplication=none,J2EEServer=none'_'modelerType': process
```

I guess the line above the stack trace is the culprit?

brian-brazil:
I'd presume so, I can't help much with debugging from here. If you could send on a PR with the exception change, I'll merge it in.

brian-brazil:
Thanks for the patch, can this be closed off now?

levilovelock:
Yep! Thank you :)

PR #18, opened 2015-04-23
ISSUE-17: Added a catch for UnmarshalException

levilovelock:
Simply contains an exception catch for something that was being thrown when I tried to connect to the jmx connector of Bamboo. mvn package and mvn test worked for me


levilovelock:
Sorry, I'm not very familiar with the github style, This was meant to be connected to #17

brian-brazil:
Thanks!

PR #20, opened 2015-05-21
Workaround lack of ordering in JSONObject.

brian-brazil:
CollectorRegistry.getSampleValue assumes that the labelNames ordering
matches that of what was provided in the sample. While this is generally
workable, it's not true when it's coming from an unordered JSON map.
See https://github.com/prometheus/client_java/issues/71

Workaround this by sorting the labels we get.


juliusv:
:+1:

PR #25, opened 2015-08-03
Handle UnsupportedOperation and ReflectionExceptions during bean scraping

wrouesnel:
These changes are necessary to allow the exporter to work with Caucho Resin, and with them allow useful data to retrieved.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -101,6 +101,12 @@ public class JmxScraper {
             } catch(java.rmi.UnmarshalException e) {
                 logScrape(mbeanName, attr, "Fail: " + e);
                 continue;
+            } catch(java.lang.UnsupportedOperationException e) {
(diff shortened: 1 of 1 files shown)


collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -101,6 +101,12 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) th
             } catch(java.rmi.UnmarshalException e) {
                 logScrape(mbeanName, attr, "Fail: " + e);
                 continue;
+            } catch(java.lang.UnsupportedOperationException e) {
+            	logScrape(mbeanName, attr, "Fail: " + e);
+            	continue;
+            } catch(javax.management.ReflectionException e) {
```
brian-brazil: We should probably squash some of these down to a javax.management.JMException

brian-brazil:
I've merged another change that catches Exception, so I'm closing off this one.

PR #26, opened 2015-08-10
Adding files for Docker build

cjhkramer:
This will allow the Docker image to be build. 
Moved the config file to sample-conf directory so that we can mount the configuration in the docker-compose file w/o mounting the whole fs.


brian-brazil:
I'm not sure this really makes sense, the JMX exporter is intended to be used as a java agent so it'd be the Docker file of whatever service you're monitoring that you should be changing.

cjhkramer:
@brian-brazil Agreed, but in this case I'm using the stock image of ActiveMQ which exposes a JMX endpoint (see http://activemq.apache.org/jmx.html). So instead of extending the ActiveMQ docker image (which works great ootb) and creating upgrade dependencies, I used the jmx_exporter as an adapter to normalize the output for prometheus.

brian-brazil:
I'm quite hesitant to take this, as there's several things here that'll require ongoing maintenance (the version of java you're requested has 22 known vulnerabilities for example), it's not the preferred use case and this seems tied to your particular deployment rather than being a generic solution.

cjhkramer:
Fair enough, lets put this in perspective :) What approach would you recommend to bridge ActiveMQ's metrics to Prometheus then w/o extending the Active MQ Docker image?

cjhkramer:
Regarding the pullrequest and it not being a generic solution; the config file can be mounted, so you can modify your configuration and location of the endpoint you want to observe (see the docker-compose.yml for an example). A valid problem what just came to mind is that it might not be able to monitor itself as the config.json only allows for a single host.

brian-brazil:
I can't find an official ActiveMQ Docker image and thus can't tell what you're using, so the my main suggestion would be to create your own that had both ActiveMQ and the jmx_exporter.

`docker-compose.yml` and the port numbers seem specific to your environment.

cjhkramer:
We're currently using the Webcenter ActiveMQ Docker image (https://hub.docker.com/r/webcenter/activemq/~/dockerfile/), but that shouldn't make a difference as the implementation should remain the same for any application that exposes JMX beans.

To my understanding the current solution is analogue to (for example) the mysqld_exporter, where instead of a jmx url you provide a data_source_name. Extending existing applications or servers such as ActiveMQ, Stash, GitLab, etc just because they use Java seems to me against the basic architecture. 

You'd either have to introduce a supervisor pattern and run multiple JVM's within the same container (which goes against the basic design principles of Docker), or pollute the classpath of the existing Java application with the dependencies of the jmx_exporter (which can cause havoc in complex applications)

cjhkramer:
I've removed the docker-compose and updated the Dockerfile to reflect how you build/use this thing. The port numbers where just an example. We're currently running this with cloudconfig and puppet.

brian-brazil:
> To my understanding the current solution is analogue to (for example) the mysqld_exporter, where instead of a jmx url you provide a data_source_name. Extending existing applications or servers such as ActiveMQ, Stash, GitLab, etc just because they use Java seems to me against the basic architecture.

The basic architecture is to get as close to the application as possible, preferably directly instrumented inside the same process. The mysqld_exporter and the like are compromises because it's not practical to do better. With JMX we can do better.

Note that future java related features will only be possible with an agent.

> You'd either have to introduce a supervisor pattern and run multiple JVM's within the same container (which goes against the basic design principles of Docker)

I don't think anyone is proposing that.

> pollute the classpath of the existing Java application with the dependencies of the jmx_exporter (which can cause havoc in complex applications)

The dependencies are shadowed, so this won't happen.

cjhkramer:
Well, with the one-jar construction, it will force the classes of the dependencies you're including onto your classpath (org.eclipse.jetty, org.json.simple, io.prometheus.client, javax.servlet). Alternatively you could use the library version and choose the version of your dependencies yourself. So unless you're running a OSGi like pluggable architecture, you will always append these classes to your classpath.

Regarding enterprise applications like Stash, JIRA, etc. You'd either have to write a plugin or extend the container and use the supervisor pattern to run two java processes in one container. If you have another solution for that I'd be happy to hear it :)

cjhkramer:
I'm closing the pullrequest for now as this solution works for us for now. When we switch to consul for SD and sidekick bridge containers is still the way to go for applications that we do not manage we need a more complex solution that is SD aware itself anyways.

PR #37, opened 2015-09-08
Minor improvements to deb packaging

federicobaldo:
- Added postinst script to create jmx_exporter user/group and change files'
  permissions accordingly
- Fixed script /usr/bin/jmx_exporter
- Added upstart script to run jmx_exporter as jmx_exporter user

### DIFF
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -36,6 +36,7 @@
 
   <build>
     <plugins>
+
       <!-- Compile version (1.5 for generics) -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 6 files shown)


brian-brazil:
I'm not sure this makes sense, as you're presuming there's exactly one jmx_exporter running on the system. It's also using the jmx_exporter standalone, whereas using it as an agent is preferred.

federicobaldo:
Hello Brian, Thank you for the feedback. Maybe I am wrong, but I think that the deb package is exclusively for the standalone version of the jmx_exporter and it should not affect the build of the java agent version. Running multiple standalone jmx_exporters, with these changes, involves just replicating the upstart conf file and add the specific port and config file path as first and second argument to /usr/bin/jmx_exporter. 

Right now my specific use case requires the standalone mode  as we're monitoring third-party apps that are not allowing us to use the java agent version :(

brian-brazil:
Does it make sense to have upstart scripts considering systemd won the vote, and that older systems will still be on SysV? Your postinst should also be a preinst, and handle the various ways it can be called. Permissions and ownership should be handled in the deb itself, not in package scripts. 

I'm not a fan of the debian packaging in this repo as it's not good packaging (see #5 for example) and am tempted to remove it in favour of having a version on maven central.

federicobaldo:
Ok, I will keep for the moment the packaging with upstart in my fork, until I can provide a comprehensive solution. 
I am closing this pr and  opening a new one just to fix the script under /usr/bin.
Thanks

PR #39, opened 2015-10-08
Added Yaml file support, reconfigured JmxCollector and test cases.  

inondle:
Used snakeYaml to update the JmxCollector to handle yaml files. I also updated the test case strings, sample_config, and the example config files to yaml.


collector/pom.xml
```diff
@@ -38,6 +38,11 @@
       <version>4.11</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
brian-brazil: Can the json dependency be removed now?

example_configs/cassandra.yml
```diff
@@ -0,0 +1,15 @@
+---
+lowercaseOutputLabelNames: true
+lowercaseOutputName: true
+rules:
+- labels:
+    address: "$2"
+  name: cassandra_$1_$3
+  pattern: org.apache.cassandra.metrics<type=(Connection|Streaming), scope=(S*), name=(S*)><>(Count|Value)
```
brian-brazil: From a readability standpoint pattern, name, labels would make more sense in the examples.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -72,43 +73,43 @@ private JmxCollector(JSONObject config) throws ParseException, MalformedObjectNa
         }
 
         if (config.containsKey("whitelistObjectNames")) {
-          JSONArray names = (JSONArray) config.get("whitelistObjectNames");
-          for (Object name : names) {
-            whitelistObjectNames.add(new ObjectName((String)name));
+          List<Object> names = (List<Object>) config.get("whitelistObjectNames");
+          for(Object name : names) {
+              whitelistObjectNames.add(new ObjectName((String)name));
```
brian-brazil: This code is currently using 2 space indents

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -49,14 +44,20 @@
...
+    public JmxCollector(Reader in) throws IOException, MalformedObjectNameException {
+        this((Map<String, Object>)new Yaml().load(in));
     }
-    public JmxCollector(String jsonConfig) throws ParseException, MalformedObjectNameException {
-        this((JSONObject)new JSONParser().parse(jsonConfig));
+    public JmxCollector(String yamlConfig) throws MalformedObjectNameException {
+        this((Map<String, Object>)new Yaml().load(yamlConfig));
     }
+    private JmxCollector(Map<String, Object> config) throws MalformedObjectNameException {
+        if(config == null) {  //Yaml config empty, set all to default values
```
brian-brazil: From a maintenance standpoint, it may be better to replace config with an empty map here. That way the defaults don't have to be kept up to date in two places.

brian-brazil:
Generally looks great, thanks!

collector/pom.xml
inondle: I removed the json dependency and it still built, I don't think anything else used json other than the collector. Although in the javaagent module's pom.xml there is a line that includes "org.json.**" in something. I'm not sure what that does specifically but that might need to be changed.

brian-brazil: That's the shadowing so that the agent doesn't have the same package names as the code. It'll need changing to org.yaml.

inondle:
Hey, I fixed those minor issues. Feel free to look over the code again and let me know if you notice something that needs changing.

brian-brazil:
Could you squash your commits please?

inondle:
I think I squashed them down to the one commit (7e163ef). This is my first time doing this so let me know if I mucked something up.

brian-brazil:
Yip, that's correct. Thanks!

The same needs doing on the cloudwatch_exporter.

brian-brazil:
Hmm, looks like this change lost all the backslashes in the sample configs.

PR #40, opened 2015-10-14
Fix compiler error: trying to instantiate a Map directly

andersschuller:
Currently the code does not compile, because a Map cannot be instantiated.


brian-brazil:
Thanks!

PR #44, opened 2015-12-01
Dropwizard, Datastax Client and Kafka consumer/producers examples

olvesh:
Hope these looks ok. They work in my settings but not sure how "correct" they are.

### DIFF
--- a/example_configs/cassandra-datastax-driver.yml
+++ b/example_configs/cassandra-datastax-driver.yml
@@ -1,0 +1,35 @@
+---
+lowercaseOutputName: true
+rules:
+# These metrics was re-added to a dropwizard-context, so they were partly renamed.
+# In jmx they will appear as:
+# $cluster_name$-metrics{name=requests}[]Count: 48
+# where $cluster_name$ is the cluster name from the server
+# So your milage might vary
+- pattern: "^metrics<name=cassandra.([a-zA-Z-]+).(.+)><>Count"
+  name: cassandra_$2
+  type: COUNTER
+  labels:
+    keyspace: "$1"
+
+
+- pattern: "^metrics<name=cassandra.([a-zA-Z-]+).(.+)><>Value"
+  name: cassandra_$2
+  type: GAUGE
+  labels:
+    keyspace: "$1"
+
+- pattern: metrics<name=cassandra.([a-zA-Z-]+).(.+)><>(FiveMinuteRate|MeanRate|OneMinuteRate|FifteenMinuteRate)
+  name: cassandra_$2_rate_second
+  type: GAUGE
+  labels:
+    keyspace: "$1"
+    calculation: "$3"
+
+
+- pattern: metrics<name=cassandra.([a-zA-Z-]+).(.+)><>(Max|MinMean|StdDev|50thPercentile|75thPercentile|95thPercentile|98thPercentile|99thPercentile|999thPercentile)
+  name: cassandra_$2_duration_milliseconds
+  type: GAUGE
+  labels:
+    keyspace: "$1"
+    calculation: "$3"
--- a/example_configs/cassandra-datastax-driver.yml
+++ b/example_configs/cassandra-datastax-driver.yml
@@ -23,18 +23,3 @@ rules:
   type: GAUGE
   labels:
     keyspace: "$1"
-
-- pattern: metrics<name=cassandra.([a-zA-Z-]+).(.+)><>(FiveMinuteRate|MeanRate|OneMinuteRate|FifteenMinuteRate)
-  name: cassandra_$2_rate_second
-  type: GAUGE
-  labels:
-    keyspace: "$1"
-    calculation: "$3"
-
-
-- pattern: metrics<name=cassandra.([a-zA-Z-]+).(.+)><>(Max|MinMean|StdDev|50thPercentile|75thPercentile|95thPercentile|98thPercentile|99thPercentile|999thPercentile)
-  name: cassandra_$2_duration_milliseconds
-  type: GAUGE
-  labels:
-    keyspace: "$1"
-    calculation: "$3"
--- a/example_configs/cassandra-datastax-driver.yml
+++ b/example_configs/cassandra-datastax-driver.yml
@@ -11,14 +11,14 @@ rules:
 # So your milage might vary
 # When adding the exiting (datastax) metrics registry in dropwizard I add the keyspace-name so it is easier to sparate the metrics.
 
-- pattern: "^metrics<name=cassandra.([a-zA-Z-]+).(.+)><>Count"
+- pattern: "^metrics<name=cassandra.([a-zA-Z0-9_]+).(.+)><>Count"
   name: cassandra_$2
   type: COUNTER
   labels:
     keyspace: "$1"
 
 
-- pattern: "^metrics<name=cassandra.([a-zA-Z-]+).(.+)><>Value"
+- pattern: "^metrics<name=cassandra.([a-zA-Z0-9_]+).(.+)><>Value"
   name: cassandra_$2
   type: GAUGE
   labels:
--- a/example_configs/cassandra-datastax-driver.yml
+++ b/example_configs/cassandra-datastax-driver.yml
@@ -1,11 +1,16 @@
 ---
 lowercaseOutputName: true
 rules:
-# These metrics was re-added to a dropwizard-context, so they were partly renamed.
+# These metrics were re-added to a dropwizard metrics registry, and partly renamed from what you would
+# see from a vanilla datastax-driver jmx beans.
+# Using these standalone from the jmx-beans exposed by the cassandra driver these must be rewritten as e.g:
+# ^(.+)-metrics<name=(.+)><>Count
+# where $1 is the cluster-name adn $2 is the metric.
 # In jmx they will appear as:
-# $cluster_name$-metrics{name=requests}[]Count: 48
-# where $cluster_name$ is the cluster name from the server
+# cluster-name-metrics{name=requests}[]Count: 48
 # So your milage might vary
+# When adding the exiting (datastax) metrics registry in dropwizard I add the keyspace-name so it is easier to sparate the metrics.
+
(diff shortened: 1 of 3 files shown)


example_configs/cassandra-datastax-driver.yml
```diff
@@ -0,0 +1,35 @@
+---
+lowercaseOutputName: true
+rules:
+# These metrics was re-added to a dropwizard-context, so they were partly renamed.
```
brian-brazil: s/was/were/

example_configs/cassandra-datastax-driver.yml
```diff
@@ -0,0 +1,35 @@
...
+    keyspace: "$1"
+
+
+- pattern: "^metrics<name=cassandra.([a-zA-Z-]+).(.+)><>Value"
+  name: cassandra_$2
+  type: GAUGE
+  labels:
+    keyspace: "$1"
+
+- pattern: metrics<name=cassandra.([a-zA-Z-]+).(.+)><>(FiveMinuteRate|MeanRate|OneMinuteRate|FifteenMinuteRate)
```
brian-brazil: The above two already cover what these metrics expose.

example_configs/cassandra-datastax-driver.yml
```diff
@@ -0,0 +1,35 @@
...
+  name: cassandra_$2_rate_second
+  type: GAUGE
+  labels:
+    keyspace: "$1"
+    calculation: "$3"
```
brian-brazil: As a general note, this is not a good label as it's not meaningful to sum/average across it.

example_configs/cassandra-datastax-driver.yml
```diff
@@ -0,0 +1,35 @@
...
+
+
+- pattern: metrics<name=cassandra.([a-zA-Z-]+).(.+)><>(Max|MinMean|StdDev|50thPercentile|75thPercentile|95thPercentile|98thPercentile|99thPercentile|999thPercentile)
```
brian-brazil: As with above

example_configs/cassandra-datastax-driver.yml
```diff
@@ -0,0 +1,35 @@
...
+- pattern: "^metrics<name=cassandra.([a-zA-Z-]+).(.+)><>Count"
+  name: cassandra_$2
+  type: COUNTER
+  labels:
+    keyspace: "$1"
+
+
+- pattern: "^metrics<name=cassandra.([a-zA-Z-]+).(.+)><>Value"
+  name: cassandra_$2
+  type: GAUGE
```
brian-brazil: This is a COUNTER

example_configs/dropwizard-0-8-x.yml
```diff
@@ -0,0 +1,153 @@
+---
+lowercaseOutputLabelNames: true
+lowercaseOutputName: true
+blackListObjectNames: ["metrics:type=*,name=jvm*"]
+
+rules:
+
+- pattern: "metrics<name=ch.qos.logback.core.Appender.(debug|error|info|trace|warn)><>Count"
+  name: log_msg_total
```
brian-brazil: it's probably best to prefix all of these with `logback`. Are you aware that the java client has a logback module?

example_configs/dropwizard-0-8-x.yml
```diff
@@ -0,0 +1,153 @@
...
+  type: COUNTER
+
+- pattern: "metrics<name=ch.qos.logback.core.Appender.(debug|error|info|trace|warn)><>(FiveMinuteRate|MeanRate|OneMinuteRate|FifteenMinuteRate)"
+  name: log_msg_rate
+  labels:
+    level: "$1"
+    calculation: "$2"
+  type: COUNTER
+
+- pattern: "metrics<name=ch.qos.logback.core.Appender.all><>Count"
```
brian-brazil: No need for this, the above covers it

example_configs/cassandra-datastax-driver.yml
```diff
@@ -0,0 +1,35 @@
+---
+lowercaseOutputName: true
+rules:
+# These metrics was re-added to a dropwizard-context, so they were partly renamed.
+# In jmx they will appear as:
+# $cluster_name$-metrics{name=requests}[]Count: 48
```
brian-brazil: I'm confused by this and the rules you have below - is the cluster name there or not?

example_configs/dropwizard-0-8-x.yml
```diff
@@ -0,0 +1,153 @@
...
+    port: "$1"
+
+
+- pattern: "metrics<name=io.dropwizard.jetty.MutableServletContextHandler.(.+)-responses><>Count"
+  name: http_response_codes_total
+  type: COUNTER
+  labels:
+    http_code: "$1"
+
+- pattern: "metrics<name=io.dropwizard.jetty.MutableServletContextHandler.active-(.+)><>Count"
```
brian-brazil: These rules can probably be squashed down.

What's in the `(.+)` ?

example_configs/dropwizard-0-8-x.yml
```diff
@@ -0,0 +1,153 @@
...
+
+
+- pattern: metrics<name=io.dropwizard.jetty.MutableServletContextHandler.dispatches><>(Max|MinMean|StdDev|50thPercentile|75thPercentile|95thPercentile|98thPercentile|99thPercentile|999thPercentile)
+  name: http_dispatches_duration_milliseconds
+  type: GAUGE
+  labels:
+    calculation: "$1"
+
+
+- pattern: "^metrics<name=(jvm.+)><>Count"
```
brian-brazil: You should drop all these, we already expose them automatically and in a more prometheus-friendly way

example_configs/dropwizard-0-8-x.yml
```diff
@@ -0,0 +1,153 @@
...
+  help: "possibly duplicated byt non-dw prefixed jvm metrics and should be ignored by blacklist above"
+
+
+- pattern: "^metrics<name=(jvm.+)><>Value"
+  name: dw_$1
+  type: GAUGE
+  help: "possibly duplicated byt non-dw prefixed jvm metrics and should be ignored by blacklist above"
+
+
+- pattern: "^metrics<name=(org.eclipse.jetty.util.thread.QueuedThreadPool.dw.+)><>Value"
```
brian-brazil: The threadpool would be a good label

example_configs/kafka-consumer-producer-0-8-2.yml
```diff
@@ -0,0 +1,108 @@
+---
+
+lowercaseOutputName: true
+rules:
+
+# Producer rules #
+- pattern: "^kafka.producer<type=(.+), name=(.+), brokerHost=(.+), brokerPort=(.*)><>Count"
+  name: kafka_producer_$1_$2_total
+  labels:
+    broker_host: "$3"
```
brian-brazil: I'd put this in one label called 'broker'

olvesh:
Regarding quantiles etc: When prometheus exposes quantiles, is the quantile label special in some regard?  If so, would it not make sense then to rewrite the labels e.g here or in the scrape config in the prometheus server?

brian-brazil:
Generally I'd not expose quantiles, the Value/Count are cheaper, aggregate and generally quite sufficient.

olvesh:
Ok - but it is possible the way I described? And to make it work do I need to relabel to e.g. `quantile: 0.99l`?

I would like to test both, but I can exclude the percentile/quantile examples from the files if you like.

example_configs/dropwizard-0-8-x.yml
olvesh: `dispatches|requests|suspended`

olvesh: The first one lists count of opened http connections on different ports.
The middle-one lists ´1xx|2xx|4xx|5xx`

example_configs/dropwizard-0-8-x.yml
olvesh: All dropwizard apps use logback. It also hijacks java.util.Logging and Log4J + SLF4J CommonsLogging (and possibly others). So I see no use in pointing out that it is logback  - but can add a label if you like.

example_configs/cassandra-datastax-driver.yml
olvesh: This is a GAUGE: (output from JmxScraper)
```
metrics{name=cassandra.gatekeeper.connected-to}[]Value: 1
metrics{name=cassandra.gatekeeper.known-hosts}[]Value: 1
metrics{name=cassandra.gatekeeper.open-connections}[]Value: 2
```

brian-brazil:
> Ok - but it is possible the way I described? And to make it work do I need to relabel to e.g. quantile: 0.99?

Yes, it's possible. The quantile label is mainly a convention.

example_configs/dropwizard-0-8-x.yml
brian-brazil: These metrics are coming from logback, thus they should have 'logback' in the name. It's important to be able to guess where a metric is from from it's name.

example_configs/dropwizard-0-8-x.yml
brian-brazil: That needs to be in a label then, otherwise the metrics will clash.

example_configs/cassandra-datastax-driver.yml
brian-brazil: Are cassandra keyspace names limited to `a-zA-Z-` ?

olvesh:
Would it make more sense to make a metric for each quantile / calculation instead of having them as labels?

brian-brazil:
The standard in Prometheus is that they're all the same metric, but I'd keep things simple and focus on the Count/Value.

example_configs/dropwizard-0-8-x.yml
olvesh: Struggling finding a good label name for this  - suggestions?

brian-brazil: Offhand, state or status.

olvesh:
Closing as it is obviously no interest in these changes.

PR #45, opened 2015-12-15
Add auto reload feature when config changes

mchataigner:
without having to restart the program it is loaded with.

### DIFF
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -1,16 +1,117 @@
 package io.prometheus.jmx;
 
+import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.exporter.MetricsServlet;
 import io.prometheus.client.hotspot.DefaultExports;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
 import java.lang.instrument.Instrumentation;
 import java.io.FileReader;
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -1,24 +1,121 @@
...
 
+import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.exporter.MetricsServlet;
 import io.prometheus.client.hotspot.DefaultExports;
+
+import java.io.File;
+import java.io.IOException;
 import java.lang.instrument.Instrumentation;
 import java.io.FileReader;
+import java.nio.file.FileSystem;
```
brian-brazil: I'm afraid this was only added in Java 7, so we can't use it. 2/3 of our users still use Java 6.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -1,24 +1,121 @@
...
+   static JmxCollector previousJmxCollector = null;
+
+  /**
+   * This will load the config located at @configPath and replace the previous one.
+   * @param configPath
+   * @throws IOException
+   */
+   public static void loadConfig(String configPath) throws IOException {
+     if(previousJmxCollector != null) {
+       CollectorRegistry.defaultRegistry.unregister(previousJmxCollector);
```
brian-brazil: There's a race here where no results may be returned. It'd simpler to have the Collector check the timestamp on each collection, and reload if needed.

mchataigner: Do you mean putting this autoreload stuff directly inside the JmxCollector?

brian-brazil: Yes, you could make it so that if a filename is passed in it'd apply the logic.

brian-brazil:
Are you still interested in working on this feature?

mchataigner:
We decided for a while to change strategy, but recently changed our mind. I will work on it this week and push you another MR.
Thanks.

elecnix:
Now that #90 has been merged, this one should be closed.

PR #51, opened 2016-01-19
Cannot connect when using credentials

jamespic:
It's not currently possible to connect to servers that use credentials, because the credentials are configured incorrectly:

The key for credentials is not `"javax.management.remote.JMXConnector.CREDENTIALS"`, but `javax.management.remote.JMXConnector.CREDENTIALS` (which is in fact equal to `"jmx.remote.credentials"`).

Also, credentials must be passed in as an array, rather than as a list.

The attached pull request rectifies these issues.


brian-brazil:
@evgenyra

brian-brazil:
Looking through the docs, this appears to be correct. Thanks!

Issue #52, opened 2016-01-20
Exports bad label names


bitplane:
Something in the MBean flattening code is causing the exporter to spit out invalid Prometheus label names. If I export everything from one of my nodes I get a row like this:
    org_apache_camel_camel_classname_Timeout{type="endpoints",name="\"some://endpoint",8_size="50000\"",} 30000.0

The "8_size" being an invalid label name in Prometheus as it starts with a number.

brian-brazil:
Thanks for the report, can you share the mbean name per the HELP string?

bitplane:
It looks like this in the output:

    # HELP org_apache_camel_camel_ourinternalname_Timeout Shutdown timeout (org.apache.camel<context=camel-ourinternalname, type=context, name="camel-ourinternalname"><>Timeout)
    # TYPE org_apache_camel_camel_ourinternalname_Timeout gauge

I couldn't find it in JConsole though

brian-brazil:
If you can use a regex to get the jmx export just to export that metric and get the HELP string, that'd be useful. Then we can use it in a unittest.

bitplane:
As per chat with Brian on IRC, it looks like it's due to the beanName containing delimiters:

    org.apache.camel:context=ourinternalname,type=endpoints,name="seda://endpointName\?concurrentConsumers=8&size=50000"

Which are parsed here:

https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxScraper.java#L150

    String[] tokens = beanName.substring(idx).split(",|=");

Need something like an ordered version of `getKeyPropertyList`

PR #53, opened 2016-01-21
Fix issue #52

jamespic:
Fix for issue #52 - uses regex based on ObjectName spec.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -3,7 +3,6 @@ package io.prometheus.jmx;
 import java.io.IOException;
 
 import java.lang.management.ManagementFactory;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -3,15 +3,12 @@
...
 
 import java.lang.management.ManagementFactory;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
```
brian-brazil: All dependencies should be explicitly listed so you can figure out where they're coming from.

jamespic: No problem. My IDE has a tendency to collect together dependencies, but I'll uncollect them.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -137,17 +151,19 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) {
...
-        if (idx > 0) {
-            String[] tokens = beanName.substring(idx).split(",|=");
-            for (int i=0; i<tokens.length-1; )
-                output.put(tokens[i++], tokens[i++]);
+        String properties = mbeanName.getKeyPropertyListString();
+        Matcher match = PROPERTY_PATTERN.matcher(properties);
+        while (match.lookingAt()) {
+            output.put(match.group(1), match.group(2));
+            properties = properties.substring(match.end());
+            if (properties.startsWith(","))
```
brian-brazil: You should use braces here for clarity

brian-brazil:
That generally looks good, thanks.

jamespic:
I've squashed in the changes you asked for

brian-brazil:
Great, thanks!

PR #54, opened 2016-01-28
Implemented arbitrary JMX URLs

jamespic:
This pull request essentially recreates the changes proposed in #12, but against the current codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ rules:
 ```
 Name     | Description
 ---------|------------
-jmxUrl | A full JMX URL to connect to. Should not be specified if hostPort is.
 hostPort | The host and port to connect to via remote JMX. If neither this nor jmxUrl is specified, will talk to the local JVM.
+jmxUrl | A full JMX URL to connect to. Should not be specified if hostPort is.
 lowercaseOutputName | Lowercase the output metric name. Applies to default format and `name`. Defaults to false.
 lowercaseOutputLabelNames | Lowercase the output metric label names. Applies to default format and `labels`. Defaults to false.
 whitelistObjectNames | A list of [ObjectNames](http://docs.oracle.com/javase/6/docs/api/javax/management/ObjectName.html) to query. Defaults to all mBeans.
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ The configuration is in YAML. An example with all possible options:
 ```
 ---
 hostPort: 127.0.0.1:1234
+jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:1234/jmxrmi
 lowercaseOutputName: false
 lowercaseOutputLabelNames: false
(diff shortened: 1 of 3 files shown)


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -26,6 +26,8 @@
 
 public class JmxCollector extends Collector {
     private static final Logger LOGGER = Logger.getLogger(JmxCollector.class.getName());
+    public static final String JMX_URL_KEY = "jmxUrl";
```
brian-brazil: We're not doing this for any other constant, and I don't think the indirection is adding much

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -60,18 +62,20 @@ private JmxCollector(Map<String, Object> config) throws MalformedObjectNameExcep
...
+              throw new IllegalArgumentException("At most one of hostPort and jmxUrl must be provided");
+          }
+          jmxUrl ="service:jmx:rmi:///jndi/rmi://" + (String)config.get(HOST_PORT_KEY) + "/jmxrmi";
+        } else if (config.containsKey(JMX_URL_KEY)) {
+          jmxUrl = (String)config.get(JMX_URL_KEY);
         }
 
         if (config.containsKey("username")) {
-            username = (String)config.get("username");
+            this.username = (String)config.get("username");
```
brian-brazil: no need for the `this` here

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -60,18 +62,20 @@ private JmxCollector(Map<String, Object> config) throws MalformedObjectNameExcep
             config = new HashMap<String, Object>();
         }
 
-        if (config.containsKey("hostPort")) {
-          hostPort = (String)config.get("hostPort");
-        } else {
-          // Default to local JVM.
```
brian-brazil: I think we should keep the else and this comment so that it's explicit what the default is.

README.md
```diff
@@ -37,7 +38,8 @@ rules:
 ```
 Name     | Description
 ---------|------------
-hostPort | The host and port to connect to via remote JMX. If not specified, will talk to the local JVM.
+jmxUrl | A full JMX URL to connect to. Should not be specified if hostPort is.
+hostPort | The host and port to connect to via remote JMX. If neither this nor jmxUrl is specified, will talk to the local JVM.
```
brian-brazil: hostPort should go first, as that'll be more common

brian-brazil:
This looks generally good, thanks.

jamespic:
I've updated it based on your suggestions

brian-brazil:
Thanks!

PR #56, opened 2016-02-03
Add label to jmx_scrape_duration_seconds and jmx_scrape_error so that…

allenxiang:
… multiple hosts can be exported by one JMX exporter.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -6,6 +6,7 @@ import java.io.PrintWriter;
 import java.io.Reader;
 import java.io.StringWriter;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


brian-brazil:
It is intended that you run one exporter per instance, preferably as an agent inside the JVM.

allenxiang:
We have a cluster of kafka nodes to be monitored and it is quite difficult to run exporter as agent since the cluster is managed by some other framework. I see there is some need to support multiple targets with one central exporter.

brian-brazil:
The ability to support multiple hosts to an exporter like this where there's no fundamental problem with running one per application is a non-goal to support. Running with one central exporter is harder to support and manage in general.

dylanmei:
@allenxiang in case the "framework" you are referring to is [mesos-kafka](https://github.com/mesos/kafka), you can bundle `jmx_exporter` with your Kafka distributable.

    tar -cvf kafka_2.10-0.8.2.1.tgz kafka_2.10-0.8.2.1 jmx_prometheus_javaagent.jar jmx_exporter.yml

The kafka scheduler can be instructed to use this distributable along with the `-javaagent` hint in the scheduler's --jvm-options arg.

    broker add ... --jvm-options -Xms1024m ... -javaagent:jmx_prometheus_javaagent.jar=1234:jmx_exporter.yml ...

Now you have one exporter per JVM.

allenxiang:
@dylanmei: Came to the same solution like you suggested last week and it works well. Thanks anyways.

brian-brazil:
All is good then, thanks.

PR #57, opened 2016-02-05
Use Java collection interfaces and add private constructor

AymanDF:
This pull request is focused on resolving occurrences of Sonar rules
squid:S1319 - “ Declarations should use Java collection interfaces such as "List" rather than specific implementation classes such as "LinkedList" ”. You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S1319
And
squid:S1118 - “ Utility classes should not have public constructors ”. You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S1118
Please let me know if you have any questions.
Ayman Abdelghany.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -33,8 +33,8 @@ public class JmxCollector extends Collector {
       String help;
       boolean attrNameSnakeCase;
       Type type = Type.GAUGE;
-      ArrayList<String> labelNames;
-      ArrayList<String> labelValues;
+      List<String> labelNames;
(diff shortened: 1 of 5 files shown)


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -186,8 +186,8 @@ void addSample(MetricFamilySamples.Sample sample, Type type, String help) {
 
       private void defaultExport(
           String domain,
-          LinkedHashMap<String, String> beanProperties,
```
brian-brazil: That this has a specific order is an important part of the interface, so these should stay as is.

PR #60, opened 2016-02-16
health endpoint

f1yegor:
just health endpoint to check that jmx_exporter is working.
because jmx_exporter now has no cache, it's heavy to use /metrics for this.

### DIFF
--- a/jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/HealthServlet.java
+++ b/jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/HealthServlet.java
@@ -1,0 +1,30 @@
+package io.prometheus.jmx;
+
+import io.prometheus.client.exporter.common.TextFormat;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.Writer;
+
+public class HealthServlet extends HttpServlet {
+
+    @Override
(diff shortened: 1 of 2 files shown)


brian-brazil:
Why do you want to add this? This result is not valid with that content type.

f1yegor:
Fixed content type.
Some services monitor health status of exporter.

brian-brazil:
Which services are these? Prometheus only uses /metrics to determine health.
We don't currently offer this sort of endpoint on any of our exporters, so if we're adding it we should be consistent across all services.

I'm hesitant to accept a way that has two different ways of indicating health (content and status code) as I've seen that get confusing in the past.

f1yegor:
I use mesos to manage exporter with health check. Unfortunately mesos provides only GET http requests for health, which is high-consuming for exporter.

brian-brazil:
Last time I checked mesos doesn't support health checks. Which scheduler are you using? Aurora at least is quite flexible in this regard, as I made it generic about a year ago.

f1yegor:
marathon

brian-brazil:
Have you considered running this as an agent rather than a http server? That's the recommended way to do things, as it avoids management hassle like this and gives you better metrics.

f1yegor:
From my point of view agent way is not so straightforward, because it's hard to figure out how to deal with different frameworks. Somebody already described how to do it for kafka-mesos, and it's not easy.

brian-brazil:
If adding a command line argument to one of your applications is difficult, I'd suggest working on it until it isn't. That's a fundamental operational ability that you shouldn't have to think about.

josdotso:
Would like to see this happen for use in Kubernetes liveness and readiness probes.

PR #61, opened 2016-02-16
parallel scraper

f1yegor:
it occurs that ask mbean server is slow. do it in parallel. 
is there reasons why project compiles to target 1.5?

### DIFF
--- a/collector/pom.xml
+++ b/collector/pom.xml
@@ -42,8 +42,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.5</source>
-          <target>1.5</target>
+          <source>1.8</source>
+          <target>1.8</target>
         </configuration>
       </plugin>
       <plugin>
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -2,11 +2,7 @@ package io.prometheus.jmx;
 
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.management.JMException;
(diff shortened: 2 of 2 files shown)


collector/pom.xml
```diff
@@ -42,8 +42,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.5</source>
-          <target>1.5</target>
+          <source>1.8</source>
```
brian-brazil: 2/3 of our users are still on 1.6, 1.8 is barely in use. This needs to stay as-is for many years to come.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -2,11 +2,7 @@
 
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
```
brian-brazil: Please leave imports explicit.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -73,9 +68,8 @@ public void doScrape() throws Exception {
                 mBeanNames.removeAll(beanConn.queryNames(name, null));
             }
 
-            for (ObjectName name : mBeanNames) {
-                scrapeBean(beanConn, name);
-            }
+            mBeanNames.parallelStream().forEach(name -> scrapeBean(beanConn, name));
```
brian-brazil: This is a Java 1.8 construct, please stick strictly to 1.6 features.

brian-brazil: parallelStream was only added in 1.8, it'll be at least 2020 before 1.8 is so widespread that we can safely assume that everyone uses at least 1.8.

In addition, what's the cpu impact of doing this in parallel? Spiky cpu usage due to monitoring is undesirable.

f1yegor:
Yeah, I afraid because of 1.5. So I'll keep this only in my fork. Feel free to decline because of it.
And yes, it's more spiky.

brian-brazil:
Spiky is bad when it comes to non-critical cpu usage, you almost never want to delay a user request for a monitoring request. If you're finding scrapes are too slow then I recommend you either blacklist mbeans, or increase your scrape interval.

PR #62, opened 2016-02-25
Add zookeeper example based on #42 from @robbiet480

bobrik:


example_configs/zookeeper.yaml
```diff
@@ -0,0 +1,23 @@
+rules:
+  - pattern: "org.apache.ZooKeeperService<name0=ReplicatedServer_id(\\d)><>(\\w+)"
+    name: "zookeeper_$2"
+    labels:
+      brokerId: "$1"
```
brian-brazil: Per #42, this shouldn't be a label from instrumentation as it's really a target label.

By convention we also do `broker_id`.

bobrik: Should I just drop it then?

brian-brazil: Yes, there can only be one broker per server.

bobrik: Done

brian-brazil:
Thanks!

PR #64, opened 2016-03-12
Add Dockerfile

jdubs:
Add a docker file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM java:8
+FROM java:8-jre
 MAINTAINER  John-William Trenholm <jw@xomar.com>
 
 WORKDIR /jmx_exporter
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM java:8
+MAINTAINER  John-William Trenholm <jw@xomar.com>
+
+WORKDIR /jmx_exporter
+COPY jmx_prometheus_httpserver/target/jmx_prometheus_httpserver-0.*-SNAPSHOT-jar-with-dependencies.jar \
+     /jmx_exporter/jmx_prometheus_httpserver.jar
+
+EXPOSE 9098
+ENTRYPOINT [ "java", "-Dcom.sun.management.jmxremote.ssl=false", \
+	     "-Dcom.sun.management.jmxremote.authenticate=false", \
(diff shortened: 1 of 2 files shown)


brian-brazil:
The recommended way to run this exporter is as a java agent, and this dockerfile isn't very useful as you need to change it to put in both a JMX port number and configuration file for your service.

alon-argus:
Hey @brian-brazil 

Given Kubenetes' feature of "Pod", would you reconsider this recommendation of running as a java agent? 

Running an independent jmx-scrape http server, in a separate container but in the same pod alongside the scraped service, seems to be more convenient and even more correct than adding a java agent to each service we monitor. 

That jmx-http-server image (dockerfile) can be generic - the port and configuration file can be injected in runtime (In Kubenetes this can be done with Config Map, for instance).

brian-brazil:
I don't see how that changes anything. It merely makes it easier to run it in a non-recommended fashion.

You want your monitoring as tightly bound to what your monitoring, and inside the JVM is the closest we can get to that with JMX.

alon-argus:
Devops-wise,

In the recommended java agent way, you have to edit each Docker image you're using - (1) add jmx-exporter's jar and config, and (2) change the JVM_OPTS or edit the CMD to add the javaagent option. 
This way you'll probably have to create and maintain an additional Docker image of your own for each monitored service.
(For example - https://hub.docker.com/r/argussecurity/cassandra-prometheus/)
Please correct me if I'm wrong.

In the other way, you can just create a generic "jmx-http-server" Docker image, and attach it to every monitored service, quite easily using Kubernetes' pod. 

Thanks for your answers.

brian-brazil:
I'd consider the jmx agent easier here, as it bring things closer to the ideal where everything lives inside exactly one container.

There's likely as much work overall attaching an additional container to a pod as putting it in the main container in the first place, and you get better metrics.

alon-argus:
May I just ask, why do you say that it's better metrics when it runs as a java agent?

brian-brazil:
We can only get the process metrics when it's running in the same process.

jdubs:
All of those metrics are exposed via JMX, there shouldn't be any difference when queried from a side car vs externally.

brian-brazil:
Not all of them come from JMX.

PR #66, opened 2016-03-30
systemd unit for CentOS7

mcortinas:
uploaded systemd example unit for centos load config.xml on /opt/jmx_exporter path


mcortinas:
just upload systemd script file

brian-brazil:
Thanks for the PR, this doesn't really make sense for this particular exporter.

Running it as a java agent is the recommended usage. Running it standalone requires per-application configuration, including things like port numbers and config file locations that are going to vary by application so we can't provide any reasonable defaults for these. You're also hardcoding assumptions about your setup into this file.

PR #67, opened 2016-05-06
Update run_sample_httpserver.sh

shpsec-dennis:
Update to .7, there's probably a better way to do this as part of the build with mvn


brian-brazil:
Thanks!

PR #69, opened 2016-06-30
Let user specify context root

magnusr:


brian-brazil:
This is a http server that only serves to expose Prometheus metrics, why do you care what the endpoint is called?

I'm wary of making the command line more complicated for the agent, it already trips people up as-is.

magnusr:
Because we scrape many endpoints from Grafana, and the path is common for all endpoints in a job. Would be nice to have our JMX endpoints be part of the same job that scrapes everything else.

brian-brazil:
Last I checked Grafana doesn't support scraping Prometheus endpoints. Prometheus does support custom metrics paths.

magnusr:
Sorry, I wasn't clear. We have Grafana reading from Prometheus, and Prometheus set up to scrape many endpoints. We're looking to add jmx_exporter endpoints to the list.
Prometheus supports only one path per job, and all our exiting endpoints use /prometheusMetrics as path. We could set up one job for the JMX endpoints and one for the rest, but we'd prefer to have a single job for all endpoints, as this simplifies the configuration. Hence the suggestion to make the jmx_exporter context path configurable.

brian-brazil:
> Prometheus supports only one path per job,

Actually it can be relabelled, though in this case two jobs is probably best.

It's not uncommon to end up with 2 metrics paths in an organisation, the standard `/metrics` and something org specific for their own directly instrumented binaries. Getting every single exporter to support configuring the metrics path is a sisyphean task, there's always going to be one more exporter you want to switch. Given the java agent configuration is already fairly rough and ready I don't think this change is as free as it is elsewhere where we have flags.

magnusr:
Fair enough.

PR #73, opened 2016-08-03
Added systemd unit file

KrestenKjaer:
I created a simple Unit file based on the Centos7 one and included it in the Debian package.

Verified to work on Debian 8 and Ubuntu 16.04

### DIFF
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -46,7 +46,7 @@
         </configuration>
       </plugin>
 
-      <!-- Build a full jar with dependencies --> 
+      <!-- Build a full jar with dependencies -->
       <plugin>
         <artifactId>maven-assembly-plugin</artifactId>
         <configuration>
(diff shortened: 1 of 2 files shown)


brian-brazil:
This is intended to be used as a java agent, so hardcoding to one particular deployment (particularly one that presumes specific paths and only one application needing JMX on a machine) doesn't make sense.

KrestenKjaer:
The hardcoded files are the files already provided by the package.

In order to change the java process, the user is required to edit the provided yaml configuration file.
The systemd service file, which I added, does nothing more than run the already provided binary with the already provided configuration file on port 9010.

I see it mainly as the missing piece, so end users can see how to run this service under systemd.

brian-brazil:
You're assuming though that the user is using systemd, which won't be a safe assumption for many years to come.

KrestenKjaer:
Hi Brian,

I fully understand where you are coming from.
Do you consider it an option to generate two Debian packages? The first one without systemd the other one with systemd.

I am not in favor of forking and patching projects in order to add minor features.

If you do not consider this an option, you may close this pull request.

Best regards
Kresten Kjær

brian-brazil:
The only practical option is to use sysV init, as that should work everywhere.

PR #74, opened 2016-08-20
Update to io.prometheus.simpleclient v0.0.16

mindprince:
It would be great if you can also make a new release of jmx_exporter. Thanks!

### DIFF
--- a/collector/pom.xml
+++ b/collector/pom.xml
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>io.prometheus</groupId>
       <artifactId>simpleclient</artifactId>
-      <version>0.0.11</version>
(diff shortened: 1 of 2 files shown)


brian-brazil:
What change are you looking for? I don't see anything relevant in the client_java releases.

mindprince:
I am using both io.prometheus.jmx.collector and io.prometheus.simpleclient in a project. I wanted to depend on io.prometheus.simpleclient v0.0.16 in that project but don't want the transitive dependency from io.prometheus.jmx.collector to conflict. Not a big deal but would be helpful.

brian-brazil:
I'm happy to merge it in, but I generally only do releases for features/bugs.

PR #80, opened 2016-09-21
add support rpm build

kozmagabor:


brian-brazil:
Thanks for the PR. I'm not too sure about this given I know little of RPMs, and I'm not the biggest fan of us  providing debs for just this exporter (which already isn't a normal exporter).

kozmagabor:
Thanks for your answer!
Anyway :) You have created deb support this exporter.
Do you any plans to support deb/rpm package build for exporters?

brian-brazil:
The deb support was added before I became maintainer, and it's not the best packaging ever. We don't currently plan on offering deb/rpm support, as it'd be an unending source of maintenance burden.

kozmagabor:
Ok, What is your opinion about package creation?
I need rpm package from this.
What should I do?  Should i find an other way to create rpm package?

brian-brazil:
There's nothing stopping you from building a package locally.

kozmagabor:
Ok, Thanks

PR #81, opened 2016-10-04
Allow jmx agent to communicate with a secure registry over SSL

jslivka:
When using an encrypted JMX connection for a java process, the JMX agent does not recognize these connections and sees the following error "ConnectIOException: non-JRMP server at remote endpoint" (ultimately a scrape error when you query the /metrics endpoint). 

You can provide a trust store to the JVM when starting the prometheus exporter as an agent, for example: `java -Djavax.net.ssl.trustStore=${/path/to/truststore} -javaagent:${/path/to/prometheus_javaagent/jar} ...` and you'll be able to see the MBeans

In the case of NoSQL, this is the same trust file provided as `oracle.kv.ssl.trustStore` in `client.security`

I've verified the code in this PR works for Oracle NoSQL enterprise and Apache Tomcat.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -31,8 +31,7 @@ import javax.management.remote.rmi.RMIConnectorServer;
 
 
 public class JmxScraper {
-    private static final Logger logger = Logger.getLogger(JmxScraper.class.getName());
-
+    private static final Logger logger = Logger.getLogger(JmxScraper.class.getName());;
     private static final Pattern PROPERTY_PATTERN = Pattern.compile(
             "([^,=:\\*\\?]+)" + // Name - non-empty, anything but comma, equals, colon, star, or question mark
                     "=" +  // Equals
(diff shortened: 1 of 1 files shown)


brian-brazil:
You're making a pile of unrelated changes here, can you cut it down to just what you want to change so that I can review?

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -84,14 +87,17 @@ public void doScrape() throws Exception {
...
           if(username != null && username.length() != 0 && password != null && password.length() != 0) {
-            credential = new HashMap();
             String[] credent = new String[] {username, password};
-            credential.put(javax.management.remote.JMXConnector.CREDENTIALS, credent);
+            env.put(javax.management.remote.JMXConnector.CREDENTIALS, credent);
           }       
 
-          jmxc = JMXConnectorFactory.connect(new JMXServiceURL(jmxUrl), credential);
+          SslRMIClientSocketFactory clientSocketFactory = new SslRMIClientSocketFactory();
+          env.put(Context.SECURITY_PROTOCOL, "ssl");
```
brian-brazil: This looks like it requires all users to use SSL, which wouldn't work out too well.

brian-brazil:
Superseded by #114

PR #83, opened 2016-10-12
Auto config reload

iggyzap:
Hi @brian-brazil ,

I've added capability to watch config for changes - it will reload config when it's modified time changes.

Had to format few poms, so that's cosmetic changes hopefully.

With regards, Ignat Zapolsky

### DIFF
--- a/jmx_prometheus_javaagent/pom.xml
+++ b/jmx_prometheus_javaagent/pom.xml
@@ -52,9 +52,19 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.5</source>
-          <target>1.5</target>
+          <source>1.7</source>
+          <target>1.7</target>
         </configuration>
+        <executions>
+          <execution>
+            <id>default-testCompile</id>
+            <phase>test-compile</phase>
+            <configuration>
+              <source>1.7</source>
+              <target>1.7</target>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
 
       <!-- Build a full shaded jar with dependencies -->
--- a/jmx_prometheus_javaagent/pom.xml
+++ b/jmx_prometheus_javaagent/pom.xml
@@ -52,8 +52,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.5</source>
+          <target>1.5</target>
         </configuration>
         <executions>
           <execution>
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -91,7 +91,7 @@ public class JavaAgent {
     static class JmxConfigurer extends TimerTask {
         final CountDownLatch startLatch;
         final File fileToWatch;
-        AtomicReference<Collector> collectorRef = new AtomicReference<>();
+        AtomicReference<Collector> collectorRef = new AtomicReference<Collector>();
 
         JmxConfigurer(CountDownLatch startLatch, File fileToWatch) {
             this.startLatch = startLatch;
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -1,35 +1,139 @@
 package io.prometheus.jmx;
 
+import io.prometheus.client.Collector;
+import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.exporter.MetricsServlet;
 import io.prometheus.client.hotspot.DefaultExports;
-import java.lang.instrument.Instrumentation;
-import java.io.FileReader;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.lang.instrument.Instrumentation;
+import java.util.Date;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
 public class JavaAgent {
-   static Server server;
+    static Server server;
 
-   public static void premain(String agentArgument, Instrumentation instrumentation) throws Exception {
-     String[] args = agentArgument.split(":");
-     if (args.length != 2) {
-       System.err.println("Usage: -javaagent:/path/to/JavaAgent.jar=<port>:<yaml configuration file>");
-       System.exit(1);
-     }
-     new JmxCollector(new FileReader(args[1])).register();
-     DefaultExports.initialize();
+    private final static Timer watchTimer = new Timer("Prometheus config watch", true);
 
-     int port = Integer.parseInt(args[0]);
-     server = new Server(port);
-     QueuedThreadPool pool = new QueuedThreadPool();
-     pool.setDaemon(true);
-     server.setThreadPool(pool);
-     ServletContextHandler context = new ServletContextHandler();
-     context.setContextPath("/");
-     server.setHandler(context);
-     context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
-     server.start();
-   }
+    private final static AtomicReference<FileInfo> lastChange = new AtomicReference<FileInfo>();
+
+    static class FileInfo {
+        final long lastChangeTs;
+        final String fileName;
+        final Throwable exception;
+
+        FileInfo(long lastChangeTs, String fileName) {
+            this(lastChangeTs, fileName, null);
+        }
+
+        public FileInfo(long lastChangeTs, Throwable exception) {
+            this(lastChangeTs, null, exception);
+        }
+
+        public FileInfo(long lastChangeTs, String fileName, Throwable exception) {
+            this.lastChangeTs = lastChangeTs;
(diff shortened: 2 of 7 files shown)


jmx_prometheus_javaagent/pom.xml
```diff
@@ -55,6 +55,16 @@
           <source>1.5</source>
           <target>1.5</target>
         </configuration>
+        <executions>
+          <execution>
+            <id>default-testCompile</id>
+            <phase>test-compile</phase>
+            <configuration>
+              <source>1.7</source>
```
brian-brazil: We support 1.6, as many of our users are still on that.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -1,35 +1,139 @@
...
 import io.prometheus.client.exporter.MetricsServlet;
 import io.prometheus.client.hotspot.DefaultExports;
-import java.lang.instrument.Instrumentation;
-import java.io.FileReader;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 
+import java.io.File;
```
brian-brazil: Shouldn't java imports go first?

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -1,35 +1,139 @@
...
+            FileReader reader = null;
+            FileInfo info = null;
+            try {
+                FileInfo lastLoad = lastChange.get();
+                long lastModified = fileToWatch.lastModified();
+                if (lastLoad != null && lastLoad.lastChangeTs == lastModified) {
+                    return;
+                }
+
+                //NB: logic below can result in duplicated metric blip for short time
```
brian-brazil: That will cause scrapes to break. This logic needs to go in the collector itself, so that updates are atomic.

jmx_prometheus_javaagent/pom.xml
iggyzap: That's only for tests so it won't break compatibility.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
iggyzap: I don't think so. It's registry responsibility that metrics extraction won't be affected by changed collector.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
iggyzap: Nope, there is no consistency in import code style.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
brian-brazil: That's not the registry's responsibility. If there's no collector registered when a scrape happens, you don't get metrics.

iggyzap: Why not a registry responsibility ? There is already code that makes sure that concurrent changes to collectors are visible in real-time, why not make sure that these changes are atomic and won't break collection?
```java
package io.prometheus.client;

public class CollectorRegistry {
  /**
   * The default registry.
   */
  public static final CollectorRegistry defaultRegistry = new CollectorRegistry();

  private final Set<Collector> collectors = 
      Collections.newSetFromMap(new ConcurrentHashMap<Collector, Boolean>());
 // other code there ...
}
```

brian-brazil: Why would this be the registry's responsibility? The intention is that a collector is added to the registry, and never removed. Trying to swap out collectors is not normal.

iggyzap: If that's not registry responsibility then why would it have methods to actually remove a collector as well as support for have real-time view of collectors being added / removed ?

brian-brazil: There's some niche uses cases where you'd want to remove a collector, it comes up very rarely.

iggyzap: Exactly with ability to reload jmx reporter config - comes only few times a lifetime - initially when setting up .yaml file and only when you need to dig deeper.

And you don't want to restart application. 

Very niche. But handy.

brian-brazil: That's not an expected use case for this. If a collector has a change in configuration, that should be handled inside the collector.

iggyzap: Can you point me in direction of some documentation what are expected use-cases for collectors and registry ?

brian-brazil: The java client is primarily documented in the class-level docstrings.

brian-brazil: As the author and maintainer of client_java, this is not the manner in which the Registry is meant to be used. I'm not aware of anyone using it as you suggest in Java or in any other language or exporter.

The registry allows for concurrent registration so that the user doesn't have to worry about thread safety in the standard use case.

Please change this to handle the swap inside the collector as I've previously requested.

iggyzap: Really ? https://github.com/prometheus/jmx_exporter/pull/45

iggyzap:
I had some headbanging session and found easy way preserve registry "api", hence wrapper that does few interesting things smoothing out replacement of JMX collector.

icetrain:
I would really like to see a config reload feature myself but wouldn't it be simpler to just check the file mtime on each scrape and reload if needed?

iggyzap:
@icetrain Not really, JMX collector already 360 lines long and dealing with changing config on the fly will add extra complexity. It also will add delay to scraping if config has to be re-read during scrape.

brian-brazil:
JMX is already slow enough that the few milliseconds it takes to parse a file isn't going to add much, cassandra typically takes 9-11 seconds for example.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -2,34 +2,133 @@
...
+                FileInfo lastLoad = lastChange.get();
+                long lastModified = fileToWatch.lastModified();
+                if (lastLoad != null && lastLoad.lastChangeTs == lastModified) {
+                    return;
+                }
+
+                reader = new FileReader(fileToWatch);
+                JmxCollector collector = new JmxCollector(reader);
+
+                // we don't care about double registering, since wrapper returns itself on set
```
brian-brazil: This comment doesn't make sense with the new code.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -2,34 +2,133 @@
...
+                info = new FileInfo(new Date().getTime(), e);
+            } finally {
+                lastChange.set(info);
+                startLatch.countDown();
+                try {
+                    if (reader != null) {
+                        reader.close();
+                    }
+                } catch (IOException e) {
+                    //ignore
```
brian-brazil: It'd be best to log here just in case there's something odd going on

jmx_prometheus_javaagent/src/test/java/io/prometheus/jmx/JavaAgentIT.java
```diff
@@ -1,19 +1,16 @@
...
-import java.io.InputStream;
-import java.io.InputStreamReader;
+import java.io.*;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
```
brian-brazil: Be explicit with imports

iggyzap: @brian-brazil Would you able to elaborate why would you need reload function in standalone ?

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
iggyzap: It clarifies that "double-registering" is OK

jmx_prometheus_javaagent/src/test/java/io/prometheus/jmx/JavaAgentIT.java
brian-brazil: As much as I'm not a fan of standalone, it should have equivalent functionality as far as is possible.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
brian-brazil: Registering a collector twice is not something you should be doing, it should happen only once.

jmx_prometheus_javaagent/src/test/java/io/prometheus/jmx/JavaAgentIT.java
iggyzap: And do you have an example why this functionality be of help to mAdMe ?

brian-brazil: This is an open source project, patches and other contributions are considered in the context of all users and use cases; not any one company in particular.

brian-brazil:
Closing in favour of #90

PR #84, opened 2016-10-12
Capability to blacklist object names via java regexp

iggyzap:
Hi @brian-brazil,

I've added capability to black list some of the object names via regexp (for some reason using wild-card ObjectNames as black-listed object names did not produce consistent results for me).

That regexp(s) is evaluated against canonical ObjectName, so it should be pretty stable, also since it's used as filter against names recovered from JMX Platform, it's quite efficient.

For example, following code drops number of lines in instrumented jboss 4.3.2 from ~3k to 1k:

```

---
blacklistPatterns:
- ^java.*?:.*|^jboss:.*|^jboss\.(admin|alerts|aop|bean|cache|console|deploy|ejb|ha|j2ee|jdbc|jgroups|jmx|management|messaging|mq|remoting|rmi|security|system|ws).*?:.*|^.*?:.*?(host=localhost|WebModule=|J2EEApplication=).*
```


collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -1,30 +1,20 @@
...
+import javax.management.openmbean.CompositeType;
 import javax.management.openmbean.TabularData;
 import javax.management.openmbean.TabularType;
-import javax.management.openmbean.CompositeType;
 import javax.management.remote.JMXConnector;
 import javax.management.remote.JMXConnectorFactory;
 import javax.management.remote.JMXServiceURL;
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.util.*;
```
brian-brazil: Please leave imports explcit

iggyzap: Fixed.

iggyzap:
Brian,

black-listing via ObjectName(s) is too cumbersome for app servers that expose a lot of stuff via JMX, which is the case of JBOSS. 

Also white-list in this case will be counter-productive, since it defeats idea of ease of discovery of new metrics.
Following black list should be equivalent to regexp, but it is 10x less performant and does not catch stuff which regexp does:
```
---
blacklistObjectNames:
 - java*:*
 - jboss.admin*:*
 - jboss.alerts:*
 - jboss.aop*:*
 - jboss.bean*:*
 - jboss.cache*:*
 - jboss.console*:*
 - jboss.deploy*:*
 - jboss.ejb*:*
 - jboss.ha*:*
 - jboss.j2ee*:*
 - jboss.jdbc*:*
 - jboss.jgroups*:*
 - jboss.jmx*:*
 - jboss.management*:*
 - jboss.messaging*:*
 - jboss.mq*:*
 - jboss:*
 - jboss.remoting*:*
 - jboss.rmi*:*
 - jboss.security*:*
 - jboss.system*:*
 - jboss.ws*:*
 - "*:*,host=localhost,*"
 - "*:*,WebModule=*,*"
 - "*:*,J2EEApplication=*,*"
```

brian-brazil:
I have difficulty believing that there would be any notable performance difference between the regex and the explicit list. Do you have a benchmark?

Blacklisting is intended for excluding one or two extremely expensive mBeans, not for large scale selection of mBeans. I'd suggest looking into using a rule with an empty name to blacklist what you want, as I'd prefer not to offer two different ways to do blacklisting.

Issue #79, opened 2016-09-19
cannot specify network interface the java agent listens on


bashims:
In some cases it is necessary to  have the JavaAgent listen on a specific network interface, after looking at the code I cannot see how this is currently possible.

Thanks,
-ben

PR #85, opened 2016-10-19
Allow java agent to bind to specific IP/hostname

icetrain:


icetrain:
Solves #79 if it's acceptable to specify address instead of interface name.

brian-brazil:
This seems fine. I'm a little worried that if there's ever additional parameters that it'll be tricky to add them syntactically. What do you think?

icetrain:
I choose this solution to not break existing configurations but I agree that the current parameter syntax will make future changes tricky. It would probably be a good idea to migrate to something like comma separated key/values or system properties at some point.

brian-brazil:
Let's figure it out when it comes up.

Issue #47, opened 2015-12-17
Export tags from mbeans


bobrik:
HBase exports tags for some mbeans:

``` json
{
    "name" : "Hadoop:service=HBase,name=JvmMetrics",
    "modelerType" : "JvmMetrics",
    "tag.Context" : "jvm",
    "tag.ProcessName" : "RegionServer",
    "tag.SessionId" : "",
    "tag.Hostname" : "pewpew",
    "MemNonHeapUsedM" : 93.03323,
    "MemNonHeapCommittedM" : 94.578125,
    "MemNonHeapMaxM" : -9.536743E-7,
    "MemHeapUsedM" : 3626.3154,
    "MemHeapCommittedM" : 5279.8945,
    "MemHeapMaxM" : 8000.6875,
    "MemMaxM" : 8000.6875,
    "GcCountParNew" : 6348,
    "GcTimeMillisParNew" : 177628,
    "GcCountConcurrentMarkSweep" : 13,
    "GcTimeMillisConcurrentMarkSweep" : 698,
    "GcCount" : 6361,
    "GcTimeMillis" : 178326,
    "ThreadsNew" : 0,
    "ThreadsRunnable" : 20,
    "ThreadsBlocked" : 0,
    "ThreadsWaiting" : 69,
    "ThreadsTimedWaiting" : 26,
    "ThreadsTerminated" : 0,
    "LogFatal" : 0,
    "LogError" : 0,
    "LogWarn" : 0,
    "LogInfo" : 0
  }
```

Here I'm interested in `tag.ProcessName` being `RegionServer`. For master server I care about `tag.isActiveMaster` being `true` or `false`.

I'd like to have an ability to export these into labels somehow.

brian-brazil:
For ProcessName, I suspect you should be doing that as a target label from your service discovery rather than from the target itself.

For the others, what I was thinking is that you could define the value of a metric to also come from the regex. This would let you put a 1 in for it, and also pull a label out.

bobrik:
> For ProcessName, I suspect you should be doing that as a target label from your service discovery rather than from the target itself.

I don't understand what you mean by that. Can you explain in a bit more detail?

Did you mean that I should make metric `hadoop.hbase.jvm.regionserver` instead of `hadoop.hbase.jvm{process="regionserver"}`?

> For the others, what I was thinking is that you could define the value of a metric to also come from the regex. This would let you put a 1 in for it, and also pull a label out.

```json
{
[...]

PR #88, opened 2016-10-22
Add configuration option to override/transform the scraped mBean value

icetrain:


icetrain:
Solves #47

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -240,38 +245,70 @@ public void recordBean(
...
               continue;
             }
           }
+
+          // Transform/cast value
+          Number value = null;
+          if (rule.value != null && !rule.value.isEmpty()) {
+            String val = matcher.replaceAll(rule.value);
+
+            Matcher math = mathPattern.matcher(val);
```
brian-brazil: This is a bit more complex than I think is needed, a `valueFactor` would be sufficient I think.

brian-brazil: I'd have `value` as a subsitiutable float, and `valueFactor` as number it's multiplied by. The use case is converting milliseconds to seconds, there's no need for arithmetic operations beyond that.

icetrain:
Updated with a simpler version that only multiplies value by a factor (if supplied).

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -240,38 +248,51 @@ public void recordBean(
...
+
+            try {
+              beanValue = Double.valueOf(val);
+            } catch (NumberFormatException e) {
+              LOGGER.fine("Unable to parse configured value '" + val + "' to number for bean: " + beanName + attrName + ": " + beanValue);
+              return;
+            }
+          }
+          if (beanValue instanceof Number) {
+            if (rule.valueFactor != null) {
```
brian-brazil: Might it be simpler to default it to 1?

brian-brazil:
Thanks!

Issue #29, opened 2015-08-29
Notice config file changes


brian-brazil:
If the config file changes, we should reload it rather than having to restart the process

andersschuller:
Did you already have some thoughts about how you would implement this? From some quick research, it seems like the best option is to use the [`WatchService`](http://docs.oracle.com/javase/7/docs/api/java/nio/file/WatchService.html), but this is only available from Java 7. If you want to maintain compatibility with Java 6, it looks like one of the options is to use this [`FileMonitor`](http://commons.apache.org/proper/commons-vfs/apidocs/org/apache/commons/vfs2/impl/DefaultFileMonitor.html) instead, but that involves polling the file system instead of receiving events.

brian-brazil:
About half of our users are still using Java 6, so we need to maintained compatibility there. SOmething as simple as checking the mtime on every scrape would be fine.

brian-brazil:
This was done.

PR #90, opened 2016-10-25
Reload configuration when file changes

icetrain:
Another version of "reload config on change" that checks mtime of the configuration file and tries to load the new configuration, it that fails it keeps running on the previous settings.

Solves #29

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -1,6 +1,7 @@
 package io.prometheus.jmx;
 
 import io.prometheus.client.Collector;
+import io.prometheus.client.Counter;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.File;
@@ -26,6 +27,14 @@ import org.yaml.snakeyaml.Yaml;
 import static java.lang.String.format;
 
 public class JmxCollector extends Collector {
+    static final Counter configReloadSuccess = Counter.build()
+      .name("jmx_config_reload_success_total")
+      .help("Number of times configuration have successfully been reloaded.").register();
+
+    static final Counter configReloadFailure = Counter.build()
+      .name("jmx_config_reload_failure_total")
+      .help("Number of times configuration have failed to be reloaded.").register();
+
     private static final Logger LOGGER = Logger.getLogger(JmxCollector.class.getName());
 
     private static class Rule {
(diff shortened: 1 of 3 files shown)


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -39,83 +40,97 @@
...
+    private static class Config {
+      String jmxUrl = "";
+      String username = "";
+      String password = "";
+      boolean lowercaseOutputName;
+      boolean lowercaseOutputLabelNames;
+      List<ObjectName> whitelistObjectNames = new ArrayList<ObjectName>();
+      List<ObjectName> blacklistObjectNames = new ArrayList<ObjectName>();
+      ArrayList<Rule> rules = new ArrayList<Rule>();
+      long lastUpdate = System.currentTimeMillis();
```
brian-brazil: This doesn't feel right. I think you should set this to the file mtime.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -39,83 +40,97 @@
...
+        config = loadConfig((Map<String, Object>)new Yaml().load(yamlConfig));
+    }
+
+    private void reloadConfig() {
+      try {
+        FileReader fr = new FileReader(configFile);
+        Map<String, Object> newYamlConfig = (Map<String, Object>)new Yaml().load(fr);
+        config = loadConfig(newYamlConfig);
+      } catch (Exception e) {
+        LOGGER.severe("Configuration reload failed: " + e.toString());
```
brian-brazil: Maybe a increment a metric here? You can use direction instrumentation.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
brian-brazil: A metric with the file mtime would be a good idea too.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -39,83 +40,97 @@
...
+      }
     }
-    private JmxCollector(Map<String, Object> config) throws MalformedObjectNameException {
-        if(config == null) {  //Yaml config empty, set config to empty map.
-            config = new HashMap<String, Object>();
+
+    private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObjectNameException {
+        Config cfg = new Config();
+
+        if(yamlConfig == null) {  //Yaml config empty, set config to empty map.
```
brian-brazil: Space after //

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -39,83 +40,97 @@
...
     }
+
     public JmxCollector(String yamlConfig) throws MalformedObjectNameException {
-        this((Map<String, Object>)new Yaml().load(yamlConfig));
+        config = loadConfig((Map<String, Object>)new Yaml().load(yamlConfig));
+    }
+
+    private void reloadConfig() {
+      try {
+        FileReader fr = new FileReader(configFile);
```
iggyzap: That's a resource leak.

icetrain:
Update with suggestions, let me know if you would like a rebase.

brian-brazil:
Thanks both.

Issue #89, opened 2016-10-22
ObjectName not found using queryNames


cfrantsen:
I found something strange with the _jboss.as:management-root=server_ mBean, `beanConn.queryNames(null, null)` and `beanConn.getMBeanInfo(mbeanName)` finds it but not `beanConn.queryNames(name, null)`.

This behavior results in the jmx_exporter not finding the mBean if I add _jboss.as:management-root=server_ to the whitelist, the only configuration workaround I have found so far is not using a whitelist at all or to whitelist the whole _jboss.as_ domain but this adds a huge amount of mBeans to the scrape list resulting in 8-10x increased scrape durations (20+ seconds on a server with no deployments or application configuration).

The following change to the scraper code that generates the whitelist seems to solve the problem

```
            for (ObjectName name : whitelistObjectNames) {
                if (name == null || name.toString().contains("*")) {
                  mBeanNames.addAll(beanConn.queryNames(name, null));
                } else {
                  mBeanNames.add(name);
                }
            }
```

brian-brazil:
Per the docs, `jboss.as:management-root=server` should work. Do you know why it isn't?

Your suggested fix is not correct, as it doesn't handle `?` and doesn't allow for the fact that these are patterns on names - not the names themselves.

icetrain:
I'm at a loss as to why that mBean doesn't work with queryNames.

My intention with the above "fix" was to detect patterns and if a bare name was found then just add it to the scrape list and let `scrapeBean()` detect and handle invalid names etc.

brian-brazil:
ObjectNames aren't bare names though, they're patterns, so we can't just use them as mBeanNames.

icetrain:
From reading https://docs.oracle.com/javase/7/docs/api/javax/management/ObjectName.html i gathered that an ObjectName can represent either an object name or an object pattern.

In my case I'm adding an object name (no wildcard) to the whitelist and this is usable as is by `getMBeanInfo()`

brian-brazil:
The problem is we can't distinguish between a pattern, and a mbean name.

icetrain:
Why not? I don't quite follow :-) 

If an ObjectName name contains wildcards it's a pattern, if not ObjectName name == mbean name no?

brian-brazil:
You're correct, I misread the docs.

This still leaves the question as to why the query isn't working.

icetrain:
[...]

PR #91, opened 2016-10-25
Use queryMBeans instead of queryNames.

icetrain:
Solves #89 

No noticable performance impact when testing scrape of a target with about 1800 MBeans.


brian-brazil:
This code should be equivalent, do you know why it's not?

icetrain:
No I'm afraid not, my testing shows that queryMBeans finds the bean I'm looking for but queryNames is not unless I broaden the scope of the query to domain:* and then apply a filter but that's much slower than using queryMBeans.

brian-brazil:
Very odd, can you add a comment pointing to this issue so it doesn't get accidentally reverted in the future?

icetrain:
Updated with comment pointing to #89

Issue #94, opened 2016-11-08
 java agent jmx_exporter port will not be released immediately


eisig:
I use jmx exporter java agent to monitor tomcat. When I shut down tomcat, the jmx_exporter port will not be released immediately;(May be 30s or longer).
So when restarting  tomcat, it will fail.

brian-brazil:
This is how TCP sockets work due to the TIME_WAIT state. It looks like we need to upgrade our jetty to include https://bugs.eclipse.org/bugs/show_bug.cgi?id=461133

brian-brazil:
Unfortunately this upgrade would require dropping Java 6 support, so we're stuck with it this way.

abuwahid007:
We have deployed jmx_prometheus_javaagent-0.3.1.jar and we are observing problem with Jboss 7 with jdk 1.7.  Kindly let me know the fix.

PR #95, opened 2016-11-10
Adding gzip compression (dynamic gzip filter)

Antwnis:
Adding optional configuration `gzip` to enable dynamic gzip compression

In our environment with gzip we noticed network I/O dropping to 1 / 14th 
i.e. 176,270 byte / sec --> 12,801 byte / sec rate

### DIFF
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -30,6 +30,11 @@
       <artifactId>jetty-servlet</artifactId>
       <version>8.1.7.v20120910</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-servlets</artifactId>
+      <version>8.1.7.v20120910</version>
+    </dependency>
   </dependencies>
 
 
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -28,12 +28,12 @@
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-servlet</artifactId>
-      <version>8.1.7.v20120910</version>
+      <version>9.3.14.v20161028</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-servlets</artifactId>
-      <version>8.1.7.v20120910</version>
+      <version>9.3.14.v20161028</version>
     </dependency>
   </dependencies>
 
--- a/jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
+++ b/jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
@@ -5,6 +5,7 @@ import java.io.File;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.servlet.ServletHolder;
+import org.eclipse.jetty.servlets.GzipFilter;
 
 public class WebServer {
    public static void main(String[] args) throws Exception {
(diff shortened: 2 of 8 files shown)


jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
```diff
@@ -18,6 +19,9 @@ public static void main(String[] args) throws Exception {
      Server server = new Server(port);
      ServletContextHandler context = new ServletContextHandler();
      context.setContextPath("/");
+     if (jc.gzipEnabled()) {
```
brian-brazil: Just hardcode it to on

jmx_prometheus_httpserver/pom.xml
```diff
@@ -30,6 +30,11 @@
       <artifactId>jetty-servlet</artifactId>
       <version>8.1.7.v20120910</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-servlets</artifactId>
+      <version>8.1.7.v20120910</version>
```
brian-brazil: While you're here, would you mind upgrading us to something recent to solve #94?

jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
```diff
@@ -5,6 +5,7 @@
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.servlet.ServletHolder;
+import org.eclipse.jetty.servlets.GzipFilter;
 
 public class WebServer {
```
brian-brazil: The java agent (which is the preferred way of doing things) also need changing.

jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
Antwnis: not sure if we should.. i'd say keep it optional - and present it as a recommended setting on the docs.. WDYT?

jmx_prometheus_httpserver/pom.xml
Antwnis: Shall we introduce 0.0.18 version of the io.prometheus dependencies and make a 0.0.18 release of the jmx_exporter ?

Antwnis:
@brian-brazil we are all too busy - but what do u think of the above ?! 
Are you happy with all parts | some parts of this ?!

jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
brian-brazil: If someone wants to turn it off we can consider it, but I'd avoid a knob we don't absolutely have to have.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -34,15 +35,20 @@ public static void premain(String agentArgument, Instrumentation instrumentation
        file = args[1];
      }
 
-     new JmxCollector(new File(file)).register();
+     JmxCollector jc = new JmxCollector(new File(file)).register();
      DefaultExports.initialize();
 
      server = new Server(socket);
+     /*
```
brian-brazil: Debug code?

Antwnis:
Please review - two commits : 
1. Add gzip to the java agent
 
https://github.com/prometheus/jmx_exporter/pull/95/commits/5235373d7dfe983cd4f5eb3b252b0e277794089b

2. Trying to fix #94 

https://github.com/prometheus/jmx_exporter/pull/95/commits/366d0a8aed8e7ff3ab1b9f6bf265bf78504fb891

On that second one - after upgrading to more recent Jetty - we can no longer explicitly set both the QueuedPool AND a Socket - on new version we can set JUST one of them 

So i'm not too sure if this has any side effects
We need to set either of them - and socket takes preference

mzagar:
hi all, any updates on this PR and when it could be merged?
we really need compression - anyway we can help?

brian-brazil:
We've just switched away from jetty so this PR no longer applies, however if you'd like to add transparent gzip support over in client_java's simpleclient_httpserver that'd be welcome.

PR #99, opened 2016-11-29
Issue 98 : Support configuring host address for WebServer exporter setup

marcelmay:
Implements configuring host, including small test.


brian-brazil:
Thanks for the PR, this code should be the same as what's already in the agent.

marcelmay:
@brian-brazil , thx for the fast feedback.

Would you prefer to move the method into shared collector module, as a shared helper class (used by agent and webserver)?

Or should I duplicate the code from agent (and remove the small test, too)?

brian-brazil:
The code is fairly simple, so I think copying it is okay.

marcelmay:
Updated pull request, now WebServer configures hostname like JavaAgent

brian-brazil:
Thanks!

PR #103, opened 2016-12-28
add read and write counts by column family

soccerties:
Read and write counters for individual column families are not included under org.apache.cassandra.metrics.
Added them from org.apache.cassandra.db

@brian-brazil


brian-brazil:
From memory, ReadLatency and WriteLatency cover these. There should be no need to use the older metrics.

soccerties:
You're right. I assumed the metric collected from those was latency but it's only collecting the count attribute.

PR #105, opened 2017-01-24
Added support for hashmaps and objects exposing metrics with best guess

faizann:
This allows support for scraping into Maps and Objects exposing. Custom MBean classes can be  provided in classpath to allow traversing through them.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -1,32 +1,33 @@
 package io.prometheus.jmx;
 
-import java.io.IOException;
-
-import java.lang.management.ManagementFactory;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
-import java.util.HashSet;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import javax.management.JMException;
 import javax.management.MBeanAttributeInfo;
 import javax.management.MBeanInfo;
 import javax.management.MBeanServerConnection;
-import javax.management.ObjectName;
 import javax.management.ObjectInstance;
+import javax.management.ObjectName;
 import javax.management.openmbean.CompositeData;
+import javax.management.openmbean.CompositeType;
 import javax.management.openmbean.TabularData;
 import javax.management.openmbean.TabularType;
-import javax.management.openmbean.CompositeType;
 import javax.management.remote.JMXConnector;
 import javax.management.remote.JMXConnectorFactory;
 import javax.management.remote.JMXServiceURL;
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.logging.Level;
+import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -266,7 +267,51 @@ private void processBeanValue(
                 }
             }
         } else if (value.getClass().isArray()) {
-            logScrape(domain, "arrays are unsupported");
+            logScrape( domain, "arrays are unsupported" );
+        } else if (value instanceof Map ) {
```
brian-brazil: Keep the whitespace consistent with the rest of the file.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -266,7 +267,51 @@ private void processBeanValue(
...
+            } catch ( Exception e ) {
+                logScrape( domain + beanProperties, attrType + "had exception" );
+            }
+        } else if (value instanceof Object ) { // always is going to be true for all objects anyway
+            logger.info( value.getClass().getDeclaredMethods().toString() );
+            for (Method method : value.getClass().getDeclaredMethods() ) {
+                try {
+
+                    String name = method.getName();
+                    if ( name.startsWith( "get" )  || name.startsWith( "is" )) {
```
brian-brazil: I don't think we should automatically call functions, a lot of them are going to change state and/or be slow.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -266,7 +267,51 @@ private void processBeanValue(
...
+                for ( Object key : map.keySet() ) {
+                    attrKeys = new LinkedList<String>(attrKeys);
+                    attrKeys.add(key.toString());
+                    processBeanValue(
+                            domain,
+                            beanProperties,
+                            attrKeys,
+                            key.toString(), // forcing it to be a string
+                            "string",
+                            "hashmap",
```
brian-brazil: map - we don't know if it's a  hashmap

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -266,7 +267,51 @@ private void processBeanValue(
...
+                Map map = ( Map ) value;
+                for ( Object key : map.keySet() ) {
+                    attrKeys = new LinkedList<String>(attrKeys);
+                    attrKeys.add(key.toString());
+                    processBeanValue(
+                            domain,
+                            beanProperties,
+                            attrKeys,
+                            key.toString(), // forcing it to be a string
+                            "string",
```
brian-brazil: We don't know this.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -266,7 +267,51 @@ private void processBeanValue(
...
+            try {
+                Map map = ( Map ) value;
+                for ( Object key : map.keySet() ) {
+                    attrKeys = new LinkedList<String>(attrKeys);
+                    attrKeys.add(key.toString());
+                    processBeanValue(
+                            domain,
+                            beanProperties,
+                            attrKeys,
+                            key.toString(), // forcing it to be a string
```
brian-brazil: It's already in the attrKeys, we shouldn't also have it as the name.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
faizann: That is true. Calling get functions is dangerous. I think this requires specifying beans in config so that user knows what he is doing.

Thanks

brian-brazil:
Did you close this by accident? There's still the hashmap support.

faizann:
Yeah I decided to close it as a whole PR. But indeed hashmap we can still keep.

PR #108, opened 2017-01-28
Added HBase example config

BasPH:
I've added an example configuration for HBase.

### DIFF
--- a/example_configs/hbase.yml
+++ b/example_configs/hbase.yml
@@ -5,6 +5,5 @@ rules:
       namespace: "$1"
       table: "$2"
       region: "$3"
-  - pattern: Hadoop<service=(\w+), name=(\w+), sub=(\w+)><>(\w+)
-    name: HBase_$2_$3_$4
-
+  - pattern: Hadoop<service=HBase, name=(\w+), sub=(\w+)><>([\w.-]+)
+    name: HBase_$1_$2_$3
--- a/example_configs/hbase.yml
+++ b/example_configs/hbase.yml
@@ -1,3 +1,5 @@
+lowercaseOutputLabelNames: true
+lowercaseOutputName: true
 rules:
   - pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_([^\W_]+)_table_([^\W_]+)_region_([^\W_]+)_metric_(\w+)
     name: HBase_$4
--- a/example_configs/hbase.yml
+++ b/example_configs/hbase.yml
@@ -1,0 +1,10 @@
+rules:
+  - pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_([^\W_]+)_table_([^\W_]+)_region_([^\W_]+)_metric_(\w+)
+    name: HBase_metric_$4
+    labels:
+      namespace: "$1"
+      table: "$2"
+      region: "$3"
+  - pattern: Hadoop<service=(\w+), name=(\w+), sub=(\w+)><>(\w+)
+    name: HBase_$2_$3_$4
+
--- a/example_configs/hbase.yml
+++ b/example_configs/hbase.yml
@@ -1,6 +1,6 @@
 rules:
   - pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_([^\W_]+)_table_([^\W_]+)_region_([^\W_]+)_metric_(\w+)
-    name: HBase_metric_$4
+    name: HBase_$4
     labels:
       namespace: "$1"
       table: "$2"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you provide an example of the metrics this produces?

BasPH:
Here's a dump from a regionserver's metrics with this config: http://pastebin.com/8giBiK7X.

brian-brazil:
HBase should be lowercase, "metric" in the name is redundant.

HBase_RegionServer_IPC_exceptions is repeated.

How do the various latency metrics work? I don't recognise that naming scheme.

BasPH:
I've made all metrics lowercase and removed "metric" in the name. The exceptions metric turned out to be named "exceptions.something" and the regex was missing the part after the dot, so I now also match dots and hyphens in the attribute name.

I'm not sure if I understand your latency metrics question. Do you mean what their scheme is? I've dumped all the HBase beans names for you to see: http://pastebin.com/KfE6pUBR.

If there is any, what is the rule of thumb for metrics like these:

1. Hadoop<service=HBase, name=Master, sub=FileSystem><>HlogSplitTime_max
2. Hadoop<service=HBase, name=Master, sub=FileSystem><>HlogSplitTime_mean
3. Hadoop<service=HBase, name=Master, sub=FileSystem><>HlogSplitTime_median
4. Hadoop<service=HBase, name=Master, sub=FileSystem><>HlogSplitTime_min
5. Hadoop<service=HBase, name=Master, sub=FileSystem><>HlogSplitTime_num_ops

Would you expose 5 different metrics or 1 metric with a label and max, mean, etc as label values?

brian-brazil:
> I'm not sure if I understand your latency metrics question. Do you mean what their scheme is? 

I mean I don't recognise the metrics library that's producing them, and I need to look at the source of it to figure out which of those are actually useful. By default I'd presume only the num_ops is useful.

example_configs/hbase.yml
```diff
@@ -0,0 +1,11 @@
+lowercaseOutputLabelNames: true
+lowercaseOutputName: true
+rules:
+  - pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_([^\W_]+)_table_([^\W_]+)_region_([^\W_]+)_metric_(\w+)
+    name: HBase_$4
```
brian-brazil: There's still capital letters here

BasPH:
Had to look into it, this is what I found.

The metrics in HBase are implemented using the Hadoop Metrics2 framework. All the min, max, mean, etc. metrics are implemented by the [MetricHistogram interface](https://github.com/apache/hbase/blob/master/hbase-hadoop-compat/src/main/java/org/apache/hadoop/metrics2/MetricHistogram.java).

I checked how the hlogSplitTime metric is computed and registered:

1. [splittime is calculated](https://github.com/apache/hbase/blob/ae21797305188e82f7017ce5675e4fde950461aa/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterWalManager.java#L323)
2. [splittime is added](https://github.com/apache/hbase/blob/ae21797305188e82f7017ce5675e4fde950461aa/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterWalManager.java#L329)
3. the call passes through a couple of classes and the [method doing the add is in FastLongHistogram](https://github.com/apache/hbase/blob/c64a1d199402d6bf2d6ff4168c00c756dcaa59e4/hbase-metrics/src/main/java/org/apache/hadoop/hbase/metrics/impl/FastLongHistogram.java#L116)

The actual splittime value itself is continuously added to this Histogram object and when a call is made for a sample of the histogram's values, all the percentiles/ranges/etc are calculated. The num_ops value is the number of samples added to the histogram. I would have liked the last registered splittime value itself exposed as well. Therefore, I find the histogram values not that useful for exporting to Prometheus.

brian-brazil:
Okay, I haven't seen this one before.

Looking at the source as I suspected only _num_ops is useful. All the rest are calculated since the process started, so should be dropped.

I'd suggest a feature request to upstream to expose the `total` field.

qgxiaozhan:
Hi ,

Region pattern would not be match if table name has  '_' , such as "aaa_bb_cc".

```
 pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_([^\W_]+)_table_([^\W_]+)_region_([^\W_]+)_metric_(\w+)
```
I think this will be better!
```
pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_(.*)_table_(.*)_region_(.*)_metric_(\w+)
```

brian-brazil:

PR #110, opened 2017-02-12
mvn package also creates a debian package for the agent

xiu:


brian-brazil:
Thanks for the PR. To be honest I'm not a big fan of the deb packages (which predate my maintenance of this package) as they tend to lead to lots of PRs by users trying to make it work for their specific setup and encouraging the httpserver over the agent. I can see the use in offering the agent as you're adding here though. This is also the only component of Prometheus that offers debs.

@TheTincho Do you know if there's any likelihood that the jmx exporter will make it into Debian?

TheTincho:
@brian-brazil Definitely!

I just did not look into it due to a lack of need (and disgust of java :)), but I can give it a try.

Is there a way to build without maven? It seems that now debian can build with maven using some specific hacks, but I'd really prefer something lighter..

brian-brazil:
You don't have to build with maven, you could build by hand but you'd be on your own for that.

TheTincho:
OK, i will take a look this week

brian-brazil:
Per above, I'd prefer not expand the deb here.

PR #114, opened 2017-02-17
Allow JMX scraping over SSL

msvab:
Hello @brian-brazil ,

This PR enables JMX scraping over SSL. If you have any issues with this, please let me know.

Cheers,
Michal

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -67,13 +67,15 @@ public class JmxScraper {
     private String jmxUrl;
     private String username;
     private String password;
+    private boolean ssl;
     private List<ObjectName> whitelistObjectNames, blacklistObjectNames;
 
-    public JmxScraper(String jmxUrl, String username, String password, List<ObjectName> whitelistObjectNames, List<ObjectName> blacklistObjectNames, MBeanReceiver receiver) {
+    public JmxScraper(String jmxUrl, String username, String password, boolean ssl, List<ObjectName> whitelistObjectNames, List<ObjectName> blacklistObjectNames, MBeanReceiver receiver) {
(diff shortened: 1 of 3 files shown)


brian-brazil:
How would a user configure this?

A trust store existing at the system level doesn't mean SSL is desired.

msvab:
Good point, I didn't think there would be other things running in that JVM, but that's not necessarily true. I'll add a way to configure it.

brian-brazil:
System properties can be configured at the machine level.

It's recommenced to use the jmx exporter as a java agent, it bypasses all this.

msvab:
I've added a ssl option to the yaml config. I was also hoping to add configuration for the ssl keystore and truststore, but after looking through the JMX api I've discovered that there is no way to provide the keystore and truststore programmatically and so it has to be configured with system properties.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -86,14 +91,19 @@ public void doScrape() throws Exception {
...
-          HashMap credential = null;
+          Map<String, Object> environment = new HashMap<String, Object>();
           if(username != null && username.length() != 0 && password != null && password.length() != 0) {
-            credential = new HashMap();
             String[] credent = new String[] {username, password};
-            credential.put(javax.management.remote.JMXConnector.CREDENTIALS, credent);
-          }       
+            environment.put(javax.management.remote.JMXConnector.CREDENTIALS, credent);
+          }
+          if(ssl) {
```
brian-brazil: space before (

brian-brazil:
This also needs docs

msvab:
I've documented the new config option. Is this sufficient?

brian-brazil:
Thanks!

msvab:
Thanks for such a fast turnaround!

PR #118, opened 2017-02-28
bump bash script to use 0.9

rafabene:
master branch uses 0.9-SNAPSHOT so the bash script should be updated


brian-brazil:
Thanks!

PR #123, opened 2017-03-15
Bump max threads to resolve #122

### LINKED ISSUE #122: Won't run on machine with more than 20 CPU cores
I'm using the JMX Exporter to monitor Apache Spark, using it as a Java agent. This works fine until I try and run it on a server with more than 20 cores. In the spark logs I see the following line as spark is starting up:

`2017-03-15 21:35:50.606:WARN:ipjsoejs.AbstractConnector:insufficient threads configured for SelectChannelConnector@0.0.0.0:19105
`

Then once spark is running the log gets spammed with lines like the following:

`2017-03-15 21:35:53.858:WARN:ipjsoeji.nio:Dispatched Failed! SCEP@3712cd1e{l(/192.168.92.52:44556)<->r(/192.168.76.1:19105),d=false,open=true,ishut=false,oshut=false,rb=false,wb=false,w=true,i=0}-{AsyncHttpConnection@905c2db,g=HttpGenerator{s=0,h=-1,b=-1,c=-1},p=HttpParser{s=-14,l=0,c=0},r=0} to io.prometheus.jmx.shaded.org.eclipse.jetty.server.nio.SelectChannelConnector$ConnectorSelectorManager@5f9d5ea6
2017-03-15 21:35:53.859:WARN:ipjsoeji.nio:Dispatched Failed! SCEP@3712cd1e{l(/192.168.92.52:44556)<->r(/192.168.76.1:19105),d=false,open=true,ishut=false,oshut=false,rb=false,wb=false,w=true,i=1r}-{AsyncHttpConnection@905c2db,g=HttpGenerator{s=0,h=-1,b=-1,c=-1},p=HttpParser{s=-14,l=0,c=0},r=0} to io.prometheus.jmx.shaded.org.eclipse.jetty.server.nio.SelectChannelConnector$ConnectorSelectorManager@5f9d5ea6
2017-03-15 21:35:53.859:WARN:ipjsoeji.nio:Dispatched Failed! SCEP@3712cd1e{l(/192.168.92.52:44556)<->r(/192.168.76.1:19105),d=false,open=true,ishut=false,oshut=false,rb=false,wb=false,w=true,i=1r}-{AsyncHttpConnection@905c2db,g=HttpGenerator{s=0,h=-1,b=-1,c=-1},p=HttpParser{s=-14,l=0,c=0},r=0} to io.prometheus.jmx.shaded.org.eclipse.jetty.server.nio.SelectChannelConnector$ConnectorSelectorManager@5f9d5ea6`

Not only is the exporter not working in this situation but spark isn't running correctly either and my disk is filling up very quickly!

To reproduce this is simple. Use a machine with >20 cores, add the JMX exporter as a java agent to any java process and watch the logs. You can prove the 20 cores thing by disabling cores until you have <=20 by doing the following (on linux):

`echo 0 > /sys/devices/system/cpu/cpuX/online`

rathers:
Resolves #122 for my use case.

I don't fully understand why this works but it does seem to! The bit that confuses me is that the thread limit of 10 seems to kick in when the cores are >20, rather than >10. The server I'm testing this on has 12 cores plus hyper-threading so the OS presents 24 cores, so maybe the HT is creating the x2 difference.

Obviously this isn't a general fix but it works for my use-case so thought I'd contribute in case it's useful

### DIFF
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -40,7 +40,7 @@ public class JavaAgent {
      server = new Server(socket);
      QueuedThreadPool pool = new QueuedThreadPool();
      pool.setDaemon(true);
-     pool.setMaxThreads(10);
+     pool.setMaxThreads(20);
      pool.setMaxQueued(10);
      pool.setName("jmx_exporter");
      server.setThreadPool(pool);


brian-brazil:
This would allow up to 20 simultaneous scrapes to run, which we're trying to avoid so as to put some limit on the performance impact of the exporter.

rathers:
As I say, it fixes #122 for my use-case and gives an example workaround. Someone cleverer than I can hopefully work out how to resolve this "properly" :)

rathers:
My guess is that there is something within jetty that is insisting on the no. threads being proportional to the no. cores and that the limit of 10 threads, as set in the JMX Exporter, is normally fine. When you end up on a big machine jetty then insists on having more than 10 threads, the limit prevents it from doing so and it gets upset. Total guess though, haven't managed to prove it yet.

brian-brazil:
This was fixed in another PR

PR #126, opened 2017-03-21
Document `/metrics` endpoint in README

Starefossen:
Documents the `/metrics` endpoint in README – I spent some time figuring out this. Feel free to discard if this is common knowledge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,14 @@ To bind the java agent to a specific IP change the port number to `host:port`.
 
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
 
+## Vieweing
+
+Metrics are available from the `/metrics` URL endpoint:
+
+```
+http://localhost:1234/metrics
+```
+
 ## Building
 
 `mvn package` to build.


brian-brazil:
/metrics is the standard Prometheus endpoint, I don't think there's a need to call it out in a separate section. A quick link near the running instructions might make sense though.

README.md
```diff
@@ -21,6 +21,14 @@ To bind the java agent to a specific IP change the port number to `host:port`.
 
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
 
+## Vieweing
```
r4j4h: I think `Vieweing` should be `Viewing`

IRobL:
What do you put for the `-jar` parameter?  I have a kafka instance that is exporting JMX data on a port, but I can't figure out how to convert that JMX stuff into the usual `/metrics` http endpoint.  Is it possible to go from the JMX data that I can view in jconsole into the http stuff?  It would be ideal if I didn't need to point the converter to a JAR because the JMX data is being exported from a docker container almost as if it were on a remote computer.

wimsymons:
Great improvement. I spent 30 minutes today looking for the same information...

brian-brazil:

PR #127, opened 2017-03-27
Fix order of example rules and add more labels

zeisss:
This fixes the example kafka rules to parse all (on my system) mbeans and move all dynamic fields like brokerhost, port, thread ids etc into labels.

I also reversed the order of two rules, as the first matching pattern wins and the second rule was more concrete.

ping @brian-brazil

### DIFF
--- a/example_configs/kafka-0-8-2.yml
+++ b/example_configs/kafka-0-8-2.yml
@@ -22,7 +22,7 @@ rules:
 - pattern : kafka.network<type=(.+), name=(\w+), networkProcessor=(.+)><>Count
   name: kafka_network_$1_$2
   labels:
-    networkProcessor: "$3"
+    request: "$3"
   type: COUNTER
 - pattern : kafka.network<type=(.+), name=(\w+), request=(\w+)><>Count
   name: kafka_network_$1_$2
--- a/example_configs/kafka-0-8-2.yml
+++ b/example_configs/kafka-0-8-2.yml
@@ -33,22 +33,36 @@ rules:
 - pattern : kafka.server<type=(.+), name=(.+)PerSec\w*><>Count
   name: kafka_server_$1_$2_total
   type: COUNTER
-- pattern : kafka.server<type=(.+), name=(.+), topic=(.+)><>(Count|Value)
+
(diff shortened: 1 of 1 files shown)


brian-brazil:
What's the cardinality of these new metrics?

It'd probably be wise to tighten up the regexes, so overmatching is avoided.

zeisss:
Hi Brian,

here is an example of the change:

Before:
```
# HELP kafka_server_fetcherlagmetrics_consumerlag_clientid_replicafetcherthread_0_2 Attribute exposed for management (kafka.server<type=FetcherLagMetrics, name=ConsumerLag, clientId=ReplicaFetcherThread-0-2, topic=listing-mirror, partition=20><>Value)
# TYPE kafka_server_fetcherlagmetrics_consumerlag_clientid_replicafetcherthread_0_2 counter
kafka_server_fetcherlagmetrics_consumerlag_clientid_replicafetcherthread_0_2{topic="category.dev1, partition=28",} 0.0
<353 more metrics with same name>
```

This is repeated 6 times with different suffix (0_1, 0_2, 1_1, 1_2, 2_1, 2_2). Notice the partition is part of the `topic` label due to overmatch.

Now:
```
# HELP kafka_server_fetcherlagmetrics_consumerlag Attribute exposed for management (kafka.server<type=FetcherLagMetrics, name=ConsumerLag, clientId=ReplicaFetcherThread-0-2, topic=listing-mirror, partition=20><>Value)
# TYPE kafka_server_fetcherlagmetrics_consumerlag gauge
kafka_server_fetcherlagmetrics_consumerlag{clientId="ReplicaFetcherThread-0-2",partition="20",topic="listing-mirror",} 0.0
<2352 more of this metric)
```

Similar things happened with other metrics.

example_configs/kafka-0-8-2.yml
```diff
@@ -22,8 +22,12 @@ rules:
 - pattern : kafka.network<type=(.+), name=(\w+), networkProcessor=(.+)><>Count
   name: kafka_network_$1_$2
   labels:
-    request: "$3"
+    networkProcessor: "$3"
```
brian-brazil: this was request on purpose

zeisss: Reverted

brian-brazil:
Thanks!

PR #130, opened 2017-04-03
Change metrics path in the httpserver.

sorobon:
We need to change the metrics path in the httpserver.
I changed the way you call server using apache commons-cli lib.
In current version must provide arguments in a specific order to call server
```bash
java -jar jmx_prometheus_httpserver.jar 5556 example_configs/httpserver_sample_config.yml
```

In new version you can provide the port, the hostname, the config file, the metrics path, the verbose mode and a flag to show usage. There is no specific order in arguments.
It will be something like this:
```bash
java -jar jmx_prometheus_httpserver.jar -p,--port <5556> -c,--config <example_configs/httpserver_sample_config.yml> [-hostname <arg>] [-path <arg>] [-v] [-h, --help]
``` 
Output when help flag is used:
```bash
 -c,--config <arg>   Config file path
 -h,--help           Show help message
 -hostname <arg>     Hostname
 -p,--port <arg>     Port
 -path <arg>         Path
 -v                  Verbose mode
```

Of course legacy way is fully supported.

@brian-brazil


brian-brazil:
You seems to be adding a lot of code and breaking changes here. Can you explain what you're trying to do?

sorobon:
Sorry, i forgot description :/

brian-brazil:
Why do you want to change the metrics path?

sorobon:
I want to homogenize metrics path. I use the prometheus java client lib with spring boot and it forces to use /prometheus path instead of the default /metrics and I think register several combinations of port/path for the same host is not maintainable.
The prometheus server provide this functionality (configure path).

https://github.com/prometheus/client_java/blob/master/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusEndpoint.java#L21

brian-brazil:
Standardising on anything other than /metrics is unlikely to be very practical, as you're relying on all 150+ integrations to support a non-standard path.

sorobon:
Default and standard path is /metrics but the possibility to change it i think its very useful.

PR #131, opened 2017-04-04
Fix bug that prevented JMX auth from being used in standalone scraper

wulczer:


brian-brazil:
Thanks!

This main function is just for development/debugging.

PR #135, opened 2017-04-19
Add Apache flink example.

mindprince:
We have this config running in production. We use Flink 1.1.4

### DIFF
--- a/example_configs/flink.yml
+++ b/example_configs/flink.yml
@@ -18,7 +18,6 @@ rules:
   - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Count
     name: flink_taskmanager_$6_total
     labels:
-      host: $1
       taskmanager_id: $2
       flink_job: $3
       operator: $4
@@ -29,7 +28,6 @@ rules:
   - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Value
     name: flink_taskmanager_$6
     labels:
-      host: $1
       taskmanager_id: $2
       flink_job: $3
       operator: $4
@@ -39,7 +37,6 @@ rules:
   - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), key6=KafkaProducer, name=([a-z].*)><>Value
     name: flink_taskmanager_kafka_producer_$6
     labels:
-      host: $1
       taskmanager_id: $2
       flink_job: $3
       operator: $4
@@ -49,7 +46,6 @@ rules:
   - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), key6=KafkaConsumer, key7=(.*), name=([a-z].*)><>Value
     name: flink_taskmanager_kafka_consumer
     labels:
-      host: $1
       taskmanager_id: $2
       flink_job: $3
       operator: $4
--- a/example_configs/flink.yml
+++ b/example_configs/flink.yml
@@ -1,0 +1,58 @@
+rules:
+  ## Job Manager ##
+  # Example: org.apache.flink.metrics<key0=127.0.0.1, key1=jobmanager, name=taskSlotsTotal><>Value
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=jobmanager, name=([a-z].*)><>Value
+    name: flink_jobmanager_$2
+    labels:
+      ip: $1
+
+  # Example: org.apache.flink.metrics<key0=127.0.0.1, key1=jobmanager, key2=My_Job, name=lastCheckpointDuration><>Value
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=jobmanager, key2=(.*), name=([a-z].*)><>Value
+    name: flink_jobmanager_$3
+    labels:
+      ip: $1
+      job: $2
+
+  ## Task Manager ##
+  # Example: org.apache.flink.metrics<key0=debian-8, key1=taskmanager, key2=407401ecfef1f050d3dcc83af9cbfb74, key3=My_Job, key4=Map, key6=1, name=numRecordsIn><>Count
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Count
+    name: flink_taskmanager_$6
+    labels:
+      host: $1
+      taskmanager_id: $2
+      job: $3
+      operator: $4
+      partition: $5
+    type: COUNTER
+
+  # Example: org.apache.flink.metrics<key0=debian-8, key1=taskmanager, key2=407401ecfef1f050d3dcc83af9cbfb74, key3=My_Job, key4=fold_->_sink, key6=0, name=lastCheckpointSize><>Value
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Value
+    name: flink_taskmanager_$6
+    labels:
+      host: $1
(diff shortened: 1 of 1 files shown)


example_configs/flink.yml
```diff
@@ -0,0 +1,58 @@
...
+      job: $2
+
+  ## Task Manager ##
+  # Example: org.apache.flink.metrics<key0=debian-8, key1=taskmanager, key2=407401ecfef1f050d3dcc83af9cbfb74, key3=My_Job, key4=Map, key6=1, name=numRecordsIn><>Count
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Count
+    name: flink_taskmanager_$6
+    labels:
+      host: $1
+      taskmanager_id: $2
+      job: $3
```
brian-brazil: job has a meaning already in Prometheus, this needs another name

example_configs/flink.yml
```diff
@@ -0,0 +1,58 @@
...
+      ip: $1
+      job: $2
+
+  ## Task Manager ##
+  # Example: org.apache.flink.metrics<key0=debian-8, key1=taskmanager, key2=407401ecfef1f050d3dcc83af9cbfb74, key3=My_Job, key4=Map, key6=1, name=numRecordsIn><>Count
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Count
+    name: flink_taskmanager_$6
+    labels:
+      host: $1
+      taskmanager_id: $2
```
brian-brazil: What's the cardinality of this like?

example_configs/flink.yml
```diff
@@ -0,0 +1,58 @@
...
+    labels:
+      ip: $1
+      job: $2
+
+  ## Task Manager ##
+  # Example: org.apache.flink.metrics<key0=debian-8, key1=taskmanager, key2=407401ecfef1f050d3dcc83af9cbfb74, key3=My_Job, key4=Map, key6=1, name=numRecordsIn><>Count
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Count
+    name: flink_taskmanager_$6
+    labels:
+      host: $1
```
brian-brazil: what is this label exactly?

example_configs/flink.yml
```diff
@@ -0,0 +1,58 @@
...
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=jobmanager, key2=(.*), name=([a-z].*)><>Value
+    name: flink_jobmanager_$3
+    labels:
+      ip: $1
+      job: $2
+
+  ## Task Manager ##
+  # Example: org.apache.flink.metrics<key0=debian-8, key1=taskmanager, key2=407401ecfef1f050d3dcc83af9cbfb74, key3=My_Job, key4=Map, key6=1, name=numRecordsIn><>Count
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Count
+    name: flink_taskmanager_$6
```
brian-brazil: these are counters, so should be suffixed with _total

example_configs/flink.yml
```diff
@@ -0,0 +1,58 @@
...
+  - pattern: org.apache.flink.metrics<key0=(.*), key1=taskmanager, key2=(.*), key3=(.*), key4=(.*), key5=([0-9]*), name=([a-z].*)><>Value
+    name: flink_taskmanager_$6
+    labels:
+      host: $1
+      taskmanager_id: $2
+      job: $3
+      operator: $4
+      partition: $5
+
+  # Example: org.apache.flink.metrics<key0=debian-8, key1=taskmanager, key2=407401ecfef1f050d3dcc83af9cbfb74, key3=My_Job, key4=Sink, key5=1, key6=KafkaProducer, name=batch-size-avg><>Value
```
brian-brazil: Are these the standard kafka metrics, just from a different mbean to normal?

example_configs/flink.yml
mindprince: Updated.

example_configs/flink.yml
mindprince: cardinality = equal to the number of taskmanagers running. Usually there's one taskmanager per node. When you restart the taskmanager (which happens rarely), the id changes.

example_configs/flink.yml
mindprince: Updated.

example_configs/flink.yml
brian-brazil: Is a taskmaster one process?

This sounds like a duplicate of the instance label.

example_configs/flink.yml
mindprince: key0 in taskmanager metrics is set to the hostname on which taskmanager is running.
So, on my dev environment it's set to `debian-8` (which is also the output of `hostname`)
On one of the prod nodes, it's set to the `hostname`.

brian-brazil: That shouldn't be exposed then, as this is already covered by the instance label.

mindprince: It's similar to the instance label. The instance label is usually like "host.example.com:PORT". This is just "host". I can remove it if you really don't want it.

example_configs/flink.yml
mindprince: Yeah, the taskmanager is one process. I would like to keep it, this helps in telling when the taskmanager was restarted.

example_configs/flink.yml
mindprince: When your flink job uses kafka sources or sinks, the kafka metrics are exposed like this.

example_configs/flink.yml
brian-brazil: It duplicates the role of the instance label, and it's up to the Prometheus user to decide what they want their instance label to be (which may not be the hostname).

mindprince: Removed host label.

example_configs/flink.yml
brian-brazil: This should be removed then, it's not helping to distinguish the metrics.

Any time the same label is being applied to an entire target from the /metrics page is a smell, and it rarely makes sense to keep it.

Is `ip` also one of those?

`process_start_time_seconds` from the standard exports will cover you for restarts.

mindprince: Done.

brian-brazil:
Thanks!

thiszyq:
Hi, i want test the JMX_exporter in flink at local ubuntu. 
I haved start a prometheus server successed and start flink successed by **/bin/start-local.sh**. and i have download **jmx_prometheus_javaagent-0.9.jar** in my flink folder, and **flink-config.yml** is [example_configs/flink.yml](https://github.com/prometheus/jmx_exporter/blob/master/example_configs/flink.yml)
my prometheus.yml as follow:
```
global:
 scrape_interval: 10s
 evaluation_interval: 10s
scrape_configs:
 - job_name: 'flink'
   static_configs:
    - targets:
      - localhost:8001
```
what command should i input in bash so that i could get flink-metrics by wget http://localhost:8001/metrics 
Thank you in advance for any help you can provide.

mindprince:
In your `flink-conf.yaml`, add
```
env.java.opts: -javaagent:/path/to/jmx_prometheus_javaagent-0.9.jar=7076:/path/to/example_configs/flink.yml
```
Then when you start the flink daemons:
The taskmanager metrics would be on http://taskmanager_host:7076/metrics
The jobmanager metrics would be on http://jobmanager_host:7076/metrics

thiszyq:
thanks 

I had operated as your said ,then use *./bin/start-local.sh* to start flink daemon. but it not work, cant get metrics from http://jobmanager_host:port/metrics. ,also taskmanager-metrics.

but  could get jobmanager-metrics by add **$JVM_ARGS** in /bin/jobmanager.sh 

`export JVM_ARGS="$JVM_ARGS -Xms"$FLINK_JM_HEAP"m -Xmx"$FLINK_JM_HEAP"m -javaagent:$PWD/jmx_prometheus_javaagent-0.9.jar=9001:$PWD/flink.yml"`
@mindprince

mindprince:
I haven't used local mode in Flink.

Basically, both the taskmanager and jobmanager JVM should know about the javaagent. If they are running on the same host, use different ports.

PR #137, opened 2017-04-25
Add gzip compression and bump version on jetty

erebe:
Mainly the same as #95 but without conflicts

mvn integration-test is ok and I use the standalone http server with gzip activated in production

### DIFF
--- a/jmx_prometheus_javaagent/pom.xml
+++ b/jmx_prometheus_javaagent/pom.xml
@@ -33,8 +33,13 @@
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-servlet</artifactId>
-      <version>8.1.7.v20120910</version>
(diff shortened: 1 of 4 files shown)


jmx_prometheus_javaagent/pom.xml
```diff
@@ -33,14 +33,19 @@
...
-      <version>8.1.7.v20120910</version>
+      <version>9.4.3.v20170317</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
```
brian-brazil: This should go above the test dependency

erebe: done

erebe:
Pom updated

brian-brazil:
Thanks!

PR #139, opened 2017-04-26
Gzip compression fixing my mistakes

erebe:
Fixing my mistakes for the gzip compression with the java agent.
My apologies for not having tested more the first time, I was lured by the test.

 + Don't create an InetSocketAddress as it will be created by the connector
 + Ask jetty to shutdown properly when the app stop
 + Fix integration test, as now jetty keeps the connection alive
 + Fail integration test if  scraping goes wrong (ie: connection refused)

P.s: The Systemd.exit(0) in the TestApplication is needed only when the app is run with maven. When launched directly with the javaagent it will stop gracefully without the systemd.exit(0)

### DIFF
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -4,7 +4,7 @@ import io.prometheus.client.exporter.MetricsServlet;
 import io.prometheus.client.hotspot.DefaultExports;
 import java.lang.instrument.Instrumentation;
 import java.io.File;
-import java.net.InetSocketAddress;
+import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.servlet.ServletContextHandler;
@@ -22,17 +22,17 @@ public class JavaAgent {
        System.exit(1);
      }
 
+     String host;
      int port;
-     InetSocketAddress socket;
      String file;
 
      if (args.length == 3) {
        port = Integer.parseInt(args[1]);
-       socket = new InetSocketAddress(args[0], port);
+       host = args[0];
        file = args[2];
      } else {
        port = Integer.parseInt(args[0]);
-       socket = new InetSocketAddress(port);
+       host = "0.0.0.0";
        file = args[1];
      }
 
@@ -45,16 +45,18 @@ public class JavaAgent {
      pool.setName("jmx_exporter");
      server = new Server(pool);
 
-     ServerConnector cnx = new ServerConnector(server);
-     cnx.setReuseAddress(true);
-     cnx.setHost(socket.getHostName());
-     cnx.setPort(port);
+     ServerConnector connector = new ServerConnector(server);
+     connector.setReuseAddress(true);
+     connector.setHost(host);
+     connector.setPort(port);
+     server.setConnectors(new Connector[]{connector});
 
      ServletContextHandler context = new ServletContextHandler();
      context.setContextPath("/");
      context.addFilter(GzipFilter.class, "/*", null);
      server.setHandler(context);
      context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
+
      server.start();
    }
 }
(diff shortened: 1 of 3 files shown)


jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -22,17 +22,17 @@ public static void premain(String agentArgument, Instrumentation instrumentation
...
 
      if (args.length == 3) {
        port = Integer.parseInt(args[1]);
-       socket = new InetSocketAddress(args[0], port);
+       host = args[0];
        file = args[2];
      } else {
        port = Integer.parseInt(args[0]);
-       socket = new InetSocketAddress(port);
+       host = "localhost";
```
brian-brazil: This should be empty, otherwise it'll only bind to localhost.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -45,16 +45,19 @@ public static void premain(String agentArgument, Instrumentation instrumentation
...
+     connector.setPort(port);
+     server.setConnectors(new Connector[]{connector});
 
      ServletContextHandler context = new ServletContextHandler();
      context.setContextPath("/");
      context.addFilter(GzipFilter.class, "/*", null);
      server.setHandler(context);
      context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
+
+     server.setStopAtShutdown(true);
```
brian-brazil: I'm not sure we should be setting this. If a server takes a long time to shutdown this would prevent monitoring it during that time.

erebe: Your call, as you know better than me the behavior of prometheus if the connection is closed in the middle of a scraping.
This is implementation details but there is a default timeout of 30sec before forcing the shutdown. We can also lower that value with server.setStopTimeout() to something like 5-10sec.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
erebe: Will change

erebe:
I updated the PR and set a stop timeout of 5 sec, tell me if you prefer to remove it completly

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
brian-brazil: It should be removed completely, this should keep running for as long as possible.

jmx_prometheus_javaagent/src/test/java/io/prometheus/jmx/JavaAgentIT.java
```diff
@@ -87,11 +90,15 @@ public void agentLoads() throws IOException, InterruptedException {
...
 
             // Tell application to stop
+            cnx.disconnect();
             app.getOutputStream().write('\n');
             try {
                 app.getOutputStream().flush();
             } catch (IOException ignored) {
             }
+        } catch (Throwable ex) {
+            assertThat("Expection catched during scraping", false);
```
brian-brazil: caught

erebe: done

erebe:
Updated

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -22,17 +22,17 @@ public static void premain(String agentArgument, Instrumentation instrumentation
...
 
      if (args.length == 3) {
        port = Integer.parseInt(args[1]);
-       socket = new InetSocketAddress(args[0], port);
+       host = args[0];
        file = args[2];
      } else {
        port = Integer.parseInt(args[0]);
-       socket = new InetSocketAddress(port);
+       host = "0.0.0.0";
```
brian-brazil: What does this do on a v6 only machine host?

erebe: An empty string binds to the loopback interface.

I don't have an IPv6 only host, but on my machine it correctly binds on ipv4 and ipv6 interfaces.
as stated here https://github.com/eclipse/jetty.project/blob/0c8273f2ca1f9bf2064cd9c4c939d2546443f759/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractNetworkConnector.java#L53

I also tried the special string "::" which is the quivalent of 0.0.0.0 for ipv6 and it does the same behavior

brian-brazil:
Thanks!

PR #140, opened 2017-04-28
Allow to cache results of scraping and expose metrics true timestamp

erebe:
I made a version of the standalone http server that is able to cache results of scraping in order to avoid hammering the service behind (I mostly use it with cassandra) while still exposing the true timestamp of the scraping. I am aware of the staleness issue and to not go beyond 5min in the past.

I don't expect for this PR to be merged as is, I just release it for other people to be able to grab it if needed. 

Anyway let me know if you want to salvage something from this, most of the work is on MetricsServletCache.java

### DIFF
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -46,8 +46,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.5</source>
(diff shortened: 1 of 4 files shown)


brian-brazil:
> I am aware of the staleness issue and to not go beyond 5min in the past.

This will likely get more complicated with 2.0, the recommendation not to cache and not to expose timestamps for general scraping will get stronger due to improved staleness semantics.

This is not something that generally makes sense to do, even with something as slow as JMX.

PR #143, opened 2017-05-08
[Feature] configurable startup delay

### LINKED ISSUE #97: [Feature Request] Add optional parameter to delay the agent startup
This would help with applications having trouble being queried by Prometheus during their startup process.

### LINKED ISSUE #128: JMX agent metrics from Wildfly/Undertow broken since v0.8
I recently started investigating how to use the `jmx_exporter` as a `-javaagent` with Wildfly 10.1.0.Final and Undertow 3.1. However, none of the Wildfly/Undertow specific metrics was picked up by `jmx_exporter` when using version 0.8 or the latest 0.9.

With some help from @cfrantsen in #87 we discovered that `jmx_exporter ` v0.7 is working correctly, suggesting that somewhere between v0.7 and v0.8 is the breaking change.

I have made a minimal proof of concept for this at [Starefossen/docker-wildfly-prometheus](https://github.com/Starefossen/docker-wildfly-prometheus) if anyone is interested in debugging this issue.

bdellegrazie:
@brian-brazil:
This is a first pass implementation of configurable startup delay for the JVM Agent.
It is intended to resolve #97 and #128

A few notes:
* There are no tests at this time (I will correct this shortly)
* There is no documentation for the property (coming)
* I had to implement the _Describable_ interface in the _Collector_ and return an empty list, breaking current behaviour which 'auto describes' JMX beans. This is only because the act of registering the JMX Collector results in a call to _collect_ as a side-effect. In my particular case it was important not to call _ManagementFactory.getPlatformMBeanServer_ before the delay expires (this is what resolves #128).

The way the code is currently structured, configuration is done at startup time and it is only the Agent Web Server that is started after the configured delay. Alternately it is possible to leave the _Collector_ as is and simply delay all configuration until the the timeout expires. This keeps existing behaviour and I'm happy to change if requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ See `./run_sample_httpserver.sh` for a sample script that runs the httpserver ag
 The configuration is in YAML. An example with all possible options:
 ```yaml
 ---
+startDelaySecs: 0
 hostPort: 127.0.0.1:1234
 jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:1234/jmxrmi
 ssl: false
@@ -48,6 +49,7 @@ rules:
 ```
 Name     | Description
 ---------|------------
+startDelaySecs | start delay before serving requests. Any requests within the delay period will result in an empty metrics set.
 hostPort | The host and port to connect to via remote JMX. If neither this nor jmxUrl is specified, will talk to the local JVM.
 jmxUrl   | A full JMX URL to connect to. Should not be specified if hostPort is.
 ssl      | Whether JMX connection should be done over SSL. To configure certificates you have to set following system properties:<br/>`-Djavax.net.ssl.keyStore=/home/user/.keystore`<br/>`-Djavax.net.ssl.keyStorePassword=changeit`<br/>`-Djavax.net.ssl.trustStore=/home/user/.truststore`<br/>`-Djavax.net.ssl.trustStorePassword=changeit`
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -26,7 +26,7 @@ import org.yaml.snakeyaml.Yaml;
 
 import static java.lang.String.format;
 
-public class JmxCollector extends Collector {
+public class JmxCollector extends Collector implements Collector.Describable {
     static final Counter configReloadSuccess = Counter.build()
       .name("jmx_config_reload_success_total")
       .help("Number of times configuration have successfully been reloaded.").register();
@@ -50,6 +50,7 @@ public class JmxCollector extends Collector {
     }
 
     private static class Config {
+      Integer startDelaySeconds = 0;
       String jmxUrl = "";
       String username = "";
       String password = "";
@@ -64,6 +65,7 @@ public class JmxCollector extends Collector {
 
     private Config config;
     private File configFile;
+    private long createTimeNanoSecs = System.nanoTime();
 
     private static final Pattern snakeCasePattern = Pattern.compile("([a-z0-9])([A-Z])");
 
@@ -106,6 +108,13 @@ public class JmxCollector extends Collector {
           yamlConfig = new HashMap<String, Object>();
         }
 
+        if (yamlConfig.containsKey("startDelaySeconds")) {
+          try {
+            cfg.startDelaySeconds = (Integer) yamlConfig.get("startDelaySeconds");
+          } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Invalid number provided for startDelaySeconds", e);
+          }
+        }
         if (yamlConfig.containsKey("hostPort")) {
           if (yamlConfig.containsKey("jmxUrl")) {
             throw new IllegalArgumentException("At most one of hostPort and jmxUrl must be provided");
@@ -398,6 +407,10 @@ public class JmxCollector extends Collector {
       JmxScraper scraper = new JmxScraper(config.jmxUrl, config.username, config.password, config.ssl, config.whitelistObjectNames, config.blacklistObjectNames, receiver);
       long start = System.nanoTime();
       double error = 0;
+      if ((config.startDelaySeconds > 0) &&
+        ((start - createTimeNanoSecs) / 1000000000L < config.startDelaySeconds)) {
+        throw new IllegalStateException("JMXCollector waiting for startDelaySeconds");
+      }
       try {
         scraper.doScrape();
       } catch (Exception e) {
@@ -418,6 +431,13 @@ public class JmxCollector extends Collector {
           "jmx_scrape_error", new ArrayList<String>(), new ArrayList<String>(), error));
       mfsList.add(new MetricFamilySamples("jmx_scrape_error", Type.GAUGE, "Non-zero if this scrape failed.", samples));
       return mfsList;
+    }
+
+    public List<MetricFamilySamples> describe() {
+      List<MetricFamilySamples> sampleFamilies = new ArrayList<MetricFamilySamples>();
+      sampleFamilies.add(new MetricFamilySamples("jmx_scrape_duration_seconds", Type.GAUGE, "Time this JMX scrape took, in seconds.", new ArrayList<MetricFamilySamples.Sample>()));
+      sampleFamilies.add(new MetricFamilySamples("jmx_scrape_error", Type.GAUGE, "Non-zero if this scrape failed.", new ArrayList<MetricFamilySamples.Sample>()));
+      return sampleFamilies;
     }
 
     /**
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 # compilation objects
 target/
 /.project
+dependency-reduced-pom.xml
--- a/collector/src/test/java/io/prometheus/jmx/JmxCollectorTest.java
+++ b/collector/src/test/java/io/prometheus/jmx/JmxCollectorTest.java
@@ -233,4 +233,17 @@ public class JmxCollectorTest {
       assertEquals(0.001, registry.getSampleValue("foo", new String[]{}, new String[]{}), .001);
     }
 
+    @Test(expected=IllegalStateException.class)
+    public void testDelayedStartNotReady() throws Exception {
+      JmxCollector jc = new JmxCollector("---\nstartDelaySeconds: 1").register(registry);
+      assertNull(registry.getSampleValue("boolean_Test_True", new String[]{}, new String[]{}));
+      fail();
+    }
+
+    @Test
+    public void testDelayedStartReady() throws Exception {
+      JmxCollector jc = new JmxCollector("---\nstartDelaySeconds: 1").register(registry);
+      Thread.sleep(2000);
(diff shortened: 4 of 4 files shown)


jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -57,6 +59,27 @@ public static void premain(String agentArgument, Instrumentation instrumentation
      server.setHandler(context);
      context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
 
-     server.start();
+     final int delayInSecs = Integer.parseInt(System.getProperty("jmx_exporter.agent.startdelay", "0"));
+
+     if (delayInSecs > 0) {
```
brian-brazil: thread.Sleep would seem simpler

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -57,6 +59,27 @@ public static void premain(String agentArgument, Instrumentation instrumentation
      server.setHandler(context);
      context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
 
-     server.start();
+     final int delayInSecs = Integer.parseInt(System.getProperty("jmx_exporter.agent.startdelay", "0"));
```
brian-brazil: This should really come from the config, and probably be down in the collector.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
bdellegrazie: Premain can't block, otherwise main doesn't start (hence the background thread)

bdellegrazie:
@brian-brazil Updated to address some of your concerns, could you please re-review?

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -1,32 +1,35 @@
...
 import java.util.regex.Pattern;
+
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 
 import org.yaml.snakeyaml.Yaml;
 
-import static java.lang.String.format;
+import io.prometheus.client.Collector;
+import io.prometheus.client.Collector.Describable;
```
brian-brazil: No need to import this, you can refer to it as Collector.Describable

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -106,6 +110,14 @@ private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObject
           yamlConfig = new HashMap<String, Object>();
         }
 
+        if (yamlConfig.containsKey("startDelay")) {
+          try{
+            cfg.startDelay = (Integer) yamlConfig.get("startDelay");
+          }
+          catch (NumberFormatException e) {
```
brian-brazil: catch on the same line as the }

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -106,6 +110,14 @@ private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObject
...
+            cfg.startDelay = 0;
```
brian-brazil: We should throw an exception here

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -106,6 +110,14 @@ private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObject
           yamlConfig = new HashMap<String, Object>();
         }
 
+        if (yamlConfig.containsKey("startDelay")) {
```
brian-brazil: startDelaySeconds, to be explicit about units

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -14,6 +14,7 @@
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
```
brian-brazil: why the blank line?

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -57,6 +59,27 @@ public static void premain(String agentArgument, Instrumentation instrumentation
      server.setHandler(context);
      context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
 
-     server.start();
+     final int delayInSecs = collector.getStartDelay();
```
brian-brazil: This feels like a layering violation, the delay would be better in the JmxCollecter I think

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
bdellegrazie: Done

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
bdellegrazie: Done

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
bdellegrazie: Done

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
bdellegrazie: Done

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
bdellegrazie: removed (left over from experimentation, injected by Eclipse to separate java. and other imports

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
bdellegrazie: I don't disagree, it definitely is a layering violation what alternative would you suggest?

bdellegrazie:
@brian-brazil updated as per your review. Please re-review at your earliest convenience

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
brian-brazil: Do it inside the collector

bdellegrazie: Do you mean inside collector.collect()?
That isn't called at startup any more (because Collector.Describable is implemented)

brian-brazil: Yes.

I don't see how that makes a difference.

bdellegrazie: The trigger for Collector.collect() is an external request right? So you'll be inserting a fixed delay into retrieving the first set of metrics, even though the container may actually be ready and available.

brian-brazil: You'd record the time in the constructor, and compare to that in the collect.

bdellegrazie:
@brian-brazil: updated as per your review, including tests.
Note that the behaviour if the JVM is not ready is to return an empty result (abar from the scrape time and error metric).
If you'd prefer we could throw an exception of some kind or record a scrape error, your call.

This version has a simple test for the delay parameter too.

README.md
```diff
@@ -29,6 +29,7 @@ See `./run_sample_httpserver.sh` for a sample script that runs the httpserver ag
 The configuration is in YAML. An example with all possible options:
 ```yaml
 ---
+startDelaySecs: 30
```
brian-brazil: I'd have this example as 0, or users will get confused when they try it

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -106,6 +108,13 @@ private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObject
           yamlConfig = new HashMap<String, Object>();
         }
 
+        if (yamlConfig.containsKey("startDelaySeconds")) {
+          try{
```
brian-brazil: space before {

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -396,17 +405,20 @@ public void recordBean(
...
       long start = System.nanoTime();
       double error = 0;
-      try {
-        scraper.doScrape();
-      } catch (Exception e) {
-        error = 1;
-        StringWriter sw = new StringWriter();
-        e.printStackTrace(new PrintWriter(sw));
-        LOGGER.severe("JMX scrape failed: " + sw.toString());
+      if ((config.startDelaySeconds == 0) ||
```
brian-brazil: I think showing an exception would be better, so the target appears as down to Prometheus.

As-is this isn't even setting jmx_scrape_error

README.md
bdellegrazie: Done

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
bdellegrazie: Done

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
bdellegrazie: Accepted, I did note that in my comment. It now throws an IllegalStateException and registers a scrape error

bdellegrazie:
@brian-brazil updated as per your previous request, could you please re-review? Thanks!

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -398,6 +407,10 @@ public void recordBean(
       JmxScraper scraper = new JmxScraper(config.jmxUrl, config.username, config.password, config.ssl, config.whitelistObjectNames, config.blacklistObjectNames, receiver);
       long start = System.nanoTime();
       double error = 0;
+      if ((config.startDelaySeconds > 0) &&
+        ((start - createTimeNanoSecs) / 1000000000L < config.startDelaySeconds)) {
+        throw new IllegalStateException("JVM not ready");
```
brian-brazil: This message might confuse the user about the JVM generally. I'd go for something like "JMXCollector waiting for startDelaySeconds "

bdellegrazie:
@brian-brazil: updated message as requested.

brian-brazil:
Thanks!

Starefossen:
Great work 👍🏼🎉

PR #144, opened 2017-05-19
Add possibility to add bean properties as labels

coder-hugo:
Hi @brian-brazil,

this PR adds the functionality to configure the following new properties for rules: `beanPropertiesAsLabels`, `whitelistBeanProperties` and `blacklistBeanProperties` which allow you to create more generic rules that automatically get useful labels based on the bean properties of your MBeans.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ See `./run_sample_httpserver.sh` for a sample script that runs the httpserver ag
 
 ## Configuration
 The configuration is in YAML. An example with all possible options:
-```
+```yaml
 ---
 hostPort: 127.0.0.1:1234
 jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:1234/jmxrmi
@@ -45,6 +45,9 @@ rules:
     help: "Cassandra metric $1 $2"
     type: GAUGE
     attrNameSnakeCase: false
+    beanPropertiesAsLables: true
+    whitelistBeanProperties: ["test"]
+    blacklistBeanProperties: ["type", "name"]
 ```
 Name     | Description
 ---------|------------
@@ -64,6 +67,9 @@ valueFactor | Optional number that `value` (or the scraped mBean value if `value
 labels   | A map of label name to label value pairs. Capture groups from `pattern` can be used in each. `name` must be set to use this. Empty names and values are ignored. If not specified and the default format is not being used, no labels are set.
 help     | Help text for the metric. Capture groups from `pattern` can be used. `name` must be set to use this. Defaults to the mBean attribute decription and the full name of the attribute.
 type     | The type of the metric, can be `GAUGE` or `COUNTER`. `name` must be set to use this. Defaults to `GAUGE`.
+beanPropertiesAsLabel | Add bean properties as label to the metric. `name` must be set to use this.
+whitelistBeanProperties | Allow only the given bean properties as labels. `beanPropertiesAsLabel` must be `true` to use this.
+blacklistBeanProperties | Don't add the given bean properties as labels. `beanPropertiesAsLabel` must be `true` to use this.
 
 Metric names and label names are sanitized. All characters other than `[a-zA-Z0-9:_]` are replaced with underscores,
 and adjacent underscores are collapsed. There's no limitations on label values or the help text.
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -8,6 +8,7 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.StringWriter;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
@@ -47,6 +48,9 @@ public class JmxCollector extends Collector {
       Type type = Type.GAUGE;
       ArrayList<String> labelNames;
       ArrayList<String> labelValues;
+      boolean beanPropertiesAsLabels;
+      List<String> whitelistBeanProperties = new ArrayList<String>();
+      List<String> blacklistBeanProperties = new ArrayList<String>();
     }
 
     private static class Config {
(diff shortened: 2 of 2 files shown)


brian-brazil:
You can already do this with the existing config file.

coder-hugo:
I don't see any option in the current version to do this. The bean properties are just added as labels for rules with no name. If you have configured a custom name for your rule you have to configure also the labels property for the rule to have labels for the resulting metrics. But this labels are static, there is currently no way to add all (or filtered by white- and blacklisted) bean properties as labels to your metrics.
With this PR we have the possibility to configure the following rules to have a very generic way to monitor metrics from all our java applications:
```yaml
  - pattern: '^com.someorg.*.metrics<type=.*, name=([^,]*).*><.*>(.*count):'
    name: $1_$2
    type: COUNTER
    attrNameSnakeCase: true
    beanPropertiesAsLabels: true
    blacklistBeanProperties: ['type', 'name']
  - pattern: '^com.someorg.*.metrics<type=.*, name=([^,]*).*><.*>(.*(?!count)):'
    name: $1_$2
    type: GAUGE
    attrNameSnakeCase: true
    beanPropertiesAsLabels: true
    blacklistBeanProperties: ['type', 'name']
```

brian-brazil:
> But these labels are static, there is currently no way to add all (or filtered by white- and blacklisted) bean properties as labels to your metrics.

That's correct, you'll need a rule for each set of labelnames.

coder-hugo:
But for this I have to know all possible bean properties and configure separate rules for this. In our case we are using the javaagent of the jmx_exporter in a tomcat. So the jmx_exporter yaml configuration is part of the tomcat configuration. If I add a new MBean to one of our web applications that has a new bean property that should be used as label I have to update the tomcat configuration. This is IMHO no good solution. With this PR I don't have to care about this I must just follow the naming convention (pattern of the rules of my last comment) for the MBean and the rest will happen automatically.

brian-brazil:
Not necessarily, labelnames can be dynamic.

It is presumed that you will need to customise the config for each metric source.

coder-hugo:
> It is presumed that you will need to customise the config for each metric source.

With this PR this is not necessary. And I'm not a fan of touching different systems for reaching a simple goal. In this case: Add a new metric to the monitoring.
With the current version I have to implement a new MBean that provides the metric. I have to test if the code is working. After this I have to write a new rule to the yaml file and have to test this as well cause if I do something wrong in the yaml file the tomcat won't start. The second task will become unnecessary if this PR is merged back.

brian-brazil:
If you don't like the default behaviour without rules, then you can specify exactly what you want to do instead. That includes choosing exactly which properties you want as labels, as it's not uncommon for some of them not to be suitable.

> With the current version I have to implement a new MBean that provides the metric.

It's recommended to use the Java client if you want to add instrumentation, it's much simpler and easier to work with. The JMX exporter is for when that's not an option.

coder-hugo:
By using the java client I have and hard dependency to prometheus within my application. By implementing MBeans for the metrics I use standard Java mechanisms for monitoring. So I can switch to a different monitoring system without touching the application. So IMHO using the jmx_exporter is the better approach for java monitoring with prometheus. And the functionality provided by this PR is the only missing thing that would make the usage of it perfect for us.

brian-brazil:
> By using the java client I have and hard dependency to prometheus within my application.

That's not the case. Our clients are specifically designed to allow working with other monitoring systems, and there's an example for graphite included.

I'm not willing to accept this change, as it presumes a standardisation which does not exist in JMX and it overlaps with what existing functionality allows for.

coder-hugo:
If I understand [this](https://github.com/prometheus/client_java#assets) correctly I have to add the `simpleclient` as a dependency to my application for the monitoring. So I have a dependency to prometheus within my application. JMX is already part of Java.

Nevertheless I played around with the currently available configuration options and found a way that lead almost to the same result.
My generic rules look now as follows:
```yaml
  - pattern: '^java.metrics.*<.*><.*>(.*count):'
    type: COUNTER
    attrNameSnakeCase: true
  - pattern: '^java.metrics.*<.*><.*>(.*(?!count)):'
    type: GAUGE
    attrNameSnakeCase: true
```
Which for example generates the following metrics:
```
# HELP java_metrics_logging_events_count Logging event count (java.metrics.logging.events<name=, context=/test/123, level=TRACE><>Count)
# TYPE java_metrics_logging_events_count gauge
java_metrics_logging_events_count{context="/test/123",level="TRACE",} 1169.0
java_metrics_logging_events_count{context="/test/123",level="INFO",} 44.0
java_metrics_logging_events_count{context="/test/123",level="DEBUG",} 1527.0
java_metrics_logging_events_count{context="/test/123",level="WARN",} 60.0
```
While doing this I found a bug in the current version. As you can see the "MBean name" matches the pattern of the first rule so the type should be counter not gauge. I'll create a separate PR for fixing this.

Furthermore I think there is still one change included in this PR that could be "merged" back. You can specify the language of a code block by writing the name after the tripple backticks. This would improve the syntax highlighting of the README.md. :wink: (See line 30 of the README.md within my commit)

brian-brazil:
> If I understand this correctly I have to add the simpleclient as a dependency to my application for the monitoring. So I have a dependency to prometheus within my application

You have a dependency on that bit of code. Not on the Prometheus monitoring system.

Don't depend on negative lookahead, that'll break some day when we get a performant version of RE2. Processing halts at the first match.

Something wrong with the types there alright. Looks like it isn't being copied. I had thought that was already fixed. PR welcome.

coder-hugo:
Having the `simpleclient` in the application is not just a dependency on a bit of code it also brings some other issues. E.g. I have to start the server that provides the metrics endpoint within my application. This is something I don't want to have within my application. As I told earlier we are writing web-applications that are deployed into a tomcat. In this case I don't have to care about network stuff (on which port, IP should the sever listen for incoming traffic) as this is done by the tomcat and can be configured by somebody else. And this should be the same for the monitoring endpoint as well. For this reason we try to use standardized solutions (like JMX) as much as possible within our applications so they can be deployed to every servlet container without the need to configure something special.

brian-brazil:
> I have to start the server that provides the metrics endpoint within my application.

You don't have to. You could be talking to a monitoring system other than Prometheus, or hook it into a HTTP server you already have.

coder-hugo:
Ok, you are right. In the README of the `simpleclient` I just saw the code block that starts a server in the Export section. After a deeper look I saw that there is also `MetricsServlet`. Nevertheless using MBeans for the monitoring is our preferred solution at the moment.
And btw. thanks for your fast answering :smiley: There are other projects on GitHub where you have to wait weeks to get an answer.

brian-brazil:
> I just saw the code block that starts a server in the Export section

Yeah, it's the same as in Python. That's there to get people up and running quickly. In reality you'll usually hook it into what you already have.

PR #145, opened 2017-05-19
Use the type provided by the rule for the defaultExport

coder-hugo:
Hi @brian-brazil,

the current version of the jmx_exporter uses the type gauge for all metrics that match a rule that has no name configured and not the type configured for the rule. This PR fixes this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ See `./run_sample_httpserver.sh` for a sample script that runs the httpserver ag
 
 ## Configuration
 The configuration is in YAML. An example with all possible options:
-```
+```yaml
 ---
 hostPort: 127.0.0.1:1234
 jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:1234/jmxrmi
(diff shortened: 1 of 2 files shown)


brian-brazil:
Could you change the default to UNTYPED while you're at it?

coder-hugo:
I just had a look to the README before changing the default value and saw that the description for `type` currently says the following: 
> The type of the metric, can be `GAUGE` or `COUNTER`. `name` must be set to use this. Defaults to `GAUGE`.

I'll change the default here as well but in this case `UNTYPED` should be at least listed as possible option. Shouldn't it? Additionally I saw in the code that there are 2 further types: `SUMMARY`, `HISTOGRAM` which I hadn't heard about yet. The current code doesn't prevent to use them as type so the question is can they be used for metrics provided by the jmx_exporter and should be listed in the README or not?

brian-brazil:
> I'll change the default here as well but in this case UNTYPED should be at least listed as possible option. Shouldn't it?

That makes sense.

> Additionally I saw in the code that there are 2 further types: SUMMARY, HISTOGRAM which I hadn't heard about yet. 

It's not possible for the JMX exporter to produce these, as it works on a time series by time series basis.

brian-brazil:
Thanks!

coder-hugo:
Just one last question :wink: Will there be a new release of the JMX exporter that contains this fix.

brian-brazil:
This change alone isn't enough for a release, as it has no impact. The types aren't used for anything currently.

PR #146, opened 2017-05-19
javaagent: Handle containers that use custom MBeanServers

### LINKED ISSUE #128: JMX agent metrics from Wildfly/Undertow broken since v0.8

bdellegrazie:
@brian-brazil 
Some containers (e.g. Wildfly 8, 9, 10) use a custom MBeanServer.
The simplest way to deal with this is to ask the user to set the factory
to either the correct class (if known) or a non-existant class and let the
container intialisation process correct it.

Any call to Management.getPlatformMBeanServer() before the container is
ready will then fail with a ClassNotFoundException until container
initialisation is complete.

In terms of scraping we swallow the resulting ClassNotFoundException to avoid
polluting the container's logs. If there happens to be a scrape the caller will
get an empty page with no metrics

Fixes #128

### DIFF
--- a/docs/Wildfly.md
+++ b/docs/Wildfly.md
@@ -1,0 +1,13 @@
+# Wildfly
+
+Wildfly 8,9 and 10 use a custom MBeanServer implementation. To use JMX Exporter:
+
+* As a _javaagent_ add:
+`-Djavax.management.builder.initial=io.prometheus.jmx.ContainerNotReadyYet` to the javaagent line.
+The class itself (ContainerNotReadyYet) does not exist and Wildfly overwrites the system property with a valid value during its startup. This *may* work with
+other containers that use custom MBeanServers.
+
+* As a remote scraper:
+Add the `jboss-cli-client.jar` from the Wildfly installation to the classpath of JMX Exporter and use the Wildfly version specific JMX URL in the configuration file.
+The JMX URL will be the same one as used to connect JConsole. The extra JAR is required because Wildfly uses a custom protocol to connect via the the HTTP(s) admin port.
+
--- a/example_configs/wildfly-10.yaml
+++ b/example_configs/wildfly-10.yaml
@@ -1,4 +1,5 @@
 ---
+# See docs/Wildfly.md for additional info
 lowercaseOutputName: true
 lowercaseOutputLabelNames: true
 rules:


README.md
```diff
@@ -21,6 +21,19 @@ To bind the java agent to a specific IP change the port number to `host:port`.
 
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
 
+### Wildfly
```
brian-brazil: I don't think we should be mentioning particular pieces of software that are doing non-standard things, particularly not above the primary reference docs.

These docs belong on Wildfly's end.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -89,7 +90,16 @@ public void doScrape() throws Exception {
         MBeanServerConnection beanConn;
         JMXConnector jmxc = null;
         if (jmxUrl.isEmpty()) {
-          beanConn = ManagementFactory.getPlatformMBeanServer();
+          try {
```
brian-brazil: Throwing an exception is the correct behaviour, as that'll make the scrape fail.

README.md
bdellegrazie: Would you prefer me to move the docs to the wildfly example configuration or a separate .md file in the repo?

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
bdellegrazie: I thought you might say that... no problem, will revert. At best this whole thing is a work-around anyway.

README.md
Starefossen: Could we have a `/docs` directory for this kind on things?

bdellegrazie: @Starefossen sure good suggestion

brian-brazil: I mean move it to the documentation of the Wildfly project. They've chosen to do things have that break standard java mechanisms, so it's really up to them to deal with the results of that.

bdellegrazie: done

bdellegrazie: I can and will try but they aren't exactly responsive. I'll see what they say on the mailing list and try to add a comment to the JMX page of the documentation. With respect, I'd appreciate you allowing this to be merged given its only a small amount of documentation.

brian-brazil: I don't think it's sane for us to document workarounds for every single application that's decided to go off the beaten path, there's enough inconsistency in JMX already for us to deal with.

bdellegrazie:
@brian-brazil: I've moved the doc from README.md and put it in docs/Wildlfy.md as per the updated PR. Is that acceptable?

brian-brazil:
What I'd suggest is putting the docs in the top of the .yaml so it's all together.

cfrantsen:
Is this PR relevant anymore? I have no problems running HEAD on WildFly 10.1.0.

brian-brazil:
Hmm, that's unexpected but don't look a gift horse in the mouth.

bdellegrazie:
You will ... it depends upon how many things you have deployed and how fast or slow Wildfly is to start up and setup its MBean server.

cfrantsen:
I'd assume tuning the startDelaySeconds parameter to more than the startup time would fix any problems no?

PR #147, opened 2017-06-02
Add example config for flink 1.2

prymitive:
Metrics exposed in flink 1.2 are different than those from earlier versions (docs: https://ci.apache.org/projects/flink/flink-docs-release-1.2/monitoring/metrics.html#system-metrics), add a new config covering both task and jobmanager

@brian-brazil

### DIFF
--- a/example_configs/flink-1-2.yml
+++ b/example_configs/flink-1-2.yml
@@ -1,0 +1,121 @@
+rules:
+  ## Job Manager ##
+  - pattern: org\.apache\.flink\.jobmanager\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)<host=(.*)><>(Count|Value|Rate)
+    name: flink_jobmanager_$1_$2_$3_$4_$5
+    type: GAUGE
+    labels:
+      host: $6
+
+  - pattern: org\.apache\.flink\.jobmanager\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)<host=(.*)><>(Count|Value|Rate)
+    name: flink_jobmanager_$1_$2_$3_$4
+    type: GAUGE
+    labels:
+      host: $5
+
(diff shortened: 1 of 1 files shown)


brian-brazil:
This can be collapsed down further, see the existing flink conifg as an example.

The naming of the files should also be changed to follow what we did for kafka.

prymitive:
Updated with most recent version I'm using, hope it's better

example_configs/flink-1-2.yml
```diff
@@ -0,0 +1,121 @@
+rules:
+  ## Job Manager ##
+  - pattern: org\.apache\.flink\.jobmanager\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)\.([a-zA-Z\-\_]+)<host=(.*)><>(Count|Value|Rate)
```
brian-brazil: You only want the Count and Value.

Rate can be better calculated on the Prometheus end from the Count.

example_configs/flink-1-2.yml
```diff
@@ -0,0 +1,121 @@
...
+    name: flink_jobmanager_$1_$2_$3_$4_$5
```
brian-brazil: The jmx exporter will put in underscores for invalid characters, you can go just one capture group for $1-$5

example_configs/flink-1-2.yml
```diff
@@ -0,0 +1,121 @@
...
+    type: GAUGE
+    labels:
+      host: $6
```
brian-brazil: What is in this host label?

example_configs/flink-1-2.yml
```diff
@@ -0,0 +1,121 @@
...
+  - pattern: org\.apache\.flink\.taskmanager\.job\.task\.operator\.(KafkaConsumer|KafkaProducer)\.([a-zA-Z\-\_]+).([a-zA-Z\-\_]+)-([0-9]+)<task_id=(.*), subtask_index=(.*), job_id=(.*), tm_id=(.*), task_attempt_id=(.*), job_name=(.*), operator_name=(.*), task_attempt_num=(.*), host=(.*), task_name=(.*)><>(Count|Value|Rate)
+    name: flink_taskmanager_job_task_operator_$1_$2
+    type: GAUGE
+    labels:
+      topic: $3
+      partition: $4
+      subtask_index: $6
+      job_id: $7
+      tm_id: $8
+      job_name: $10
```
brian-brazil: Do the job_id and job_name provide the same information?

PR #161, opened 2017-07-06
Prevent unnecessary surcharging of pattern string anchors

obourdon:
By default starring anchor ^.* and ending anchor .*$ are [always added](https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxCollector.java#L170) to matching pattern
If user specified anchors in his own pattern this is overkill and might be misleading/generate errors

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -99,6 +99,17 @@ public class JmxCollector extends Collector {
       }
     }
 
+    private String defaultAnchoring(String pat) {
+      // Do not surcharge anchoring patterns
+      if (!pat.startsWith("^")) {
+        pat = "^.*" + pat;
+      }
+      if (!pat.endsWith("\\$")) {
+        pat = pat + ".*$";
+      }
+      return pat;
+    }
+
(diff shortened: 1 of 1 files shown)


brian-brazil:
Hmm, this pattern is incorrect as-is as it doesn't consider alternation (e.g. `a|b` becomes `(^.*a)|(b.*$)`.). This should be changed to `^(?:.*pattern.*)$`

obourdon:
@brian-brazil I am not sure I follow you there ? Current code at [https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxCollector.java#L170](https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxCollector.java#L170) is already wrong, is this what you meant ?
I had a case on my side where my pattern was:
^xxxxxxx$
and debugging the exporter showed that my pattern had been changed to
^.*^xxxxx$.*$
by the code mentioned above
Please advise

brian-brazil:
Yes, the current code is wrong.

The additions would have had no impact on your pattern.

brian-brazil:
The bug in the existing code is fixed.

Issue #155, opened 2017-06-12
Switch to plain http server


brian-brazil:
Using jetty has caused a number of issues in this exporter. As JMX is more likely to be used with odd/old/unusual software, I think we should implement a httpserver exposition in client_java and then use that instead of jetty.

leogomes:
Do you want any help with this?

Cheers,
Leo

brian-brazil:
That'd be great, details are over in #156.

brian-brazil:
Are you currently working on this @leogomes?

leogomes:
@brian-brazil Haven't done much, but just tested using https://github.com/prometheus/client_java/pull/73 HTTPServer and it worked fine for my basic JMX settings. If I understood correctly you'd expect me to also clean-up that PR, so that it can be merged, right? Should we discuss there on what exactly needs to be done?

You can check how I used it here: https://github.com/leogomes/jmx_exporter/commit/16f3735714b1df19d6abc4d0f670eeaa1dcf4b7e

brian-brazil:
I've just started working on cleaning up that PR. We won't get SO_REUSEADDR, but otherwise it seems workable.

leogomes:
Other thing I noticed, comparing to the current implementation in JMX exporter, is that it doesn't allow you to specify the host only the port.

`mServer.bind(new java.net.InetSocketAddress(port), 3);`

It may be a problem for machines with multiple NICs.

brian-brazil:
Both the agent and separate process currently support setting the host.

leogomes:
Yes, they do, but not the HTTP server that is in the PR you're reviewing ;)

brian-brazil:
https://github.com/prometheus/client_java/tree/httpserver is what I have so far

leogomes:
OK, I can add the host thing once it's merged and modify my PR then. Let me know if you want me to do that.

PR #162, opened 2017-07-10
Use simple HTTP Server

leogomes:
Motivation:
https://github.com/prometheus/jmx_exporter/issues/155

Modifications:
Changed the code of both the agent and external HTTP process to use the simple HTTP server provided on https://github.com/prometheus/client_java/pull/73

Result:
JMX exporter works with the simple HTTP Server.

### DIFF
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -7,7 +7,6 @@
     <version>0.10-SNAPSHOT</version>
   </parent>
 
-  <groupId>io.prometheus.jmx</groupId>
   <artifactId>jmx_prometheus_httpserver</artifactId>
   <description>
     See https://github.com/prometheus/jmx_exporter/blob/master/README.md
@@ -26,9 +25,9 @@
       <version>0.0.21</version>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.jetty</groupId>
-      <artifactId>jetty-servlet</artifactId>
-      <version>8.1.7.v20120910</version>
+      <groupId>io.prometheus</groupId>
+      <artifactId>simpleclient_httpserver</artifactId>
+      <version>0.0.9-SNAPSHOT</version>
     </dependency>
   </dependencies>
 
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -21,11 +21,6 @@
     </dependency>
     <dependency>
       <groupId>io.prometheus</groupId>
-      <artifactId>simpleclient_servlet</artifactId>
-      <version>0.0.25</version>
-    </dependency>
-    <dependency>
-      <groupId>io.prometheus</groupId>
       <artifactId>simpleclient_httpserver</artifactId>
       <version>0.0.25</version>
     </dependency>
--- a/jmx_prometheus_httpserver/pom.xml
+++ b/jmx_prometheus_httpserver/pom.xml
@@ -22,12 +22,12 @@
     <dependency>
       <groupId>io.prometheus</groupId>
       <artifactId>simpleclient_servlet</artifactId>
-      <version>0.0.26-SNAPSHOT</version>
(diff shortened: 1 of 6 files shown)


jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
```diff
@@ -18,24 +14,14 @@ public static void main(String[] args) throws Exception {
 
      String[] hostnamePort = args[0].split(":");
      int port;
-     InetSocketAddress socket;
 
      if (hostnamePort.length == 2) {
        port = Integer.parseInt(hostnamePort[1]);
-       socket = new InetSocketAddress(hostnamePort[0], port);
```
brian-brazil: We'll want to keep the existing host support around.

leogomes: Sure, I will put it back once HTTP server is able to take a host and port. I can do the modification and test, if you want.

jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
```diff
@@ -14,14 +16,24 @@ public static void main(String[] args) throws Exception {
...
        port = Integer.parseInt(hostnamePort[0]);
+       socket = new InetSocketAddress(port);
      }
 
-     JmxCollector jc = new JmxCollector(new File(args[1])).register();
-     HTTPServer server = new HTTPServer(port);
+     new JmxCollector(new File(args[1])).register();
+     final HTTPServer server = new HTTPServer(socket, CollectorRegistry.defaultRegistry);
+     
+     Runtime.getRuntime().addShutdownHook(new Thread() {
```
brian-brazil: We don't need this, the httpserver should keep going until the process dies.

There could be a long time between the TERM and the process actually dying.

jmx_prometheus_httpserver/pom.xml
```diff
@@ -22,12 +22,12 @@
     <dependency>
       <groupId>io.prometheus</groupId>
       <artifactId>simpleclient_servlet</artifactId>
-      <version>0.0.21</version>
+      <version>0.0.26-SNAPSHOT</version>
```
brian-brazil: 0.0.25. We don't use snapshot releases in production.

jmx_prometheus_httpserver/src/main/java/io/prometheus/jmx/WebServer.java
```diff
@@ -28,14 +26,7 @@ public static void main(String[] args) throws Exception {
...
-
-     Server server = new Server(socket);
-     ServletContextHandler context = new ServletContextHandler();
-     context.setContextPath("/");
-     server.setHandler(context);
-     context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
-     server.start();
-     server.join();
+     new JmxCollector(new File(args[1])).register();
+     new HTTPServer(socket, CollectorRegistry.defaultRegistry);
```
brian-brazil: There're nothing here to wait, this will shutdown immediately.

leogomes: Not really, HTTPServer creates threads that will keep running, even if the main thread exits, c.f. [termination of main thread exit](https://stackoverflow.com/questions/5642802/termination-of-program-on-main-thread-exit).

You can run the [PR locally as well](https://help.github.com/articles/checking-out-pull-requests-locally/).

jmx_prometheus_httpserver/pom.xml
```diff
@@ -23,12 +22,12 @@
     <dependency>
       <groupId>io.prometheus</groupId>
       <artifactId>simpleclient_servlet</artifactId>
```
brian-brazil: You don't need this any more

brian-brazil:
Thanks!

Issue #159, opened 2017-06-30
30+ mins HTTP response time using remote JMX host


agolomoodysaada:
Releases Tested: `0.9` and `0.7`

I'm trying to setup jmx_exporter with a remote host exposing jmx metrics. I can already access JMX metrics using `jconsole` without any issues. However, using the below configuration doesn't return any http responses when hitting `localhost:5556/metrics` and keeps spinning forever. I do see some verbose logs showing up when I hit the endpoint.

What am I doing wrong?

Docker compose
```yml
  jmx_exporter:
    image: java:8-alpine
    command: java -verbose -jar /opt/jmx_exporter/jmx_exporter.jar 5556 /opt/jmx_exporter/config.yml
    ports:
      - '5556:5556'
    volumes:
      - './jmx_prometheus_httpserver-0.9-jar-with-dependencies.jar:/opt/jmx_exporter/jmx_exporter.jar'
      - './kafka-jmx-config.yml:/opt/jmx_exporter/config.yml'
```

Kafka config file based on sample config
```yml
hostPort: <MY_REMOTE_HOST>:<MY_REMOTE_PORT>
ssl: false
lowercaseOutputName: true
startDelaySecs: 0
rules:
  - pattern : kafka.cluster<type=(.+), name=(.+), topic=(.+), partition=(.+)><>Value
    name: kafka_cluster_$1_$2
    labels:
      topic: "$3"
      partition: "$4"
  - pattern : kafka.log<type=Log, name=(.+), topic=(.+), partition=(.+)><>Value
    name: kafka_log_$1
    labels:
      topic: "$2"
      partition: "$3"
  - pattern : kafka.controller<type=(.+), name=(.+)><>(Count|Value)
    name: kafka_controller_$1_$2
  - pattern : kafka.network<type=(.+), name=(.+)><>Value
    name: kafka_network_$1_$2
  - pattern : kafka.network<type=(.+), name=(.+)PerSec, request=(.+)><>Count
    name: kafka_network_$1_$2_total
    labels:
      request: "$3"
  - pattern : kafka.network<type=(.+), name=(\w+), networkProcessor=(.+)><>Count
    name: kafka_network_$1_$2
    labels:
      request: "$3"
    type: COUNTER
  - pattern : kafka.network<type=(.+), name=(\w+), request=(\w+)><>Count
    name: kafka_network_$1_$2
    labels:
      request: "$3"
  - pattern : kafka.network<type=(.+), name=(\w+)><>Count
[...]

PR #163, opened 2017-07-12
Docker Image for HTTP exporter

amimimor:
I've created a dir named 'docker' that contains the necessary files to build and run a Docker container of the http exporter. If the PR is of interest, I would happily add docs on the correct 'docker build ...' cmd as well as the 'docker run ...' cmd as I've already outlined in [my docker hub repo](https://hub.docker.com/r/amimimor/jmx-prometheus/). Currently the image is hosted on my docker hub account. If you'd like I can change the builder to create the image under a different account (prometheus?)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,29 @@
+FROM openjdk:8u131-jdk-alpine
+
+MAINTAINER "Amit Mor" <amimimor@gmail.com>
+
+RUN apk update && apk add bash maven
+ARG RUNDIR
+ENV RUNDIR ${RUNDIR:-/usr/local/jmx_exporter}
+WORKDIR $RUNDIR
+ADD jmx_prometheus_httpserver/ jmx_prometheus_httpserver/
+ADD jmx_prometheus_javaagent/ jmx_prometheus_javaagent/
+ADD collector/ collector/
+ADD pom.xml pom.xml
+# get the version from the pom.xml
+ENV mvn_evaluate "org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate  -Dexpression=project.version"
+# let maven import the plugin dependcies
+RUN mvn $mvn_evaluate > /dev/null
+# extract the project version and pass it into the container as env var, to be consumed by the start.sh script
+RUN VERSION=$(mvn $mvn_evaluate | grep -v -e "^\[INFO\]") && echo export JMX_PROMETHEUS_VERSION=$VERSION > /root/.bashrc && printf "\nFound version: %s" $VERSION
+# compile
+RUN mvn compile package
+ADD docker/start.sh /start.sh
+RUN chmod +x /start.sh
+# config volume
+RUN mkdir -p $RUNDIR/jmx_prometheus_httpserver/config/ 
+VOLUME $RUNDIR/jmx_prometheus_httpserver/config/
+# ports
+EXPOSE 39390
+# cmd
+CMD /start.sh
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -x
+
+mvn_evaluate="org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate  -Dexpression=project.version"
+mvn $mvn_evaluate > /dev/null
+VERSION=$(mvn $mvn_evaluate | grep -v -e "^\[INFO\]") 
+
+docker build -t amimimor/jmx-prometheus:$VERSION -f docker/Dockerfile .
+
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+# Script to run a java application for testing jmx4prometheus.
+env
+
+RUNDIR=${RUNDIR:-/usr/local/jmx_exporter}
+echo RUNDIR $RUNDIR
+BIND_HOST=${BIND_HOST:-localhost}
+echo BINDING TO HOST:PORT $BIND_HOST:39390
+CONFIG_FILE_DIR=${CONFIG_FILE_DIR:-/usr/local/jmx_exporter/jmx_prometheus_httpserver/config/config.yml}
+echo GOING TO USE $CONFIG_FILE_DIR AS INPUT CONFIG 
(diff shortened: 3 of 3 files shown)


brian-brazil:
Thanks for sharing. This exporter is intended to be primarily run as a Java agent, so a Docker image doesn't help much towards that as everyone would have to create their own.

amimimor:
Hi Brian and thanks for the reply.
The Docker image doesn't need to be created by anybody - I just added the way to do that if someone want to modify the sources and recreate the image.
The image can be consumed as it is.
The 'docker run ...' defines custom params such as to which port the process should bind, etc. but besides that, the user only has to create a certain config.yml (as with the javaagent or http server)

On 12 Jul 2017, at 22:37, Brian Brazil <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for sharing. This exporter is intended to be primarily run as a Java agent, so a Docker image doesn't help much towards that as everyone would have to create their own.

brian-brazil:
That doesn't help with running an agent, as the jar would need to be added into the application's docker image.

amimimor:
Agreed.
But that is exactly why I created the Image, because using the javaagent seemed to me as a too 'involved' mechanism exposing processes which are not my own.
For example, I needed the exporter to be exporting Kafka JMX metrics.
Our Kafka is dockerized. I didn't think it would be a too good idea to 'fork' the wurstmeister/kafka image, then to edit its start script so that it would be wrapped by the javaagent.
However, just 'side kicking' the exporter container and pointing it to the Kafka container was quite easy and not 'involved' at all.

Anyways. This is OSS. I totally get that my goals do not align with the projects' goals so no hard feelings if it gets rejected.
Cheers,
Amit M.
On 12 Jul 2017, at 22:43, Brian Brazil <notifications@github.com<mailto:notifications@github.com>> wrote:


That doesn't help with running an agent, as the jar would need to be added into the application's docker image.

brian-brazil:
We generally try to encourage the java agent (for example another user recently discovered that the http server form combined with a kafka with many many partitions doesn't end well), and using official builds so a Docker image like this isn't something I'm aiming for right now.

amimimor:
Cool. Is there any license violation or any principle/personal objection with me keeping the Docker image publicly available in my docker hub repo?
It is, btw, tagged with the official mvn version (0.10.0-SNAPSHOT) if I recall.


On 12 Jul 2017, at 22:54, Brian Brazil <notifications@github.com<mailto:notifications@github.com>> wrote:


We generally try to encourage the java agent (for example another user recently discovered that the http server form combined with a kafka with many many partitions doesn't end well), and using official builds so a Docker image like this isn't something I'm aiming for right now.

brian-brazil:
This is all Apache 2.0, there's no problems there.

agolomoodysaada:
@brian-brazil since you seem to be referring to me here https://github.com/prometheus/jmx_exporter/issues/159

> for example another user recently discovered that the http server form combined with a kafka with many many partitions doesn't end well

I'd like to mention that my issue was resolved and mainly due to me using the exporter from laptop against a remote JMX endpoint. By running the same docker container on the remote machine side-by-side the Kafka process, everything worked significantly better.

r4j4h:
Can also confirm that running it in a sidecar worked for us. We used it for the same reasons as @amimimor in a Kubernetes environment for metrics scraping. The only interesting thing we encountered with this approach wasn't performance related but was the fact that the jmx exporter http server serves its own jmx metrics as well :)

Should point out that according to our memory usage metrics those sidecars seem to average ~90MB of RAM though, so it's likely more efficient to have them combined.

brian-brazil:
> The only interesting thing we encountered with this approach wasn't performance related but was the fact that the jmx exporter http server serves its own jmx metrics as well :)

No it doesn't, that'd cause confusion.

> Should point out that according to our memory usage metrics those sidecars seem to average ~90MB of RAM though, so it's likely more efficient to have them combined.

The Prometheus architecture is not to have such "uberagents" as they're harder to manage.

amimimor:
Let me just add that since we are microserviced, and using Docker, managing anything is an issue. Gladly it is not ours since we let cluster managers such as Kubernetes to do that, therefore a few uberagents around is not an issue at all, at least for us.

On 19 Jul 2017, at 9:31, Brian Brazil <notifications@github.com<mailto:notifications@github.com>> wrote:


The only interesting thing we encountered with this approach wasn't performance related but was the fact that the jmx exporter http server serves its own jmx metrics as well :)

No it doesn't, that'd cause confusion.

Should point out that according to our memory usage metrics those sidecars seem to average ~90MB of RAM though, so it's likely more efficient to have them combined.

The Prometheus architecture is not to have such "uberagents" as they're harder to manage.

brian-brazil:
You misunderstand the management issues with uberagents. Docker and Kubernetes do not make a difference, see https://www.robustperception.io/one-agent-to-rule-them-all/

eedugon:
@amimimor , @r4j4h , @agolomoodysaada : I agree with your view, and I don't know if we are really missing something or the real benefits of running the exporter as a javaagent. For us the javaagent implies changes in the main java service, and we prefer to scrape the mbeans from the outside (as a sidecar in a k8s pod for example).

Anyway, I have written a Dockerfile + start.sh to create a jmx-exporter http docker with the following features:

- Automatic Configuration: config.xml creation based on environment variables and templates.

- Static Configuration: if a config.xml is mapped into /opt/jmx_exporter/config/, then environment variables won't be considered, and the provided configuration will be used (static).

- Flexible rules: via environment variables we can select the rules to apply to the configuration (only if Automatic Configuration feature is used and only 1 rules file allowed).

- check_init : With check init enabled (by default is disabled), jmx_exporter will wait until the remote jmx port is available (this feature uses nagios check_jmx plugin) before starting the exporter itself.

Check it out if you want: https://github.com/eedugon/jmx-exporter-dockerized

@brian-brazil : I assume that you don't, but if you ever consider it useful don't hesitate on reusing the code or asking me to share it.

Regards,
   Eduardo

balboah:
I can also add that I'm successfully using it as a sidecar in kubernetes instead of building a custom image with an agent baked in, basically same as @eedugon. Wanted to see if submitting a PR would be accepted but I guess not after finding this issue :)

tmc:
Just another peep that having this available as an addon container in k8s would be useful.

PR #164, opened 2017-07-13
Add metrics domain renaming feature

obourdon:
This allows you to rename all metrics at once without having to deal with "complex" regexp patterns
For example if you scrape several processes on the same machine like ZooKeeper, Kafka and Cassandra you might want to express renaming for Cassandra as follow:
`rules:
  - pattern: '.*'
    replaceDomains: {'^java.lang(.*)$' : 'cassandra.jvm$1', '^org.apache.cassandra(.*)$' : 'cassandra$1' } 
`

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -37,6 +37,11 @@ public class JmxCollector extends Collector implements Collector.Describable {
 
     private static final Logger LOGGER = Logger.getLogger(JmxCollector.class.getName());
 
+    private static class DomainRename {
+      Pattern pattern;
+      String replacement;
+    }
+
     private static class Rule {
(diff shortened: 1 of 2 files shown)


brian-brazil:
This can be done with the existing configs, and I don't think it's a great user experience to have two different way do things.

In addition your particular example is an anti-pattern, as a) the JVM metrics are taken care of for you when using the java agent and b) they shouldn't be prefixed with the application name - that's what labels are for.

obourdon:
@brian-brazil I am not sure I understand how it could be done exactly  with current configs neither your point a). However I understand that this might not be the way to do things. Thanks for the feedback

brian-brazil:
The example configs for Cassandra and Kafka provide pretty good metrics, and I'd advise sticking with those so you can reuse others dashboards etc.

Unfortunately Zookeeper doesn't expose much in the way of metrics.

PR #165, opened 2017-07-14
Servlet path for Java agent can be configured by runtimes args.

otigges:
@brian-brazil,

JMX exporter works great in our project as Java agent, but we need to be able to change the servlet path (currently it is "/metrics"). Hope this PR is helpful. Otherwise please tell me what needs to be changed. Thanks in advance.

Oliver


brian-brazil:
Thanks for the PR. All this code is about to change, and generally this is not something you should need to tweak as it is the Prometheus default.

otigges:
Thanks for your reply. I'm aware that /metrics is the default. But we needed to change the context-root in our systems in order to set different permissions, so we use /management/metrics. Then we'll need to stay with our forked/patched version.

brian-brazil:
That seems like something for a reverse proxy which shouldn't care about the path. You can always use the JmxCollector class in your own code, it is intended for that to work (though rarely needed).

PR #174, opened 2017-08-19
Add Apache Solr example.

mnagaya:
We have this config running in production.

### DIFF
--- a/example_configs/solr.yml
+++ b/example_configs/solr.yml
@@ -1,0 +1,12 @@
+---
+whitelistObjectNames: [ "java.lang:*", "org.eclipse.jetty.util.thread:*", "org.eclipse.jetty.server.handler:*", "solr/*:type=/select,*", "solr/*:type=/update,*", "solr/*:type=searcher,*", "solr/*:type=documentCache,*", "solr/*:type=updateHandler,*"]
+
+rules:
+  - pattern: "^java(.(?!OperatingSystem))*$"
+  - pattern: "^solr/([a-zA-Z0-9_]+)_([a-zA-Z0-9]+)_([a-zA-Z0-9]+)<type=(.+), id=(.+)><>(.*):"
+    name: solr_$4_$6
+    labels:
+      collection: $1
+      shard: $2
+      replica: $3
+  - pattern: ".*"


example_configs/solr.yml
```diff
@@ -0,0 +1,12 @@
+---
+whitelistObjectNames: [ "java.lang:*", "org.eclipse.jetty.util.thread:*", "org.eclipse.jetty.server.handler:*", "solr/*:type=/select,*", "solr/*:type=/update,*", "solr/*:type=searcher,*", "solr/*:type=documentCache,*", "solr/*:type=updateHandler,*"]
+
+rules:
+  - pattern: "^java(.(?!OperatingSystem))*$"
```
brian-brazil: Exclude all these, they're already taken care of by the default exports

example_configs/solr.yml
```diff
@@ -0,0 +1,12 @@
+---
+whitelistObjectNames: [ "java.lang:*", "org.eclipse.jetty.util.thread:*", "org.eclipse.jetty.server.handler:*", "solr/*:type=/select,*", "solr/*:type=/update,*", "solr/*:type=searcher,*", "solr/*:type=documentCache,*", "solr/*:type=updateHandler,*"]
```
brian-brazil: Why are you adding this?

example_configs/solr.yml
```diff
@@ -0,0 +1,12 @@
...
+
+rules:
+  - pattern: "^java(.(?!OperatingSystem))*$"
+  - pattern: "^solr/([a-zA-Z0-9_]+)_([a-zA-Z0-9]+)_([a-zA-Z0-9]+)<type=(.+), id=(.+)><>(.*):"
+    name: solr_$4_$6
+    labels:
+      collection: $1
+      shard: $2
+      replica: $3
```
brian-brazil: Does this happen to be a hostname?

example_configs/solr.yml
```diff
@@ -0,0 +1,12 @@
...
+  - pattern: ".*"
```
brian-brazil: You shouldn't fall back to the defaults here, instead create rules for what you want.

mnagaya:
Sorry for late reply.  Solr7.6 now provides solr-exporter:
https://lucene.apache.org/solr/guide/7_6/monitoring-solr-with-prometheus-and-grafana.html

PR #183, opened 2017-09-01
Only retrieve mbeans and attributes that match rules.

janssk1:
This approach reduced the scape time on my (jboss) setup to about 25% (4 times less).

Change-Id: Ia3a933d2ba0e38143847f98bbd21ba4061056ea1

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -7,18 +7,12 @@ import java.io.PrintWriter;
 import java.io.File;
 import java.io.FileReader;
 import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import javax.management.MBeanInfo;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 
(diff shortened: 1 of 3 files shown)


brian-brazil:
What is this intended to do?

In general as the pattern is a regex and the value is needed before we apply the pattern, everything must be retrieved.

janssk1:
This optimization will do nothing if your patterns are very 'wide'
However, if the patterns are explicit about what attributes and mbeans must be collected, this makes sure only those are retrieved.  Some of my mbeans have a lot of arguments but if my rules only express interest in one of them no need to collect them all. 

This is especially important on jboss, which if notoriously slow in  getting JMX attribute values.

brian-brazil:
I'm not really sure what your code is doing, and I don't think it's correct.

You do seem to be using additional regex matches. We know with some systems that the regexes end up being the expensive part, so this change would make those worse.

janssk1:
Suppose a pattern like this:
jboss.as<subsystem=datasources, data-source=(?:.*), statistics=pool><>(InUse|Available|Active)Count

The pull request will make sure that: 
* JMX metadata is only retrieved for mbeans that match the mbean part of (one or more) patterns
* Only attribute values that match the 'attribute' part of the regexp will be retrieved from JMX (so only the InUseCount, AvailableCount and ActiveCount attribute will be retrieved from JMX in this example)

About correctness: It passes the unit tests
About regexp evaluations: There will be less now, since the complete rulelist is only evaluated once per mbean now instead of once per attribute

Anyway, for performance optimizations i prefer measuring over guessing. At least for my usecases, the gains are substantial.

brian-brazil:
> (so only the InUseCount, AvailableCount and ActiveCount attribute will be retrieved from JMX in this example)

That's not correct, this pattern can match attributes other than these.

> About correctness: It passes the unit tests

The unittests are based on the current code, they haven't been extended to cover all the edge and corner cases your logic would need to deal with.

> At least for my usecases, the gains are substantial.

Your use case is not the only one.

janssk1:
> That's not correct, this pattern can match attributes other than these.

What other attributes do you mean ? Composite attributes ? Agreed, that could be. An mbean with attribute 'X' that has a composite field named inusecount will be ignored by the patch.

Given that there are border cases where this 'shortcut' won't work, i can make the behavior conditional to be on the safe side. 

Would you be willing to pull this in once I add that ?

brian-brazil:
> What other attributes do you mean ?

Consider that `pool><>(InUse|Available|Active)Count` could be the part of the value mBean as there's a `.*` in there.

> Would you be willing to pull this in once I add that ?

There's already a way to whitelist mBeans, so you're not adding that much on top of this. I think that getting this correct would be very complex, and I'm not currently convinced that such complexity is worth it.

PR #185, opened 2017-09-12
Support labels that contain tokens to be replaced with sysprops and env vars

jmazzitelli:
@brian-brazil 

This allows the JMX exporter to be told to create label names or values that contain values derived from Java system properties and operating system environment variables.

If a label (name or value) has within it a token in the form `${env.ABC}` it will be replaced with the value of the environment variable "ABC". If no environment variable named "ABC" exists then the token is replaced with a default value or empty string if no default is specified. To define a default, place it after the environment variable name separated with a colon, e.g. `${env.ABC:my-default}`.

The analogous functionality exists for system properties. e.g. `${sysprop.XYZ}` or `${sysprop.XYZ:my-default}`.

If a given label name or value has multiple tokens, they are all replaced.

For example, if a label name or value is defined as:

`user_${env.USER}_something_${sysprop.foo.bar}_${env.UNK:xyz}`

and the environment variable "USER" has a value of "bob", the Java system property "foo.bar" has a value of "123", and there is no environment variable named "UNK" then the label name or value will be:

`user_bob_something_123_xyz`

Another artificial, but more realistic, use case is:

```
labels:
  user: ${env.USER:anonymous}
  datacenter: ${sysprop.data.center}
  ${env.CORP_NAME}_${env.CORP_GROUP}: uxf_${env.CORP_ID:unknown}
```

This is useful when you want to attach additional metadata to your JMX rules such that your timeseries can be further refined based on system properties or environment variables set in the MBeanServer JVM. I can also see this being useful within Kubernetes or OpenShift pods running the Java app (passing in env vars or sys props defined by the pod template to help make the timeseries unique across pods, as an example).

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -68,6 +68,7 @@ public class JmxCollector extends Collector implements Collector.Describable {
     private long createTimeNanoSecs = System.nanoTime();
 
     private static final Pattern snakeCasePattern = Pattern.compile("([a-z0-9])([A-Z])");
+    private static final Pattern propertyTokenPattern = Pattern.compile("\\$\\{(env\\.|sysprop\\.)([^\\}]+)\\}");
 
     public JmxCollector(File in) throws IOException, MalformedObjectNameException {
         configFile = in;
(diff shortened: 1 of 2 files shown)


brian-brazil:
In the Prometheus ecosystem we leave things like this up to configuration management, rather than creating another layer of configuration management/templating.

In addition, the labels in you examples are all target labels rather than instrumentation labels and thus they belong on the Prometheus rather than the exporter side. I can think of no valid use cases for this offhand.

PR #195, opened 2017-10-10
make test platform-independent

Griesbacher:
The original Integration-Test would not work on Windows systems.


brian-brazil:
Thanks!

PR #198, opened 2017-10-13
Add an example of weblogic configuration.

Yannig:
This PR is just an example of jmx_exporter working with WebLogic.

The tricky part is in the jmxUrl format.

With this configuration, you can retrieve almost all important internals metrics of WebLogic (datasource, jms, gc activity, JVM memory, etc.).

### DIFF
--- a/example_configs/weblogic.yml
+++ b/example_configs/weblogic.yml
@@ -1,0 +1,48 @@
+---
+# In order to get WebLogic mbeans, you need to add the following options in server's options:
+# -Djavax.management.builder.initial=weblogic.management.jmx.mbeanserver.WLSMBeanServerBuilder
+
+attrNameSnakeCase: true
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+whitelistObjectNames:
+  - "com.bea:Name=*,Type=ServerRuntime"
+  - "com.bea:ServerRuntime=*,Type=ApplicationRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=JDBCDataSourceRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=JMSDestinationRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=JDBCStoreRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=FileStoreRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=SAFRemoteEndpointRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=ThreadPoolRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=JMSRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=SAFRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=WorkManagerRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=MessagingBridgeRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=PersistentStoreRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=WebServerRuntime,*"
+
+rules:
+  # ex: com.bea<ServerRuntime=AdminServer, Name=default, ApplicationRuntime=moduleJMS, Type=WorkManagerRuntime><>CompletedRequests
+  - pattern: "^com.bea<ServerRuntime=(.+), Name=(.+), (.+)Runtime=(.*), Type=(.+)Runtime><>(.+):"
+    name: weblogic_$3_$5_$6
+    attrNameSnakeCase: true
+    labels:
+      runtime: $1
+      name: $2
+      application: $4
+
+  # ex: com.bea<ServerRuntime=AdminServer, Name=dsName, Type=JDBCDataSourceRuntime><>Metric
+  - pattern: "^com.bea<ServerRuntime=(.+), Name=(.+), Type=(.+)Runtime><>(.+):"
+    name: weblogic_$3_$4
+    attrNameSnakeCase: true
+    labels:
+      runtime: $1
+      name: $2
+
+  # ex: com.bea<ServerRuntime=AdminServer, Name=bea_wls_cluster_internal, Type=ApplicationRuntime><OverallHealthStateJMX>IsCritical
+  - pattern: "^com.bea<ServerRuntime=(.+), Name=(.+), Type=(.+)Runtime><(.+)>(.+):"
+    name: weblogic_$3_$4_$5
+    attrNameSnakeCase: true
+    labels:
+      runtime: $1
+      name: $2


example_configs/weblogic.yml
```diff
@@ -0,0 +1,60 @@
+---
+jmxUrl: service:jmx:iiop:///jndi/iiop://localhost:7001/weblogic.management.mbeanservers.runtime
```
brian-brazil: We always recommend running as a JMX agent, so this should not be present.

Yannig:
It run as an agent with the following command options:

```-javaagent:$DOMAIN_HOME/jmx_prometheus_javaagent-0.10.jar=7019:$DOMAIN_HOME/weblogic-jmx-exporter.yml```

If I remove the jmxUrl, I lost all interesting values about datasource, jms informations etc.

**Comment by Yannig** [comment]
An output result of this configuration: https://gist.github.com/Yannig/cb81fcdc44813ef12a93281fa92af222

File: example_configs/weblogic.yml [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+whitelistObjectNames:
+  - "com.bea:Name=*,Type=ServerRuntime"
+  - "com.bea:ServerRuntime=*,Type=ApplicationRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=JDBCDataSourceRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=JMSDestinationRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=SAFRemoteEndpointRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=ThreadPoolRuntime,*"
+
+rules:
+  - pattern: "^com.bea<ServerRuntime=.+, Name=(.+), Type=JDBCDataSourceRuntime><>(.+): (Running|Suspended|Shutdown|Overloaded|Unknown)$"
```
brian-brazil: As this is doing something a bit unusual, it'd be clearer to change the 2nd capture group to `State`

example_configs/weblogic.yml
```diff
@@ -0,0 +1,60 @@
...
+    attrNameSnakeCase: true
+    name: weblogic_messaging_$2
+    labels:
+      destination: $1
+
+  - pattern: "^com.bea<ServerRuntime=.+, Name=ThreadPoolRuntime, Type=ThreadPoolRuntime><>(.+):"
+    attrNameSnakeCase: true
+    name: weblogic_threadpool_$1
+
+  - pattern: "^com.bea<ServerRuntime=.+, Name=(.+), Type=ApplicationRuntime><HealthStateJMX>(.+): HEALTH_(.+)"
```
brian-brazil: These ones also

example_configs/weblogic.yml
```diff
@@ -0,0 +1,60 @@
...
+      name: $1
+      state: $3
+
+  - pattern: "^com.bea<ServerRuntime=.+, Name=(.+), Type=JDBCDataSourceRuntime><>(.+):"
+    attrNameSnakeCase: true
+    name: weblogic_datasource_$2
+    labels:
+      destination: $1
+
+  - pattern: "^com.bea<ServerRuntime=.+, Name=(.+), Type=JMSDestinationRuntime, JMSServerRuntime=.+><>(.+):"
```
brian-brazil: This isn't producing anything

example_configs/weblogic.yml
```diff
@@ -0,0 +1,60 @@
...
+rules:
+  - pattern: "^com.bea<ServerRuntime=.+, Name=(.+), Type=JDBCDataSourceRuntime><>(.+): (Running|Suspended|Shutdown|Overloaded|Unknown)$"
+    attrNameSnakeCase: true
+    name: weblogic_datasource_$2
+    value: 1
+    labels:
+      name: $1
+      state: $3
+
+  - pattern: "^com.bea<ServerRuntime=.+, Name=(.+), Type=JDBCDataSourceRuntime><>(.+):"
```
brian-brazil: Can you trim the `Count` off the end of these, where present? That should only be on Summary/Histograms, not Gauges.

example_configs/weblogic.yml
brian-brazil: This is still here

Yannig: If I remove this line, I get no metrics at all.

brian-brazil: Why not? What's non-standard with this application?

Yannig: Well, I don't know exactly. Maybe there's something wrong with WebLogic in the way the agent retrieve informations.
Is there something I can do about that?

brian-brazil: Is it using something other than the default mbean server?

Yannig: Good idea. If I add the following parameters ```-Djavax.management.builder.initial=weblogic.management.jmx.mbeanserver.WLSMBeanServerBuilder``` to my weblogic server I get all the weblogic metrics.
Is there any option I can use to specify this class?

brian-brazil: You can pass it on that command line, I'd suggest putting that information in a comment up the top of the file.

Yannig: Fine, I update my PR with this and your previous comments.
Thanks for your feedback!

Yannig:
Updated sample:  https://gist.github.com/Yannig/cb81fcdc44813ef12a93281fa92af222

example_configs/weblogic.yml
```diff
@@ -0,0 +1,49 @@
...
+  - "com.bea:ServerRuntime=*,Type=WebServerRuntime,*"
+
+rules:
+  #- pattern: ".+"
+  # ex: com.bea<ServerRuntime=AdminServer, Name=default, ApplicationRuntime=moduleJMS, Type=WorkManagerRuntime><>CompletedRequests
+  - pattern: "^com.bea<ServerRuntime=(.+), Name=(.+), (.+)Runtime=(.*), Type=(.+)Runtime><>(.+):"
+    name: weblogic_$3_$5_$6
+    attrNameSnakeCase: true
+    labels:
+      instance: $1
```
brian-brazil: Instance is already used by Prometheus, maybe runtime instead?

Yannig: Done.

example_configs/weblogic.yml
```diff
@@ -0,0 +1,49 @@
...
+  - "com.bea:ServerRuntime=*,Type=ThreadPoolRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=JMSRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=SAFRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=WorkManagerRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=MessagingBridgeRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=PersistentStoreRuntime,*"
+  - "com.bea:ServerRuntime=*,Type=WebServerRuntime,*"
+
+rules:
+  #- pattern: ".+"
```
brian-brazil: This line shouldn't be here

Yannig: Done.

brian-brazil:
Thanks!

Yannig:
Nice :)

PR #199, opened 2017-10-14
Update example config for Spark

alonl:
* Fixed regex escaping
* Add Structured Streaming metrics

@brian-brazil

### DIFF
--- a/example_configs/spark.yml
+++ b/example_configs/spark.yml
@@ -1,32 +1,41 @@
 rules:
+
   # These come from the master
   # Example: master.aliveWorkers
-  - pattern: "metrics<name=master.(.*)><>Value"
+  - pattern: "metrics<name=master\\.(.*)><>Value"
     name: spark_master_$1
 
   # These come from the worker
   # Example: worker.coresFree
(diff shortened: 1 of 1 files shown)


alonl:
1. The current example didn't even work - so I escaped the dots in the metrics' names and now it's working.
2. Added scraping for Spark Structured Streaming metrics (See: https://issues.apache.org/jira/browse/SPARK-17731)
3. Regarding `lowercaseOutputName` and `attrNameSnakeCase`, I was actually not 100% sure about that.

example_configs/spark.yml
```diff
@@ -1,40 +1,53 @@
+---
+lowercaseOutputName: true
```
brian-brazil: I wouldn't add these, I'm starting to feel that munging names causes more confusion than it does clarity

alonl: Delete both `lowercaseOutputName` and `attrNameSnakeCase` or just the former?

brian-brazil: Both please, they weren't here before.

alonl:
@brian-brazil I've updated the PR.

brian-brazil:
Thanks!

PR #202, opened 2017-10-25
Prevent hanging on exit

### LINKED ISSUE #170: Java agent preventing startup
After upgrading from jmx_exporter 0.8 to 0.10, the java agent is now blocking Cassandra (3.0) from starting up. The metrics endpoint is serving metrics properly, but there is nothing related to cassandra in there as it has not yet started and is not accepting connections.

gpaggi:
Fixes #170


brian-brazil:
Thanks!

Issue #111, opened 2017-02-14
Have some example rules of generic JVM metrics?


EdSchouten:
Hi there,

We're making use of the JMX exporter in combination with Tomcat and Cassandra. While using it, we've noticed that none of the examples contain any generic rules for extracting properties out of the JVM, like memory usage. Right now we're planning on adding something like this to our configs:

```
- pattern: 'java.lang<type=Memory><(\w+)MemoryUsage>(\w+):'
  name: java_memory_usage_$2_bytes
  labels:
    area: "$1"
  help: Java $2 $1 memory usage
  type: GAUGE
```

Resulting in metrics like these:

```
java_memory_usage_committed_bytes{area="Heap",} 1.489731584E9
java_memory_usage_committed_bytes{area="NonHeap",} 3.9223296E8
java_memory_usage_max_bytes{area="Heap",} 4.294770688E9
java_memory_usage_max_bytes{area="NonHeap",} 4.17333248E8
java_memory_usage_init_bytes{area="Heap",} 2.73888048E8
java_memory_usage_init_bytes{area="NonHeap",} 3.69557504E8
java_memory_usage_used_bytes{area="Heap",} 9.61119192E8
java_memory_usage_used_bytes{area="NonHeap",} 2.09771528E8
```

Question: would it make sense to have some kind of examples file containing rules like these?

brian-brazil:
These shouldn't be in the examples, as the standard exports of the agent already provide these metrics in a better format.

EdSchouten:
Oh, wow. Running it separately or as an agent determines which metrics are exported by default? Interesting. Is this documented somewhere? What kind of metrics are exported by default when running as an agent?

brian-brazil:
Those metrics should always be exported, but they're not enabled for the http server as they'd be confusing.

EdSchouten:
Should we document this somewhere?

wfindley:
Oh.  I just ran into this.  How can I induce the http server to do this?

brian-brazil:
Run it as a java agent.

hejix:
Hi Brian,  There are many good reasons to run the jmx_exporter separately as a server- the main one being able to change the metrics captured without having to restart the underlying process.  Would appreciate either being able to turn on the standard jvm metrics or knowing the rules to enable them as mentioned by earlier posters.

brian-brazil:
> the main one being able to change the metrics captured without having to restart the underlying process.

That doesn't require a restart.

[...]

PR #204, opened 2017-11-06
Document that running as an agent is required for generic JVM metrics.

EdSchouten:
As discussed in issue #111, it is not intended that the JMX exporter
extracts generic JVM metrics when running as a separate process. These
values are only exposed when running the exporter as an agent as part of
the process to be monitored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,15 @@
 JMX Exporter
 =====
 
-JMX to Prometheus exporter.
+JMX to Prometheus exporter: a collector that can configurably scrape and
+expose mBeans of a JMX target.
 
-A Collector that can configurably scrape and expose mBeans of a JMX target. It
-meant to be run as a Java Agent, exposing an HTTP server and scraping the local
-JVM.
-
-This can be also run as an independent HTTP server and scrape remote JMX targets.
+This exporter is intended to be run as a Java Agent, exposing a HTTP server
+and serving metrics of the local JVM. It can be also run as an independent
+HTTP server and scrape remote JMX targets, but this has various
+disadvantages, such as being harder to configure and being unable to expose
+process metrics (e.g., memory and CPU usage). Running the exporter as a Java
+Agent is thus strongly encouraged.
 
 ## Running
 


README.md
```diff
@@ -1,13 +1,13 @@
...
+expose mBeans of a JMX target.
 
-A Collector that can configurably scrape and expose mBeans of a JMX target. It
-meant to be run as a Java Agent, exposing an HTTP server and scraping the local
-JVM.
-
-This can be also run as an independent HTTP server and scrape remote JMX targets.
+This exporter is intended to be run as a Java Agent, exposing a HTTP server
+and serving metrics of the local JVM. It can be also run as an independent
+HTTP server and scrape remote JMX targets, but this has the downside of not
```
brian-brazil: It has more downsides than that, you also lose process metrics and configuration&management is much more challenging. Most of our support requests are from users needlessly using the http server, so it should be more strongly discouraged.

EdSchouten: If we're discouraging it so strongly, wouldn't it be smarter to either deprecate this functionality or at least document that it's unsupported?

brian-brazil: There are valid use cases for it, as some users (largely for political rather than technical reasons) can't use the agent. Thus it remains a supported feature.

EdSchouten: Got it. What do you think of this instead?

README.md
```diff
@@ -1,13 +1,15 @@
...
-JVM.
-
-This can be also run as an independent HTTP server and scrape remote JMX targets.
+JMX to Prometheus exporter: a collector that can configurably scrape and
+expose mBeans of a JMX target.
+
+This exporter is intended to be run as a Java Agent, exposing a HTTP server
+and serving metrics of the local JVM. It can be also run as an independent
+HTTP server and scrape remote JMX targets, but this has various
+disadvantages, such as the inability to expose core JVM metrics (e.g.,
```
brian-brazil: technically you can get all the JVM metrics, as they're just mBeans. Process metrics you can't though.

I'd mention harder to configure for the http server too.

EdSchouten: What do you think of this version?

brian-brazil:
Thanks!

PR #206, opened 2017-11-09
Add a note on slow /metrics endpoint

### LINKED ISSUE #175: Metrics endpoint is slow
I've added jmx_exporter to a WildFly 10 setup and notices that the `/metrics` endpoint is pretty slow (in relation to various other endpoints from applications and exporters). For me `/metrics` returns data in around 7-10 seconds (without any user load on the server) and would occasionally exceed the prometheus default scrape timeout of 10 seconds.

If this is normal (I'm new to all this Java/JMX/WildFly stuff) a short note in the readme would be cool (at least for people like me :smiley: ).

jayme-github:
Sorry, I did not realize that I created this PR for the wrong repo.

fixes #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ To bind the java agent to a specific IP change the port number to `host:port`.
 
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
 
+Please note that due to the nature of JMX the `/metrics` endpoint might exceed Prometheus default scrape timeout of 10 seconds.
+
 ## Building
 
 `mvn package` to build.


README.md
```diff
@@ -21,6 +21,8 @@ To bind the java agent to a specific IP change the port number to `host:port`.
 
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
 
+Please note that the `/metrics` endpoint is expected to be slow (~10 seconds) and might exceed Prometheus default scrape timeout of 10 seconds.
```
brian-brazil: That really depends on what you're monitoring

jayme-github: Okay. My interpretation of your reply to #175 was that ~10 seconds are expected. Would you be okay with it when I just remove `(~10 seconds)`?

brian-brazil: 10+ seconds is expected for Cassandra, but other things can be fast. I'd word it as a possibility rather than a certainty, and indicate that this is due to the nature of JMX.

jayme-github: What about:
> Please note that due to the nature of JMX the `/metrics` endpoint might exceed Prometheus default scrape timeout of 10 seconds.

brian-brazil: Sounds perfect

brian-brazil:
Thanks!

finkr:
_JMX is not particularly efficient_ (quote Brian Brazil)

However, using `whitelistObjectNames`, It is often possible to dramatically **speed-up things** (dropped from 15 seconds to 0.9s. in my use case).
See ->  https://github.com/prometheus/jmx_exporter/pull/284

Wouldn't it be worth to add this fact to README?

PR #208, opened 2017-11-14
Fix missing metrics

niuhf0452:
NullPointerException would be throw if a tabular contains null key. And doScrape() break out. Some metrics will never be scraped.

### DIFF
--- a/collector/src/test/java/io/prometheus/jmx/JmxCollectorTest.java
+++ b/collector/src/test/java/io/prometheus/jmx/JmxCollectorTest.java
@@ -248,11 +248,10 @@ public class JmxCollectorTest {
     }
 
     @Test
-    public void testNullTabularValue() throws Exception {
-        MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
-        Tabular.registerBean(mbs);
-        JmxCollector jc = new JmxCollector("---\nstartDelaySeconds: 1").register(registry);
-        Thread.sleep(2000);
-        assertEquals(1, registry.getSampleValue("tabular_Test_Foo", new String[]{}, new String[]{}), .001);
+    public void testNullTabularKey() throws Exception {
+      MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
+      Tabular.registerBean(mbs);
+      JmxCollector jc = new JmxCollector("").register(registry);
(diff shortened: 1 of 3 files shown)


niuhf0452:
@brian-brazil

collector/src/test/java/io/prometheus/jmx/TabularMXBean.java
```diff
@@ -0,0 +1,28 @@
...
+
+class Tabular implements TabularMXBean {
+    public static void registerBean(MBeanServer mbs) throws javax.management.JMException {
+        ObjectName mbeanName = new ObjectName("tabular:type=Test");
+        Tabular mbean = new Tabular();
+        mbs.registerMBean(mbean, mbeanName);
+    }
+
+    public Map<String, String> getTable() {
+        return Collections.singletonMap(null, "null key");
```
brian-brazil: This seems to be testing a null key rather than a null value

collector/src/test/java/io/prometheus/jmx/JmxCollectorTest.java
```diff
@@ -246,4 +246,13 @@ public void testDelayedStartReady() throws Exception {
       Thread.sleep(2000);
       assertEquals(1.0, registry.getSampleValue("boolean_Test_True", new String[]{}, new String[]{}), .001);
     }
+
+    @Test
+    public void testNullTabularValue() throws Exception {
+        MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
+        Tabular.registerBean(mbs);
+        JmxCollector jc = new JmxCollector("---\nstartDelaySeconds: 1").register(registry);
```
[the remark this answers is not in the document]
niuhf0452: My mistake. Delay is unnecessary.

collector/src/test/java/io/prometheus/jmx/TabularMXBean.java
niuhf0452: Yes. It causes error, not null value.

brian-brazil: I'm not understanding, I don't see how this tests the code change.

brian-brazil:
Superseded by #253

PR #215, opened 2017-12-09
Fixed tests for Windows

snuyanzin:
In Windows the symbol ':' is a part of filename while at the same time it plays a delimiter role for java agent => for Windows the logic in the class io.prometheus.jmx.JavaAgent is broken (tests are also failed) To fix it it is required to take into account that in case of Windows there will be 1 more ':'

### DIFF
--- a/jmx_prometheus_javaagent/pom.xml
+++ b/jmx_prometheus_javaagent/pom.xml
@@ -34,7 +34,6 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>hamcrest</groupId>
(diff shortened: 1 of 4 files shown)


brian-brazil:
This change would break relative paths on Windows.

snuyanzin:
Yes  you are right. Thank you
For that reason added one more test for relative paths

jmx_prometheus_javaagent/pom.xml
```diff
@@ -34,7 +34,6 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
-      <scope>test</scope>
```
brian-brazil: This should stay as-is

brian-brazil:
I believe this is now obsolete.

PR #217, opened 2017-12-15
Enhancements on jmx-exporter

lucasponce:
JmxScrape uses queryMBeans as fetching method for MBeans.
It happens that some servers *might* filter some MBeans on queryMBeans, so these MBeans are only accesible via getObjectInstance.
Wildfly has a case for jboss.as:management-root=server MBean which is special and it stores important info for monitoring.
So, a proposal to handle this corner cases is to introduce a new pair of  configuration properties called whitelistObjectInstances/blacklistObjectInstances with the similar use than whitelistObjectNames/blacklistObjectNames but on these cases MBeans are fetched directly (so, no patterns allowed).
On this case, there is backward compatibility with previous behaviour but we can extend these special use cases.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -57,6 +57,8 @@ public class JmxCollector extends Collector implements Collector.Describable {
       boolean ssl = false;
       boolean lowercaseOutputName;
       boolean lowercaseOutputLabelNames;
+      List<ObjectName> whitelistObjectInstances = new ArrayList<ObjectName>();
+      List<ObjectName> blacklistObjectInstances = new ArrayList<ObjectName>();
       List<ObjectName> whitelistObjectNames = new ArrayList<ObjectName>();
       List<ObjectName> blacklistObjectNames = new ArrayList<ObjectName>();
       ArrayList<Rule> rules = new ArrayList<Rule>();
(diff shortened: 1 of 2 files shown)


brian-brazil:
Is that in line with the JMX spec? How many applications behave in this way?

I'm wary of adding features to accommodate one application that doesn't follow the normal rules, that's better handled with a bug against that application.

lucasponce:
> I'm wary of adding features to accommodate one application that doesn't follow the normal rules, that's better handled with a bug against that application.

I understand and yes, it should not be "special" MBeans, but it happens.
I also think this proposal is clean, as user may whitelist specific MBeans, without using queryMBeans.

brian-brazil:
> I understand and yes, it should not be "special" MBeans, but it happens.

Have you filed bugs against these applications?

> I also think this proposal is clean, as user may whitelist specific MBeans, without using queryMBeans.

Adding workaround for a bug rather than fixing the bug doesn't sound very clean to me.

lucasponce:
> Have you filed bugs against these applications?
This issue is discussed with Wildfly engineers and a JIRA is on the way.

> Adding workaround for a bug rather than fixing the bug doesn't sound very clean to me.
The workaround is to run a full scan in all MBeans of a Wildfly server, which has a negative impact on the performance of the overall system.
This proposal enhances the jmx-exporter allowing to directly fetch the MBeans needed.
So, in tradeoff terms, I recommend this approach.

lucasponce:
Related JIRA for this PR
https://issues.jboss.org/browse/WFCORE-3461

brian-brazil:
> The workaround is to run a full scan in all MBeans of a Wildfly server, which has a negative impact on the performance of the overall system.

The JMX exporter walks all the MBeans anyway. If you provide a monitoring/control system you can't be surprised when users then use it.

lucasponce:
> The JMX exporter walks all the MBeans anyway. If you provide a monitoring/control system you can't be surprised when users then use it.

As commented, with this approach, you don't have to scrape all MBeans to collect, it's an extension of the whitelistObject mechanism.

brian-brazil:
Looks like wildfly fixed this on their end.

PR #218, opened 2017-12-18
Cassandra example improvements

JensRantil:


example_configs/cassandra.yml
```diff
@@ -2,12 +2,19 @@
...
+    destination: "$5"
+    "$1": "$4"
+    "$2": "$3"
+- pattern: org.apache.cassandra.metrics<type=(Connection|Streaming), scope=(\S*), name=(\S*)><>(?:Count|Value)
   name: cassandra_$1_$3
   labels:
     address: "$2"
 - pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?,
-    name=(\S*)><>(Count|Value)
+    name=(\S*)><>(?:Count|Value)
```
brian-brazil: This subgroup is used

example_configs/cassandra.yml
```diff
@@ -2,12 +2,19 @@
 lowercaseOutputLabelNames: true
 lowercaseOutputName: true
 rules:
-- pattern: org.apache.cassandra.metrics<type=(Connection|Streaming), scope=(\S*), name=(\S*)><>(Count|Value)
+- pattern: org.apache.cassandra.metrics<type=(HintedHandOffManager)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?,
```
brian-brazil: Can you give an example of the before and after?

JensRantil: See commit message for a092b3e517f3d655a3e0e07340b65bbf3f2c55b9.

example_configs/cassandra.yml
JensRantil: AFAIK, it's not. I can't find a `$5`.

brian-brazil: It's in the name.

example_configs/cassandra.yml
brian-brazil: This config doesn't seem to match the example output. Are you sure this is what you're using?

If you could share the help string that'd help my understanding.

example_configs/cassandra.yml
JensRantil: Meh, I see now. My apologies.

JensRantil: Force-pushed a fix.

brian-brazil:

PR #222, opened 2017-12-31
Support for IPv6 bind

bobrik:
This enables bind to IPv6 address in square brackets:

```
-javaagent:/usr/share/jmx_exporter/agent.jar=[f00::123]:9097:/etc/kafka/jmx-exporter.yaml
```

### DIFF
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -13,6 +13,17 @@ public class JavaAgent {
    static HTTPServer server;
 
    public static void premain(String agentArgument, Instrumentation instrumentation) throws Exception {
+     // Bind to all interfaces by default (this includes IPv6).
+     String host = "0.0.0.0";
+
+     // If we have IPv6 address in square brackets, extract it first and then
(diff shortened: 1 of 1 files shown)


jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -13,6 +13,13 @@
    static HTTPServer server;
 
    public static void premain(String agentArgument, Instrumentation instrumentation) throws Exception {
+     String host = "0.0.0.0";
```
brian-brazil: Will this work as expected for v6?

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -13,6 +13,13 @@
...
+     Integer indexOfClosingSquareBracket = agentArgument.indexOf("]:");
```
brian-brazil: A comment here would be useful.

bobrik:
Added comments.

brian-brazil:
Thanks!

bobrik:
Thanks for the merge! Do you mind cutting a new release?

PR #224, opened 2018-01-10
Added Apache ActiveMQ Artemis 2.x example

n1ko-w1ll:
@brian-brazil I added an example for Apache ActiveMQ Artemis. It only works for version newer than 2 because they changed the JMX MBean object names in this version.

Best regards
Niko

### DIFF
--- a/example_configs/artemis-2.yml
+++ b/example_configs/artemis-2.yml
@@ -6,20 +6,16 @@ rules:
     attrNameSnakeCase: true
     name: artemis_$2
     type: COUNTER
-    labels:
-        host: $1
   - pattern: "^org.apache.activemq.artemis<broker=\"([^\"]*)\",\\s*component=addresses,\\s*address=\"([^\"]*)\"><>([^:]*):\\s(.*)"
     attrNameSnakeCase: true
     name: artemis_$3
     type: COUNTER
     labels:
-        host: $1
         address: $2
   - pattern: "^org.apache.activemq.artemis<broker=\"([^\"]*)\",\\s*component=addresses,\\s*address=\"([^\"]*)\",\\s*subcomponent=(queue|topic)s,\\s*routing-type=\"([^\"]*)\",\\s*(queue|topic)=\"([^\"]*)\"><>([^: ]*):\\s(.*)"
     attrNameSnakeCase: true
     name: artemis_$7
     type: COUNTER
     labels:
-        host: $1
         address: $2
         "$5": $6
--- a/example_configs/artemis-2.yml
+++ b/example_configs/artemis-2.yml
@@ -1,0 +1,25 @@
+---
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+rules:
+  - pattern: "^org.apache.activemq.artemis<broker=\"([^\"]*)\"><>([^:]*):\\s(.*)"
+    attrNameSnakeCase: true
+    name: artemis_$2
+    type: COUNTER
+    labels:
+        host: $1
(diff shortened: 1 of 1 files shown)


example_configs/artemis-2.yml
```diff
@@ -0,0 +1,25 @@
+---
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+rules:
+  - pattern: "^org.apache.activemq.artemis<broker=\"([^\"]*)\"><>([^:]*):\\s(.*)"
+    attrNameSnakeCase: true
+    name: artemis_$2
+    type: COUNTER
+    labels:
+        host: $1
```
brian-brazil: A host label is unusual. Can you share example output?

n1ko-w1ll: Sure:
```
artemis_message_count{address="ExpiryQueue",host="infra1-live",} 0.0
artemis_message_count{queue="ExpiryQueue",address="ExpiryQueue",host="infra1-live",} 0.0
artemis_message_count{queue="DLQ",address="DLQ",host="infra1-live",} 0.0
artemis_message_count{queue="notif.cf762ce6-f5df-11e7-ab73-0050568f7196.ActiveMQServerImpl_serverUUID=71f2d8ca-e3f6-11e7-9700-0050568f7196",address="activemq.notifications",host="infra1-live",} 0.0
artemis_message_count{address="DLQ",host="infra1-live",} 0.0
artemis_message_count{address="activemq.notifications",host="infra1-live",} 0.0
artemis_message_count{queue="$.artemis.internal.sf.example-cluster.71f2d8ca-e3f6-11e7-9700-0050568f7196",address="$.artemis.internal.sf.example-cluster.71f2d8ca-e3f6-11e7-9700-0050568f7196",host="infra1-live",} 0.0
artemis_message_count{address="$.artemis.internal.sf.example-cluster.71f2d8ca-e3f6-11e7-9700-0050568f7196",host="infra1-live",} 0.0
```
This is the output of one Artemis broker. The `host` label is the broker name (could be renamed to that if you want).

brian-brazil: Is the broker name the broker that we're talking to?

n1ko-w1ll: yes

brian-brazil: That should be removed then, as that's already covered by the instance label.

n1ko-w1ll: The instance label is IP address and port which is hard to identify when several instances are running on the same machine. The broker name does not need to be the host name. It can be configured for each Artemis broker instance.

brian-brazil: That's still a matter for service discovery on Prometheus's end.

n1ko-w1ll: Sorry, I'm no expert for Prometheus. How would I lookup a specific broker without having it's name as label?

n1ko-w1ll:
@brian-brazil I removed the `host` label.

brian-brazil:
Thanks!

sugoisugoiJR:
@n1ko-w1lll Do you all attributes on puruse to counter?
For example:
The Uptime attribute would not be an counter or TotalMessageCount would be also a gauge, because TotalMessagesAdded would be counter.

n1ko-w1ll:
@sugoisugoiJR this project ended for me more than one year ago and after that I had no chance to look at it again. AFAIK it's still in use like this but I don't know how useful these metrics are in the daily business :(

sugoisugoiJR:
@n1ko-w1ll Thank you for your answer. (=

Issue #233, opened 2018-01-23
Reduce jmx_exporter's mBean regex loading


afalko:
From the mailing list conversation: https://groups.google.com/forum/#!topic/prometheus-developers/EDou8gu4Qvs

While profiling some applications on I've noticed that the following regex shows up very high in my top functions list: https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxScraper.java#L177

To make matters worse, when my applications get highly loaded, the scrape time grows linearly. That regex is invoked on every scrape, slows down my systems, and causes delayed metrics feedback. The crux of the problem is that getKeyPropertyList is created on every scrape: https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxScraper.java#L161 . I'd like that to happen a lot less frequently.

Solution to this is currently being discussed on the mailing list. Refer to the link at the top to follow it.

PR #235, opened 2018-01-24
Create cache for mBean key properties.

afalko:
Part 1 fix for issue #233 

- [x] Utests for whitelist clearing


afalko:
@brian-brazil This fix did make a different to the performance issue identified, but I'll need to do a part 2 to reduce instances of this: https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxCollector.java#L312

I think it'll be better to keep that fix separate.

brian-brazil:
That's got a memory leak, as old values are never deleted.

Have you considered writing a small parser rather than using a regex?

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -108,17 +96,19 @@ public void doScrape() throws Exception {
...
-            Set<ObjectInstance> mBeanNames = new HashSet();
+            Set<ObjectInstance> mBeanInstances = new HashSet<ObjectInstance>();
             for (ObjectName name : whitelistObjectNames) {
-                mBeanNames.addAll(beanConn.queryMBeans(name, null));
+                mBeanInstances.addAll(beanConn.queryMBeans(name, null));
             }
             for (ObjectName name : blacklistObjectNames) {
-                mBeanNames.removeAll(beanConn.queryMBeans(name, null));
+                Set<ObjectInstance> queriedMBeans = beanConn.queryMBeans(name, null);
+                jmxMBeanPropertyCache.removeMBeans(queriedMBeans);
```
afalko: @brian-brazil I tried to cover the memleak here. Did you have something else in mind?

afalko:
I'm considering one for part 2 that I mentioned above. Even the best written parser in the part 1 case will cause a slow down compared to this cache.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
brian-brazil: The leak is when the mBeans change from scrape to scrape, which this doesn't cover.

brian-brazil:
What would a parser be relative to the existing code? What proportion of the time is your use case spending in this code?

afalko:
I'm not sure yet how I could build a parser for part 2. The user input there is a regex itself. The only thing I can create a parser for is the snippet that sets things to snake case; but that'll be a micro-optimization at best. 

Before any changes: 
20% of all time, including the app (Kafka) was spent crunching through the regexes. If I break it down by regexes in 
https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxScraper.java#L177 (part 1)
vs.
https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxCollector.java#L312 (part 2)

Then part 1 takes 25% vs. 75% for part 2. i.e. in aggregate 5% for part 1, 15% for part 2.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
afalko: Gotcha; I think I can fix this by keeping a list of queryMBeans that get collected via whilelist loop above. I'd compare that list to what it was before and call removeMBean for anything still in the list from previous run.

afalko: See my updated change; I clear the mBeans if they are no longer in whitelist; I still need to test this out though. Is a good way to test this case to remove things from whitelist in the config and verify that mBean gets removed? I'll write up some utests for this too if you consider the approach agreeable. 

Do you have an app that has mBeans chaning from scrape-to-space that I can test with?

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -108,17 +89,27 @@ public void doScrape() throws Exception {
...
+            Set<ObjectName> mBeanNames = new HashSet<ObjectName>();
             for (ObjectName name : whitelistObjectNames) {
-                mBeanNames.addAll(beanConn.queryMBeans(name, null));
+                for (ObjectInstance instance : beanConn.queryMBeans(name, null)) {
+                    mBeanNames.add(instance.getObjectName());
+                }
             }
+
+            // Now that we have the whitelisted mBeans, remove any old ones from the cache:
+            jmxMBeanPropertyCache.onlyKeepMBeans(mBeanNames);
```
brian-brazil: Why not do this after the the blacklisting?

afalko: That'll make it better, thanks!

collector/src/test/java/io/prometheus/jmx/JmxMBeanPropertyCacheTest.java
```diff
@@ -0,0 +1,139 @@
...
+    @Test
+    public void testIdempotentGet() throws Throwable {
+        JmxMBeanPropertyCache testCache = new JmxMBeanPropertyCache();
+        ObjectName testObjectName = new ObjectName("com.organisation:name=value");
+        LinkedHashMap<String, String> parameterListFirst = testCache.getKeyPropertyList(testObjectName);
+        LinkedHashMap<String, String> parameterListSecond = testCache.getKeyPropertyList(testObjectName);
+        assertEquals(parameterListFirst, parameterListSecond);
+    }
+
+    @Test
```
afalko: I added three utests here @brian-brazil

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -69,13 +69,21 @@
 
     private static final Pattern snakeCasePattern = Pattern.compile("([a-z0-9])([A-Z])");
 
+    private final JmxMBeanPropertyCache jmxMBeanPropertyCache;
```
brian-brazil: Wouldn't this make more sense in the Scraper?

collector/src/main/java/io/prometheus/jmx/JmxMBeanPropertyCache.java
```diff
@@ -0,0 +1,77 @@
...
+    // Implement a version of ObjectName.getKeyPropertyList that returns the
+    // properties in the ordered they were added (the ObjectName stores them
+    // in the order they were added).
+    private final Map<ObjectName, LinkedHashMap<String, String>> keyPropertiesPerBean;
+
+    public JmxMBeanPropertyCache() {
+        this.keyPropertiesPerBean = new ConcurrentHashMap<ObjectName, LinkedHashMap<String, String>>();
+    }
+
+    Map<ObjectName, LinkedHashMap<String, String>> getKeyPropertiesPerBean() {
```
brian-brazil: This is only used in testing, why not make the field protected?

collector/src/main/java/io/prometheus/jmx/JmxMBeanPropertyCache.java
```diff
@@ -0,0 +1,77 @@
...
+        return keyPropertiesPerBean;
+    }
+
+    public LinkedHashMap<String, String> getKeyPropertyList(ObjectName mbeanName) {
+        if (keyPropertiesPerBean.containsKey(mbeanName)) {
```
brian-brazil: get will return null if it is empty, no need for the separate contains. You've also a race here due to this.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
afalko: I'm happy to change that, but my read of the code was that, Scraper is initialized in `public List<MetricFamilySamples> collect()`. Since that function is public, `collect()` might be invoked multiples times; so sharing the cache would provide most benefit there. I'm happy to move to scraper if you think that is best.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -69,13 +69,21 @@
...
+
+    private JmxCollector() {
+        jmxMBeanPropertyCache = new JmxMBeanPropertyCache();
```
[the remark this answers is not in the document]
afalko: Good point, I'll move it there. Out of habit, I initialize in constructor so that I can more easily mock objects out. In this case it is just a map behind the scenes, so doubt we'd need to mock anything.

collector/src/main/java/io/prometheus/jmx/JmxMBeanPropertyCache.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This object stores a mapping of mBean objectNames to mBean key property lists. The main purpose of it is to reduce
+ * the frequency with which we invoke PROPERTY_PATTERN when discovering mBeans.
+ */
+public class JmxMBeanPropertyCache {
```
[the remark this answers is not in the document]
afalko: Good point, will fix.

collector/src/main/java/io/prometheus/jmx/JmxMBeanPropertyCache.java
afalko: No modifier makes it slight more "protected" than than `protected`.

collector/src/main/java/io/prometheus/jmx/JmxMBeanPropertyCache.java
afalko: 👍 great suggest and thanks for the review!

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
brian-brazil: Right you are.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -2,14 +2,7 @@
...
 import java.lang.management.ManagementFactory;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
```
brian-brazil: Please leave imports explicit

afalko: Fixed

brian-brazil:
Thanks!

PR #236, opened 2018-01-30
Replace expensive snakeCase regex transform

afalko:
A small improvement to address https://github.com/prometheus/jmx_exporter/issues/233

@brian-brazil You can see the difference here for yourself: https://github.com/afalko/testrepo/blob/master/TimingBench.java

The bottom result is the new method:
```
$ java TimingBench
start_time_$1_$2
507
start_time_$1_$2
149
```

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -67,8 +67,6 @@ public class JmxCollector extends Collector implements Collector.Describable {
     private File configFile;
     private long createTimeNanoSecs = System.nanoTime();
 
-    private static final Pattern snakeCasePattern = Pattern.compile("([a-z0-9])([A-Z])");
-
     private final JmxMBeanPropertyCache jmxMBeanPropertyCache = new JmxMBeanPropertyCache();
 
     public JmxCollector(File in) throws IOException, MalformedObjectNameException {
@@ -221,6 +219,24 @@ public class JmxCollector extends Collector implements Collector.Describable {
 
     }
 
+    static String toSnakeAndLowerCase(String attrName) {
+      if (attrName == null || attrName.isEmpty()) {
+        return attrName;
+      }
+      char firstChar = attrName.subSequence(0, 1).charAt(0);
+      boolean prevCharIsUpperCaseOrUnderscore = Character.isUpperCase(firstChar) || firstChar == '_';
+      StringBuilder resultBuilder = new StringBuilder().append(Character.toLowerCase(firstChar));
+      for (char attrChar : attrName.substring(1).toCharArray()) {
+        boolean charIsUpperCase = Character.isUpperCase(attrChar);
+        if (!prevCharIsUpperCaseOrUnderscore && charIsUpperCase) {
+          resultBuilder.append("_");
+        }
+        resultBuilder.append(Character.toLowerCase(attrChar));
+        prevCharIsUpperCaseOrUnderscore = charIsUpperCase || attrChar == '_';
+      }
+      return resultBuilder.toString();
+    }
+
     class Receiver implements JmxScraper.MBeanReceiver {
       Map<String, MetricFamilySamples> metricFamilySamplesMap =
         new HashMap<String, MetricFamilySamples>();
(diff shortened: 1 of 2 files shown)


collector/src/test/java/io/prometheus/jmx/SnakeCaseAttrTest.java
```diff
@@ -0,0 +1,41 @@
...
+    private final String input;
+
+    public SnakeCaseAttrTest(String expected, String input) {
+        this.expected = expected;
+        this.input = input;
+    }
+
+    @Test
+    public void testAttrToSnakeAndLowerCase() {
+        JmxCollector.Receiver rc = mock(JmxCollector.Receiver.class);
```
afalko: I can't use the constructor because I can't access it via non-static context.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -297,6 +297,21 @@ private void defaultExport(
           type, help);
       }
 
+      String toSnakeAndLowerCase(String attrName) {
+        if (attrName == null || attrName.isEmpty()) {
+          return attrName;
+        }
+        StringBuilder resultBuilder = new StringBuilder(attrName.substring(0,1).toLowerCase());
```
brian-brazil: space after comma

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -297,6 +297,21 @@ private void defaultExport(
...
+        for (char attrChar : attrName.substring(1).toCharArray()) {
+          if (Character.isUpperCase(attrChar)) {
```
brian-brazil: This isn't doing exactly the same thing. Consider TCP for which should become tcp but this function would return t_c_p.

collector/src/test/java/io/prometheus/jmx/SnakeCaseAttrTest.java
brian-brazil: Why not make the function static?

afalko: I can solve it that way if you're OK with it. Another way to solve it is to pull it out into its own file.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
afalko: I'll fix this; I'll add these two tests:
```
TCP -> tcp
testTCPTest -> test_tcptest
```

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -221,7 +219,7 @@ private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObject
 
     }
 
-    class Receiver implements JmxScraper.MBeanReceiver {
+    static class Receiver implements JmxScraper.MBeanReceiver {
```
afalko: Setting this static doesn't work. 

```
[ERROR] /home/afalko/jmx_exporter/collector/src/main/java/io/prometheus/jmx/JmxCollector.java:[275,12] error: non-static variable config cannot be referenced from a static context
```

I'd have to make this static too:
```
private Config config;
```

afalko: I've made all the other fixes and rebased. I've updated the change to make config static. I normally avoid `static` as much as possible and either mock or pull nested classes out.

afalko:
Latest iteration is even faster (I think because I switch to use `subsequence` for the first char):

Regex: start_time_$1_$2
508
Non-regex: start_time_$1_$2
152
Underscore handling: start_time_$1_$2
131

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
brian-brazil: Just make toSnakeAndLowerCase static. It doesn't use any fields of the class.

afalko: Fixed, but had to move it to parent class b/c static is not allowed in inner classes.

brian-brazil:
Thanks!

PR #238, opened 2018-02-04
add support for bulk jmx attributes request

kevsanders:
Hi Brian,
If you give this a spin to see if it is of interest - I left an ugly feature toggle so you can try before/after
I found that scraping from my remote jmx bean was very slow (probably because I was mistakenly running across a WAN) so I added this.  As I mentioned the latency dropped from 255 seconds to 11 for me.
I have not extensively tested it

Regards
Kevin


brian-brazil:
The java agent is the recommended way to use this exporter, what difference does it make these?

kevsanders:
running the javaagent from trunk

bulk version
jmx_scrape_duration_seconds 0.109733737
jmx_scrape_duration_seconds 0.099713492
jmx_scrape_duration_seconds 0.095011474
jmx_scrape_duration_seconds 0.09212449
jmx_scrape_duration_seconds 0.094129824

original
jmx_scrape_duration_seconds 0.158636148
jmx_scrape_duration_seconds 0.140345366
jmx_scrape_duration_seconds 0.123538856
jmx_scrape_duration_seconds 0.120296189
jmx_scrape_duration_seconds 0.111793064

brian-brazil:
So a 25% performance improvement, which is nice. Which server was this against?

Can you make this a proper diff?

kevsanders:
not sure what you mean by proper diff?

below the script I ran to create the diffs (from my fork)

cd /cygdrive/d/sources/_github/forks/jmx_exporter

for toggle in false true
do

  java -DBULK_FETCH=$toggle -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=5555 -jar jmx_prometheus_httpserver/target/jmx_prometheus_httpserver-0.2.1-SNAPSHOT-jar-with-dependencies.jar 5556 example_configs/httpserver_sample_config.yml &
  pid=$!
  echo "started agent with pid: $pid for toggle: $toggle"

  >$toggle.txt
  count=5
  while (( --count >= 0 )); 
  do 
    curl "http://localhost:5556/metrics" 2>&1 /dev/null|grep jmx_scrape_duration_seconds|grep -v "#" |tee -a $toggle.txt
  done
  
  $(kill $pid) 2>&1 /dev/null
  $(jobs) 2>&1 /dev/null

done

#compare samples
old=$(awk '{sum+=$NF}END{print sum;}' false.txt)
new=$(awk '{sum+=$NF}END{print sum;}' true.txt)
echo "diff $old -> $new"

brian-brazil:
Please remove the big if statement, it's too difficult to review your code currently.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -135,31 +137,38 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) {
...
+                processBeanValue(
+                        mbeanName.getDomain(),
+                        jmxMBeanPropertyCache.getKeyPropertyList(mbeanName),
+                        new LinkedList<String>(),
+                        attr.getName(),
+                        attr.getType(),
+                        attr.getDescription(),
+                        attribute.getValue()
+                );
+            }else {
```
brian-brazil: space after }

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -135,31 +137,38 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) {
...
-            } catch(Exception e) {
-                logScrape(mbeanName, attr, "Fail: " + e);
-                continue;
+            name2AttrInfo.put(attr.getName(), attr);
+        }
+        AttributeList attributes;
+        try {
+            attributes = beanConn.getAttributes(mbeanName, name2AttrInfo.keySet().toArray(new String[0]));
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to read attributes " + name2AttrInfo.keySet() + " from " + mbeanName);
```
brian-brazil: Handle things like this gracefully, many JMX servers are broken and attributes can change

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -135,21 +137,31 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) {
...
-            } catch(Exception e) {
-                logScrape(mbeanName, attr, "Fail: " + e);
-                continue;
+            name2AttrInfo.put(attr.getName(), attr);
+        }
+        AttributeList attributes = new AttributeList();
+        try {
+            attributes = beanConn.getAttributes(mbeanName, name2AttrInfo.keySet().toArray(new String[0]));
+        } catch (Exception e) {
+            //couldn't get them all in one go, try them 1 by 1
```
brian-brazil: Why do you need this?

kevsanders: following your review comment
"Handle things like this gracefully, many JMX servers are broken and attributes can change"
...rather than throw the exception to say I couldn't get any attributes
I thought why not try them individually (just incase one attribute is adversely affecting the others - ie revert back to the original behaviour) on the up side you may get at least some of the attributes back on the down side you will get a longer list of error messages

brian-brazil: That's not the failure mode I was worried about, you can just log here.

kevsanders: do you mean - don't try the one at a time approach - just log the failure and continue?

brian-brazil: Yes

kevsanders: ok - will be able to check that in a bit later

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -135,21 +137,23 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) {
...
-            try {
-                value = beanConn.getAttribute(mbeanName, attr.getName());
-            } catch(Exception e) {
-                logScrape(mbeanName, attr, "Fail: " + e);
-                continue;
-            }
-
+            name2AttrInfo.put(attr.getName(), attr);
+        }
+        AttributeList attributes = new AttributeList();
```
brian-brazil: The object created here isn't used

kevsanders: oh - i thought this was setting up an empty collection so that if the getAttributes call fails then we can just log and continue.  then the next block can safely iterate over the empty coolection

brian-brazil: You can return in the catch

brian-brazil:
Thanks!

PR #243, opened 2018-02-10
Require JDK 6 for compiling and running

marcelmay:
Switch to JDK 6 as baseline for all modules.

This fixes the following issues, and simplifies project setup (no JDK baseline mix, project wide Maven compiler plugin configuration):

- The collector module declares JDK 5 target but uses JDK 6 API:
  In collector/src/main/java/io/prometheus/jmx/JmxCollector.java L223, L343, L370, L395 :
  [isEmpty() is JDK 6 String API](https://docs.oracle.com/javase/6/docs/api/java/lang/String.html#isEmpty())

- The collector module also uses simpleclient, which compiles for JDK 6.

- The module jmx_prometheus_httpserver also compiles for JDK 5, but requires collector module. Therefore min version is also JDK 6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ To bind the java agent to a specific IP change the port number to `host:port`.
 
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
 
+Note: Requires minimum JDK version 1.6
+
 ## Building
 
 `mvn package` to build.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -23,6 +23,8 @@ To bind the java agent to a specific IP change the port number to `host:port`.
 
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
 
+Note: Requires minimum JDK version 1.6
```
brian-brazil: I'm not sure this needs to be mentioned here, even 6 is fairly old by now.

marcelmay: Removed the note and updated PR.

brian-brazil:
Thanks!

PR #248, opened 2018-02-22
add bat version of httpserver run script for dev purpose

hwasin:
Hi. Think we can add run_sample_httpserver.bat to run standalone extractor on windows environments for development purposes.

@brian-brazil

### DIFF
--- a/run_sample_httpserver.bat
+++ b/run_sample_httpserver.bat
@@ -1,0 +1,5 @@
+@echo off
+:: Script to run a java application for testing jmx4prometheus.
+
+:: Note: You can use localhost:5556 instead of 5556 for configuring socket hostname.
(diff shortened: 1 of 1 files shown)


brian-brazil:
What sort of development are you doing that involves using this script? It's only intended as a simple demonstration, and it's expected you'll use it as a java agent on your own JVMs.

hwasin:
As i see we can use http_server to scrape remote JMX targets without restarting them, for example cassandra, and java-agent is not applicable in this case. In my sandbox i start 2 cassandra nodes,  start http_server after some time and scrape metrics using http_server

brian-brazil:
That's not the recommended way to use the jmx exporter, and in any case I don't see how this PR helps you with it.

hwasin:
Ok. Thank you. One more question: Is http_server mode planned to be supported in the future?

brian-brazil:
There's no plans to drop it.

Issue #252, opened 2018-03-09
JMX scrape failed: java.lang.NullPointerException (JmxScraper.java:240)


juliohm1978:
I'm facing an issue with version 0.3.0, where JmxScraper throws with the following exception:

```
SEVERE: JMX scrape failed: java.lang.NullPointerException
	at io.prometheus.jmx.JmxScraper.processBeanValue(JmxScraper.java:240)
	at io.prometheus.jmx.JmxScraper.scrapeBean(JmxScraper.java:159)
	at io.prometheus.jmx.JmxScraper.doScrape(JmxScraper.java:117)
	at io.prometheus.jmx.JmxCollector.collect(JmxCollector.java:456)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.findNextElement(CollectorRegistry.java:183)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:216)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:137)
	at io.prometheus.client.exporter.common.TextFormat.write004(TextFormat.java:22)
	at io.prometheus.client.exporter.HTTPServer$HTTPMetricHandler.handle(HTTPServer.java:59)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:79)
	at sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:83)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:82)
	at sun.net.httpserver.ServerImpl$Exchange$LinkHandler.handle(ServerImpl.java:675)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:79)
	at sun.net.httpserver.ServerImpl$Exchange.run(ServerImpl.java:647)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

I managed to debug the code and worked around it with a simple fix to `JmxScraper.java`. If anyone needs an urgent fix, the following patch should work.

```
Index: collector/src/main/java/io/prometheus/jmx/JmxScraper.java
IDEA additional info:
[...]

PR #253, opened 2018-03-09
Fix NullPointerException at JmxScraper.java:240

juliohm1978:
Follow up to #252. This PR fixes a NullPointerException at JmxScraper.java:240.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -237,7 +237,9 @@ class JmxScraper {
                 if (valu instanceof CompositeData) {
                     CompositeData composite = (CompositeData) valu;
                     for (String idx : rowKeys) {
-                        l2s.put(idx, composite.get(idx).toString());
+                        if ( composite.get(idx) != null ) {
+                            l2s.put(idx, composite.get(idx).toString());
+                        }
                     }
                     for(String valueIdx : valueKeys) {
                         LinkedList<String> attrNames = extendedAttrKeys;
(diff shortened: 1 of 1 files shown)


collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -237,7 +237,9 @@ private void processBeanValue(
                 if (valu instanceof CompositeData) {
                     CompositeData composite = (CompositeData) valu;
                     for (String idx : rowKeys) {
-                        l2s.put(idx, composite.get(idx).toString());
+                        if ( composite.get(idx) != null ) {
```
brian-brazil: You don't need the whitespace inside the parens

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -237,7 +237,9 @@ private void processBeanValue(
...
+                            l2s.put(idx, composite.get(idx).toString());
```
brian-brazil: You've a race here, this get might return null.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -237,7 +237,9 @@ private void processBeanValue(
                 if (valu instanceof CompositeData) {
                     CompositeData composite = (CompositeData) valu;
                     for (String idx : rowKeys) {
-                        l2s.put(idx, composite.get(idx).toString());
+                        if ((composite != null) && (composite.get(idx) != null)) {
```
brian-brazil: I don't see how composite can be null here, it'd have failed the instanceof. What you want to do is call `composite.get(idx)` only once.

juliohm1978: You are right. I misunderstood your last comment. Fixing it now.

brian-brazil:
Thanks!

Issue #249, opened 2018-02-26
jmx_prometheus_httpserver is unable to connect to a remote Wildfly on port 9990 - Unsupported protocol: http-remoting-jmx


juliohm1978:
I'm trying to collect metrics from a remote Wildfly 8.2.1, and for this particular instance I cannot configure the agent locally at that server. Management port 9990, however, is available.

Running jmx_prometheus_httpserver with the following configuration:

```
java -cp jboss-client.jar -jar jmx_prometheus_httpserver-0.2.1-SNAPSHOT-jar-with-dependencies.jar 127.0.0.1:28686 jmx_exporter.yaml
```

This is what `jmx_exporter.yaml` looks like:

```
startDelaySeconds: 0
jmxUrl: service:jmx:http-remoting-jmx://<remote_wildfly_host>:9990
username: myuser
password: mypassword
ssl: false
```

The server starts well, but as soon as I try to scrape metrics, the following error comes up.

```
Feb 26, 2018 4:45:32 PM io.prometheus.jmx.JmxCollector collect
SEVERE: JMX scrape failed: java.net.MalformedURLException: Unsupported protocol: http-remoting-jmx
	at javax.management.remote.JMXConnectorFactory.newJMXConnector(JMXConnectorFactory.java:359)
	at javax.management.remote.JMXConnectorFactory.connect(JMXConnectorFactory.java:269)
	at io.prometheus.jmx.JmxScraper.doScrape(JmxScraper.java:94)
	at io.prometheus.jmx.JmxCollector.collect(JmxCollector.java:456)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.findNextElement(CollectorRegistry.java:180)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.<init>(CollectorRegistry.java:144)
	at io.prometheus.client.CollectorRegistry.filteredMetricFamilySamples(CollectorRegistry.java:131)
	at io.prometheus.client.exporter.HTTPServer$HTTPMetricHandler.handle(HTTPServer.java:60)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:79)
	at sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:83)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:82)
	at sun.net.httpserver.ServerImpl$Exchange$LinkHandler.handle(ServerImpl.java:675)
[...]

PR #254, opened 2018-03-10
Make jmx_exporter compatible with Willdfly 8/9/10

juliohm1978:
Following up on @SakaSun's suggestion from #249, I was able to modify the `maven-assembly-plugin` to create a new jar-with-dependencies with the necessary deps to make jmx_exporter compatible with Wildfly 8/9/10.

This patch consists of further customization to the `maven-assembly-plugin`, by providing two assembly descriptors `src/assembly/assembly.xml` and `src/assembly/assembly-wildfly.xml`. This creates different jars in different plugin executions.

I'm hoping to keep backwards compatibility by creating a new build target called `jmx_prometheus_httpserver-*-jar-with-dependencies-wildfly.jar`. It's slightly bigger in size because it includes a few more dependencies. Anyone who does not need that, can still to use the standard `jmx_prometheus_httpserver-*-jar-with-dependencies.jar`, which is still built as usual with `src/assembly/assembly.xml`.


brian-brazil:
I'd suggest keeping this code in another repository, such as Wildfly's. I don't want to maintain N different special cases for applications that decided to provide non-standard JMX implementations.

juliohm1978:
Sounds reasonable. That's not a problem for me at all. My team can easily keep a modified build in-house for our use case.

Feel free to keep this in mind if you change your mind in the future.

PR #259, opened 2018-03-16
Use the version in pom.xml to run the sample HTTP server

dlespiau:
`0.2.1-SNAPSHOT` didn't cut it anymore. Use the current `SNAPSHOT` version from `pom.xml` to always be up to date.


brian-brazil:
Thanks!

PR #260, opened 2018-03-16
README: Update example usage with version 0.3.0

dlespiau:
Both the download link and command line example were still using 0.2.0.


brian-brazil:
Thanks!

PR #264, opened 2018-04-03
Add Presto example

piyushnarang:
Spent a bit of time hunting around for examples and how to hook up Presto with the jmx_exporter. Thought it might be of use to others in the future. I've included some of the ones in the list here (https://github.com/prestodb/presto/wiki/Monitoring) and a few others. I've left out the java.lang metrics, but I can include them if needed.

### DIFF
--- a/example_configs/presto.yml
+++ b/example_configs/presto.yml
@@ -1,48 +1,38 @@
 lowercaseOutputName: true
 lowercaseOutputLabelNames: true
-whitelistObjectNames: ["com.facebook.presto.hive:type=NamenodeStats,*", "com.facebook.presto.hive.metastore:type=ThriftHiveMetastore,*", "com.facebook.presto.failureDetector:name=HeartbeatFailureDetector,*", "com.facebook.presto.execution:name=QueryManag\
-er,*", "com.facebook.presto.execution:name=TaskManager,*", "com.facebook.presto.memory:type=ClusterMemoryPool,*", "com.facebook.presto.memory:name=ClusterMemoryManager,*"]
-blacklistObjectNames: []
 rules:
   - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+): (.*)"
-    name: "presto:hive_namenodestats_$1"
+    name: "presto hive_namenodestats_$1"
     help: "Presto Hive: NameNodeStats $1"
-    value: $2
     type: GAUGE
     attrNameSnakeCase: false
-  - pattern: "com.facebook.presto.hive.metastore<type=ThriftHiveMetastore, name=hive><>(.+): (.*)"
-    name: "presto:hivemetastorestats_$1"
+  - pattern: "com.facebook.presto.hive.metastore.thrift<type=ThriftHiveMetastore, name=hive><>(.+): (.*)"
+    name: "presto hivemetastorestats_$1"
     help: "Presto Hive: Metastore Stats $1"
-    value: $2
     type: GAUGE
     attrNameSnakeCase: false
   - pattern: "com.facebook.presto.execution<name=TaskManager><>(.+): (.*)"
-    name: "presto:taskmanager_$1"
+    name: "presto taskmanager_$1"
     help: "Presto: TaskManager Stats $1"
-    value: $2
     type: GAUGE
     attrNameSnakeCase: false
   - pattern: "com.facebook.presto.failureDetector<name=HeartbeatFailureDetector><>ActiveCount: (.*)"
-    name: "presto:heartbeatdetector_activecount"
+    name: "presto heartbeatdetector_activecount"
     help: "Presto: HeartBeat Detector ActiveCount"
-    value: $1
     type: GAUGE
     attrNameSnakeCase: false
   - pattern: "com.facebook.presto.execution<name=QueryManager><>(.+): (.*)"
-    name: "presto:querymanager_$1"
+    name: "presto querymanager_$1"
     help: "Presto: QueryManager $1"
-    value: $2
     type: GAUGE
     attrNameSnakeCase: false
   - pattern: "com.facebook.presto.memory<name=ClusterMemoryManager><>(.+): (.*)"
-    name: "presto:clustermemorymanager_$1"
+    name: "presto clustermemorymanager_$1"
     help: "Presto: ClusterMemoryManager $1"
-    value: $2
     type: GAUGE
     attrNameSnakeCase: false
   - pattern: "com.facebook.presto.memory<type=ClusterMemoryPool, name=(.*)><>(.+): (.*)"
-    name: "presto:clustermemorypool_$1_$2"
+    name: "presto clustermemorypool_$1_$2"
     help: "Presto ClusterMemoryPool: $1 $2"
-    value: $3
     type: GAUGE
     attrNameSnakeCase: false
--- a/example_configs/presto.yml
+++ b/example_configs/presto.yml
@@ -1,0 +1,48 @@
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+whitelistObjectNames: ["com.facebook.presto.hive:type=NamenodeStats,*", "com.facebook.presto.hive.metastore:type=ThriftHiveMetastore,*", "com.facebook.presto.failureDetector:name=HeartbeatFailureDetector,*", "com.facebook.presto.execution:name=QueryManag\
+er,*", "com.facebook.presto.execution:name=TaskManager,*", "com.facebook.presto.memory:type=ClusterMemoryPool,*", "com.facebook.presto.memory:name=ClusterMemoryManager,*"]
+blacklistObjectNames: []
+rules:
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+): (.*)"
+    name: "presto:hive_namenodestats_$1"
+    help: "Presto Hive: NameNodeStats $1"
+    value: $2
+    type: GAUGE
+    attrNameSnakeCase: false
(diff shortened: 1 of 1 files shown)


piyushnarang:
@brian-brazil - do take a look when you get a chance

example_configs/presto.yml
```diff
@@ -0,0 +1,48 @@
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+whitelistObjectNames: ["com.facebook.presto.hive:type=NamenodeStats,*", "com.facebook.presto.hive.metastore:type=ThriftHiveMetastore,*", "com.facebook.presto.failureDetector:name=HeartbeatFailureDetector,*", "com.facebook.presto.execution:name=QueryManag\
+er,*", "com.facebook.presto.execution:name=TaskManager,*", "com.facebook.presto.memory:type=ClusterMemoryPool,*", "com.facebook.presto.memory:name=ClusterMemoryManager,*"]
+blacklistObjectNames: []
+rules:
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+): (.*)"
+    name: "presto:hive_namenodestats_$1"
+    help: "Presto Hive: NameNodeStats $1"
+    value: $2
```
brian-brazil: You shouldn't need this

example_configs/presto.yml
```diff
@@ -0,0 +1,48 @@
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+whitelistObjectNames: ["com.facebook.presto.hive:type=NamenodeStats,*", "com.facebook.presto.hive.metastore:type=ThriftHiveMetastore,*", "com.facebook.presto.failureDetector:name=HeartbeatFailureDetector,*", "com.facebook.presto.execution:name=QueryManag\
+er,*", "com.facebook.presto.execution:name=TaskManager,*", "com.facebook.presto.memory:type=ClusterMemoryPool,*", "com.facebook.presto.memory:name=ClusterMemoryManager,*"]
+blacklistObjectNames: []
+rules:
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+): (.*)"
+    name: "presto:hive_namenodestats_$1"
```
brian-brazil: metric names from instrumentation should not contain colons

example_configs/presto.yml
```diff
@@ -0,0 +1,48 @@
+lowercaseOutputName: true
```
brian-brazil: I'm tending towards avoiding these these days.

example_configs/presto.yml
```diff
@@ -0,0 +1,48 @@
...
+lowercaseOutputLabelNames: true
+whitelistObjectNames: ["com.facebook.presto.hive:type=NamenodeStats,*", "com.facebook.presto.hive.metastore:type=ThriftHiveMetastore,*", "com.facebook.presto.failureDetector:name=HeartbeatFailureDetector,*", "com.facebook.presto.execution:name=QueryManag\
```
brian-brazil: Does this make a large enough difference to be worth maintaining it as new mbeans are added?

example_configs/presto.yml
piyushnarang: Ok, will drop it.

example_configs/presto.yml
piyushnarang: Thanks, will update.

example_configs/presto.yml
piyushnarang: I don't have strong preferences on this. Noticed it had been set to lower in a couple of the other examples and liked the consistent casing is generated so I added it. If you'd like I can drop this.

example_configs/presto.yml
piyushnarang: Yeah that's a good question. We are explicitly pattern matching the beans we are interested in so only those seem to be making it through. I'll drop this in favor of the rules.

piyushnarang:
@brian-brazil - updated. I've got a dump of the output in this gist: https://gist.github.com/piyushnarang/fd55b489c60467d0374644459fce3c17
(In my output I've only got the oneminute metrics as that's what we're interested in. The PR doesn't filter them as I wasn't sure if other users also only care about one min metrics)

example_configs/presto.yml
piyushnarang: ok, I can drop this.

example_configs/presto.yml
```diff
@@ -0,0 +1,38 @@
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+rules:
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+): (.*)"
+    name: "presto hive_namenodestats_$1"
+    help: "Presto Hive: NameNodeStats $1"
+    type: GAUGE
+    attrNameSnakeCase: false
```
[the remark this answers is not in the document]
piyushnarang: ok, I've dropped it

brian-brazil:
Just the alltime. I'm wondering now what it's doing under the covers, as I haven't seen that pattern of metric names before.

piyushnarang:
They are using this stats library: https://github.com/airlift/airlift/tree/master/stats/src/main/java/io/airlift/stats
An example is: https://github.com/airlift/airlift/blob/master/stats/src/main/java/io/airlift/stats/DistributionStat.java

brian-brazil:
Okay, the alltime_count is a counter so we can use that.

piyushnarang:
So looking at this a bit more. I don't think it's the right value though. A given time statistic (e.g. `listlocatedstatus` which measure latency to a HDFS api call) has the oneMinute, 5min, 15 min and all time stats. Each of these stats objects (1min .. allTime) is a distribution. The count variable within that though is the number of samples added to the QuantileDigest. I don't think users will want that, they'll want some latency snapshot for that API call. The options that we have are to provide those at the 1min, 5min, 15min or all time level. And then within each of these time range buckets - we have various percentiles - min, max, p50, p75, p90, p95, p99. 

There's also a couple of others - `CounterStat` - this tracks totalCount, oneMinute decaying count, 5 min decaying count and a 15 min decaying count. This is used for metrics like ioExceptions / failures talking to HDFS etc. 

Another is `DistributionStat` which is tracking a 1min (decaying), 5min (decaying), 15min (decaying) and allTime distributions (p01, p5, p10 .. p99, min, max). This is used for metrics like bytesRead from a given source. 

So it seems like we could expose the full `allTime` distribution + `totalCount` values. Couple of potential concerns / questions:
1) I'm not sure I know what the end user experience would be like when the application restarts though. I don't think these values are being persisted. So on restarts a lot of the total count values will drop to 0.
2) I'm also not sure if the all time distributions will capture intermittent spikes. I'd imagine the 1min / 5 min / 15 min time range buckets would capture that (unless Prometheus will handle this). 

The other option is for us to expose all these metrics and leave it up to users to decide which ones they pick up and use. It really depends on what the ultimate purpose of these example configs is.

brian-brazil:
> The count variable within that though is the number of samples added to the QuantileDigest. I don't think users will want that

The count is useful.

> The options that we have are to provide those at the 1min, 5min, 15min or all time level.

You want the all time. Combined with the count you can then calculate average latency over any time period.

> Another is DistributionStat which is tracking a 1min (decaying), 5min (decaying), 15min (decaying) and allTime distributions (p01, p5, p10 .. p99, min, max).

That's fairly limited. There's a count in there.

> So on restarts a lot of the total count values will drop to 0

Prometheus is designed to handle that.

> I'm also not sure if the all time distributions will capture intermittent spikes.

This is one of the reasons why exponentially decaying averages are not the best option.

> The other option is for us to expose all these metrics and leave it up to users to decide which ones they pick up and use.

We should try to only keep stuff that is useful.

piyushnarang:
Ok, thanks for clarifying @brian-brazil. I've updated to include the AllTime / TotalCount metrics for the beans that have them. I ended up using a `COUNTER` for the TotalCounts as if I understand the prometheus docs correctly that the type to use for increasing values.

brian-brazil:
What does the output look like now?

piyushnarang:
@brian-brazil  - put it in a gist here if you'd like to take a look (https://gist.github.com/piyushnarang/3fa46279fe332bab110b44c76a432733). Our setup also has some JVM based stats that we expose which I've not included in the PR.

example_configs/presto.yml
```diff
@@ -0,0 +1,41 @@
+rules:
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+AllTime.+): (.*)"
+    name: "presto hive_namenodestats_$1"
+    help: "Presto Hive: NameNodeStats $1"
+    type: GAUGE
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+TotalCount.*): (.*)"
+    name: "presto hive_namenodestats_$1"
+    help: "Presto Hive: NameNodeStats $1"
+    type: COUNTER
+  - pattern: "com.facebook.presto.hive.metastore.thrift<type=ThriftHiveMetastore, name=hive><>(.+AllTime.+): (.*)"
```
brian-brazil: This is pulling in the percentile too. You can also remove the "alltime" from the output, as it doesn't add anything.

example_configs/presto.yml
```diff
@@ -0,0 +1,41 @@
+rules:
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+AllTime.+): (.*)"
+    name: "presto hive_namenodestats_$1"
+    help: "Presto Hive: NameNodeStats $1"
+    type: GAUGE
+  - pattern: "com.facebook.presto.hive<type=NamenodeStats, name=hive><>(.+TotalCount.*): (.*)"
+    name: "presto hive_namenodestats_$1"
```
brian-brazil: _total is the usual suffix for a counter.

brian-brazil:

PR #275, opened 2018-05-27
Correct main class name in README example

vromero:


kpolszewski:
👍 i had the same problem

brian-brazil:
The jar name isn't right either.

brian-brazil:

PR #279, opened 2018-06-06
Registering JMX Exporter build info

### LINKED ISSUE #278: Add build info metric
It would be really helpful to have something like `jmx_exporter_build_info` (e.g. with the jar version), so that it'd be easier to track which JVMs are running which versions of the exporter.

@brian-brazil if you don't object, we might have a go at implementing this.

picadoh:
By adding the Implementation-Version manifest entry it is possible
to retrieve the lib version at runtime by checking the class Package.

This allows the tracking of which JVMs are running which versions of
the exporter.

Closes #278

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/BuildInfoCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/BuildInfoCollector.java
@@ -3,9 +3,10 @@ package io.prometheus.jmx;
 import io.prometheus.client.Collector;
 import io.prometheus.client.GaugeMetricFamily;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
 
-import static java.util.Collections.singletonList;
+import static java.util.Arrays.asList;
 
 /**
  * Collects jmx_exporter build version info.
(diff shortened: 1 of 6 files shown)


picadoh:
Hi @brian-brazil. Would you please review this PR regarding #278?

brian-brazil:
Does this have to apply only to the agent, could it not work for the http server too?

picadoh:
@brian-brazil I was not able to retrieve the implementation-version from the http server package in the same way. Some research is required. I may open a new PR for that as soon as I find a solution for it. Is this ok?

picadoh:
@brian-brazil Actually, it works for the HTTP Server as well if we run through the fat jar, as I'm able to inject the version into the manifest through the assembly plugin. Shall I proceed with the changes to allow it in this PR?

brian-brazil:
Please go ahead.

collector/src/main/java/io/prometheus/jmx/BuildInfoCollector.java
```diff
@@ -0,0 +1,40 @@
+package io.prometheus.jmx;
+
+import io.prometheus.client.Collector;
+import io.prometheus.client.GaugeMetricFamily;
+
+import java.util.*;
```
brian-brazil: Please be explicit in your imports

picadoh:
@brian-brazil I pushed the changes to enable this to HTTP Server as well. Also, besides the version, I included the name as it allows to distinguish between both the agent or the HTTP server. Could you please check?

brian-brazil:
Thanks!

PR #280, opened 2018-06-11
Make the agent attachable to a running VM

picadoh:
By adding the agentmain method, we enable the ability of attaching
the JMX Exporter Java Agent to a JVM after the VM starts, providing
a more dynamic approach to managing thousands of deployed VMs.

Usage example:
```
VirtualMachine vm = VirtualMachine.attach(processPID);
vm.loadAgent("./jmx_prometheus_javaagent-0.3.2.jar=8080:config.yaml");
vm.detach();
```

Important to note that this change does not break any of the existing behaviours of the agent nor it adds any extra responsibility to it, it only enables this ability on top of it.

@brian-brazil Can you please review? Could this be something that we could add to the agent?

### DIFF
--- a/jmx_prometheus_javaagent/pom.xml
+++ b/jmx_prometheus_javaagent/pom.xml
@@ -82,6 +82,7 @@
             <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
              <manifestEntries>
               <Premain-Class>io.prometheus.jmx.shaded.io.prometheus.jmx.JavaAgent</Premain-Class>
+              <Agent-Class>io.prometheus.jmx.shaded.io.prometheus.jmx.JavaAgent</Agent-Class>
               <Implementation-Version>${project.version}</Implementation-Version>
               <Implementation-Title>${project.artifactId}</Implementation-Title>
              </manifestEntries>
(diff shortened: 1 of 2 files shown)


brian-brazil:
What are you imagining the lifecycle as here? This seems pretty fragile to me due to the extra moving parts, a java agent does the same without that issue.

picadoh:
Thanks for your reply @brian-brazil. I'd like to highlight that this change does not alter the way JMX exporter works, but it makes it easy to attach it dynamically to a JVM (instead of only statically). That will facilitate the upgrade of the agent JAR in entire infrastructure thus improving the operational productivity of the teams.

brian-brazil:
Why not restart the applications? That's something you generally should have the capability to do.

picadoh:
The applications will be "eventually" restarted, but as soon as they are, the latest version of the agent will be attached automatically by another piece of infrastructure, removing the need to maintain the config management of these Java options.

brian-brazil:
That sounds like a sane, though advanced, setup. I say we change the pom, but keep this undocumented. I can see users tying themselves in knots with this feature.

picadoh:
Thanks @brian-brazil. Makes sense. I just made it undocumented.

brian-brazil:
Thanks!

PR #282, opened 2018-06-13
Upgrading compiler target version to 1.8

### LINKED ISSUE #281: Maven Compiler target version
The current default target version is 1.6 (parent pom.xml) but the jmx_exporter does not compile with 1.6. It requires at least 1.7 because of its dependencies.

Example here:

https://jitpack.io/com/github/prometheus/jmx_exporter/-parent-0.3.1-g2db4f0c-3/build.log
```
Caused by: java.lang.UnsupportedClassVersionError: org/codehaus/mojo/buildhelper/ReserveListenerPortMojo : Unsupported major.minor version 51.0
```

For tools that select the compiler to use based on the target version (e.g. Jitpack) this upgrade is necessary.

picadoh:
The current default target version is 1.6 (parent pom.xml) but
the jmx_exporter does not compile with 1.6. It requires at least
1.7 because of its dependencies.

Upgrading to 1.8 as it is the latest LTS version.

Closes #281

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,8 +55,8 @@
               <artifactId>maven-compiler-plugin</artifactId>
               <version>2.3.2</version>
               <configuration>
-                  <source>1.6</source>
-                  <target>1.6</target>
+                  <source>1.8</source>
+                  <target>1.8</target>
               </configuration>
           </plugin>
           <plugin>


brian-brazil:
This exporter is intended to work with Java 6. What dependency requires higher?

picadoh:
The artifact `org.codehaus.mojo:build-helper-maven-plugin:1.10` requires Java 7.

brian-brazil:
That's only used in unittests, so not the end of the world. Is there an older version of it that works with 6?

picadoh:
On a quick research it seems like 1.9.1 should be the last one compatible with Java 6.

picadoh:
Just tested it on my fork, thanks for the tip @brian-brazil. I raised PR #283 specifically for that. By accepting it we may reject this PR and close the issue associated with it.

PR #283, opened 2018-06-13
Downgrading maven builder helper plugin to 1.9.1

picadoh:
This downgrade allows JMX Exporter to keep compatible with Java 6.

The version 1.9.1 of build-helper-maven-plugin is the last version compatible with Java 6.

By accepting this change, issue #281 and PR #282 become not necessary and can be closed as well.


brian-brazil:
Thanks!

Issue #246, opened 2018-02-20
/metrics endpoint is way too slow under Wildfly 8


juliohm1978:
This was earlier reported on #175, which was closed by adding a footnote in the documentation.

I'm running Wildfly instances in a few Docker containers and the /metrics entpoint takes way too long. From inside the container:

```
$ time curl http://localhost:28686/metrics

## clipped output (metrics are returned as filtered)

real	1m28.087s
user	0m0.020s
sys	0m0.024s
```

We are having better results using Glassfish 3.1.2.2 or Wildfly 10. Any particular reason why this is taking so long under Wildfly 8? Any tips on how to filter metrics to make the endpoint faster?

Otherwise, we'll have to configure Prometheus with a `scrape_interval` of at least 2 minutes!

afalko:
Have you tried latest version version currently in master branch? Several perf improvements were made in master that haven't been released yet.

brian-brazil:
This is something to take up with Wildfly, we don't control how slow JMX endpoints are.

juliohm1978:
Just tried your suggestion and compiled a fresh version of the master branch: `jmx_prometheus_javaagent-0.2.1-SNAPSHOT.jar`.

The results are better, down to 30 seconds!

```
$ time curl http://localhost:28686/metrics

## clipped output (metrics are returned as filtered)

real	0m32.259s
user	0m0.012s
sys	0m0.016s
```

Still not ideal, but it does fall under an acceptable range for my team.

Thanks a lot!

n3v3rf411:
I'm having the same same issue as yours where the usual scrape time takes about 43 seconds even with the latest code. In my case, I already the know the object names in advance so I set `whitelistObjectNames` accordingly like as follows:
```
---
whitelistObjectNames: ["jboss.as:subsystem=undertow,server=*,http-listener=*"]
lowercaseOutputName: true
lowercaseOutputLabelNames: true
rules:
  - pattern: "^jboss.as<subsystem=undertow, server=(.+), http-listener=(.+)><>(bytes_.+|error_count|processing_time|request_count):"
    attrNameSnakeCase: true
    name: wildfly_undertow_$3
    labels:
      server: $1
      http_listener: $2
```

This improves scrape time from 43s to only 0.028s.

finkr:
I have submitted a PR to add whitelistObjectNames, see #284 

Thanks @n3v3rf411

PR #284, opened 2018-06-14
wildfly : Whitelist objects for performance reason

finkr:
see https://github.com/prometheus/jmx_exporter/issues/246#issuecomment-367573931
With this patch, the scrape time (`jmx_scrape_duration_seconds`) dropped from 15 seconds to 0.9s. (using with jmx_explorer 0.7)

I can therefore avoid increasing  `scrape_timeout` and `scrape_interval`.

Signed-off-by: Frank Lin Piat <fpiat@klabs.be>

### DIFF
--- a/example_configs/wildfly-10.yaml
+++ b/example_configs/wildfly-10.yaml
@@ -1,6 +1,17 @@
 ---
 lowercaseOutputName: true
 lowercaseOutputLabelNames: true
+whitelistObjectNames: 
+ # Whitelist objects to be collected, for performance reason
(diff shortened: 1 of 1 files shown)


brian-brazil:
You'll need to squash that now.

finkr:
Do you think whitelistObjectNames is the proper way to workaround JMX performance issue ?

What do you think about this PR ?
Should we carry on with this patch (improve/merge it), or do you think it's a bad idea for some reason and we should just close the PR ?

brian-brazil:
It's a valid workaround. I was just waiting on getting the DCO sorted out before merging, as project policy requires it.

PR #301, opened 2018-08-03
Add an example configuration for Kafka 2.0.0

ankon:
This is based on kafka-0_8_2.yml, with changes to better work with Kafka 2.0:

* "Histograms" are exposed as summaries (without the _sum)
  Note that this changes some metrics to now have a _count suffix compared to
  what kafka-0_8_2.yml would provide!
* Simplifications to capture the structure of "name, type" + "one or two labels"
  better

---
Here's a example list of metrics for comparison/understanding the available dimensions: 
[example-metrics.txt](https://github.com/prometheus/jmx_exporter/files/2257502/example-metrics.txt)

### DIFF
--- a/example_configs/kafka-2_0_0.yml
+++ b/example_configs/kafka-2_0_0.yml
@@ -1,0 +1,86 @@
+lowercaseOutputName: true
+
+rules:
+# Special cases and very specific rules
+- pattern : kafka.server<type=(.+), name=(.+), clientId=(.+), topic=(.+), partition=(.*)><>Value
+  name: kafka_server_$1_$2
+  type: GAUGE
+  labels:
+    clientId: "$3"
+    topic: "$4"
+    partition: "$5"
+- pattern : kafka.server<type=(.+), name=(.+), clientId=(.+), brokerHost=(.+), brokerPort=(.+)><>Value
+  name: kafka_server_$1_$2
+  type: GAUGE
+  labels:
+    clientId: "$3"
+    broker: "$4:$5"
+
+# Generic per-second counters with 0-2 key/value pairs
+- pattern: kafka.(\w+)<type=(.+), name=(.+)PerSec\w*, (.+)=(.+), (.+)=(.+)><>Count
+  name: kafka_$1_$2_$3_total
+  type: COUNTER
+  labels:
+    "$4": "$5"
+    "$6": "$7"
+- pattern: kafka.(\w+)<type=(.+), name=(.+)PerSec\w*, (.+)=(.+)><>Count
+  name: kafka_$1_$2_$3_total
+  type: COUNTER
+  labels:
+    "$4": "$5"
+  type: COUNTER
+- pattern: kafka.(\w+)<type=(.+), name=(.+)PerSec\w*><>Count
+  name: kafka_$1_$2_$3_total
+  type: COUNTER
+
+# Generic gauges with 0-2 key/value pairs
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+), (.+)=(.+)><>Value
+  name: kafka_$1_$2_$3
+  type: GAUGE
+  labels:
+    "$4": "$5"
+    "$6": "$7"
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+)><>Value
+  name: kafka_$1_$2_$3
+  type: GAUGE
+  labels:
+    "$4": "$5"
+- pattern: kafka.(\w+)<type=(.+), name=(.+)><>Value
+  name: kafka_$1_$2_$3
+  type: GAUGE
+
+# Emulate Prometheus 'Summary' metrics for the exported 'Histogram's.
+#
+# Note that these are missing the '_sum' metric!
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+), (.+)=(.+)><>Count
+  name: kafka_$1_$2_$3_count
+  type: COUNTER
+  labels:
+    "$4": "$5"
+    "$6": "$7"
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.*), (.+)=(.+)><>(\d+)thPercentile
+  name: kafka_$1_$2_$3
+  type: GAUGE
+  labels:
+    "$4": "$5"
+    "$6": "$7"
+    quantile: "0.$8"
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+)><>Count
+  name: kafka_$1_$2_$3_count
+  type: COUNTER
+  labels:
+    "$4": "$5"
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.*)><>(\d+)thPercentile
+  name: kafka_$1_$2_$3
+  type: GAUGE
+  labels:
+    "$4": "$5"
+    quantile: "0.$6"
+- pattern: kafka.(\w+)<type=(.+), name=(.+)><>Count
+  name: kafka_$1_$2_$3_count
+  type: COUNTER
+- pattern: kafka.(\w+)<type=(.+), name=(.+)><>(\d+)thPercentile
+  name: kafka_$1_$2_$3
+  type: GAUGE
+  labels:
+    quantile: "0.$4"


example_configs/kafka-2_0_0.yml
```diff
@@ -0,0 +1,81 @@
...
+  labels:
+    clientId: "$3"
+    broker: "$4:$5"
+
+- pattern: kafka.(\w+)<type=(.+), name=(.+)PerSec\w*, (.+)=(.+), (.+)=(.+)><>Count
+  name: kafka_$1_$2_$3_total
+  labels:
+    "$4": "$5"
+    "$6": "$7"
+  type: COUNTER
```
brian-brazil: If you're going to list it for some, list it for them all.

example_configs/kafka-2_0_0.yml
```diff
@@ -0,0 +1,81 @@
...
+    "$6": "$7"
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+)><>Value
+  name: kafka_$1_$2_$3
+  labels:
+    "$4": "$5"
+- pattern: kafka.(\w+)<type=(.+), name=(.+)><>Value
+  name: kafka_$1_$2_$3
+  type: COUNTER
+
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+), (.+)=(.+)><>Count
```
brian-brazil: A comment that this is emulating a summary would be useful, particularly as _sum is missing.

example_configs/kafka-2_0_0.yml
```diff
@@ -0,0 +1,81 @@
...
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+), (.+)=(.+)><>Value
+  name: kafka_$1_$2_$3
+  labels:
+    "$4": "$5"
+    "$6": "$7"
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+)><>Value
+  name: kafka_$1_$2_$3
+  labels:
+    "$4": "$5"
+- pattern: kafka.(\w+)<type=(.+), name=(.+)><>Value
```
brian-brazil: Are you sure this is a counter?

ankon:
Thanks for the feedback!

I've hopefully addressed your points, and reviewed the metric types to add them consistently.

example_configs/kafka-2_0_0.yml
```diff
@@ -0,0 +1,89 @@
...
+  type: GAUGE
+  labels:
+    topic: "$3"
+    partition: "$4"
+- pattern : kafka.server<type=(.+), name=(.+), clientId=(.+), brokerHost=(.+), brokerPort=(.+)><>Value
+  name: kafka_server_$1_$2
+  type: GAUGE
+  labels:
+    clientId: "$3"
+    broker: "$4:$5"
```
brian-brazil: Should metrics really be broken down by broker? That seems a bit over-specific to me. You care about how much you're communicating to the cluster overall, not where a partition lives.

ankon: Interesting question!

One thing I would see in favor for breaking things down by broker is environments where there are differences in latency: Assume you're spreading your kafka cluster over multiple data centers, then it is likely that some brokers are "further away" than others. Generally I assume that the approach with Prometheus is to expose "raw data", and have either Prometheus recording rules or the visualization side (Grafana/...) responsible for making sense of that raw data. Aggregating already here would hide that information.

With that said though: How else could this be aggregated on the exporter side?

brian-brazil: The issue is that such data could be too high cardinality to start with, per broker per client thread could add up pretty fast.

ankon:
> The other question would be, is there a way to get this data at the Java level that isn't JMX? This may be better done as a collector.

Also a good question to consider. The basic setup of Kafka exposes things via JMX, but: 
> Apache Kafka brokers and clients report many internal metrics. JMX is the default reporter, *though you can add any pluggable reporter.*

(from https://docs.confluent.io/current/kafka/monitoring.html, highlights by me)

Now, there are a couple of these listed in the Confluent.io confluence at https://cwiki.apache.org/confluence/display/KAFKA/JMX+Reporters, none for Prometheus. I tried finding a guide though on how to implement these, but that seems to not exist. For future googlers: Start reading at https://github.com/apache/kafka/blob/2.0.0/clients/src/main/java/org/apache/kafka/common/metrics/MetricsReporter.java :)

So, yes, one should be able to get the data without going through JMX.

The question I would pose back though is: Who would start doing that?
- People with huge clusters, specialized setups, ...: Yes, I can see that happening at one point
- People initially starting out with Kafka and just wanted to get some form of content into their Prometheus: Probably not.

I would consider the jmx_exporter example configs as that: An example to get you up and running, so that you have some monitoring rather than none -- but the moment your needs evolve you should also look back and review your implementation choices. This in a way also would be my answer to above discussion regarding broker-specific metrics: When that becomes a problem, you probably are in that state where JMX might no longer be the right answer -- but where you do have resources to actually think about the right answer.

The then bigger question becomes: Would it make sense to provide such examples, given that people tend to not review their implementation choices? :)

I would say "Yes", because right now the example for 0.8.2 actually gets things wrong to the point of being misleading (exposing the count of something as the value of it).

brian-brazil:
The thing with the jmx exporter is that it's for cases where you have no control over the source code, which is probably not the case when you're talking about kafka clients rather than kafka itself. A MetricsReporter would allow Prometheus users who are already instrumenting their application to pull in kafka metrics along with everything else in an efficient way. JMX should be considered a last resort, it's not exactly the best.

brian-brazil:
Ah wait, I'm confusing you with the other Kafka PR that's open at the moment.

example_configs/kafka-2_0_0.yml
brian-brazil: Okay so this is every broker times every kafka client - that's not at all sane cardinality. This set of metrics should be removed.

ankon: You got me a bit worried here, so I did a check to see what kind of metrics we're actually getting with these rules. IT turns out that this "client id" is not a "user client of kafka", but rather refers to some internal clients related to the replication system:

~~~~
$ echo beans | java -jar ~/Downloads/jmxterm-1.0.0-uber.jar  -l 30371 | grep -E 'ConsumerLag'
Welcome to JMX terminal. Type "help" for available commands.
$>beans
#domain = JMImplementation:
#domain = com.sun.management:
#domain = java.lang:
#domain = java.nio:
#domain = java.util.logging:
#domain = kafka:
#domain = kafka.cluster:
#domain = kafka.controller:
#domain = kafka.coordinator.group:
#domain = kafka.coordinator.transaction:
#domain = kafka.log:
#domain = kafka.network:
#domain = kafka.server:
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=0,topic=collaborne-events,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=0,topic=ml1-errors,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=0,topic=ml1-events,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=0,topic=ml3-commands,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=0,topic=ml3-events,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=0,topic=ml3-query-responses,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=0,topic=ml4-errors,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=1,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=10,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=13,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=16,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=19,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=22,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=25,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=28,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=31,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=34,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=37,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=4,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=40,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=43,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=46,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=49,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-1,name=ConsumerLag,partition=7,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=collaborne-commands,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=collaborne-errors,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=collaborne-query-responses,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=meta-topic,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=ml2-commands,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=ml2-errors,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=ml2-events,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=ml3-errors,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=0,topic=qa-query-responses,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=12,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=15,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=18,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=21,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=24,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=27,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=3,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=30,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=33,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=36,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=39,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=42,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=45,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=48,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=6,topic=__consumer_offsets,type=FetcherLagMetrics
kafka.server:clientId=ReplicaFetcherThread-0-2,name=ConsumerLag,partition=9,topic=__consumer_offsets,type=FetcherLagMetrics
#domain = kafka.utils:
~~~~

~~~~
$ echo beans | java -jar ~/Downloads/jmxterm-1.0.0-uber.jar  -l 30371 | grep -E 'broker'
Welcome to JMX terminal. Type "help" for available commands.
$>beans
#domain = JMImplementation:
#domain = com.sun.management:
#domain = java.lang:
#domain = java.nio:
#domain = java.util.logging:
#domain = kafka:
#domain = kafka.cluster:
#domain = kafka.controller:
#domain = kafka.coordinator.group:
#domain = kafka.coordinator.transaction:
#domain = kafka.log:
#domain = kafka.network:
#domain = kafka.server:
#domain = kafka.utils:
kafka.server:broker-id=1,fetcher-id=0,type=replica-fetcher-metrics
$>kafka.server:broker-id=2,fetcher-id=0,type=replica-fetcher-metrics
kafka.server:brokerHost=localhost,brokerPort=9093,clientId=ReplicaFetcherThread-0-1,name=BytesPerSec,type=FetcherStats
kafka.server:brokerHost=localhost,brokerPort=9093,clientId=ReplicaFetcherThread-0-1,name=RequestsPerSec,type=FetcherStats
kafka.server:brokerHost=localhost,brokerPort=9094,clientId=ReplicaFetcherThread-0-2,name=BytesPerSec,type=FetcherStats
kafka.server:brokerHost=localhost,brokerPort=9094,clientId=ReplicaFetcherThread-0-2,name=RequestsPerSec,type=FetcherStats
~~~~

These metrics only appear when a topic is actually having multiple replicas, the above output is from a 1-zookeeper/3-kafka local test setup, with 10 "real" clients connected to about 30 topic/partitions.

> Beware: ConsumerLag is an overloaded term in Kafka—it can have the above definition, but is also used to represent the offset differences between partition leaders and their followers. If you see kafka.server:type=FetcherLagMetrics,name=ConsumerLag,clientId=([-.\w]+),topic=([-.\w]+),partition=([0-9]+), it is in reference to the latter, not the former.

(https://github.com/DataDog/the-monitor/blob/master/kafka/monitoring-kafka-performance-metrics.md)

> `kafka.server:type=FetcherLagMetrics,name=ConsumerLag,clientId=([-.\w]+),topic=([-.\w]+),partition=([0-9]+)`
> Lag in number of messages per follower replica. This is useful to know if the replica is slow or has stopped replicating from the leader.

(https://docs.confluent.io/current/kafka/monitoring.html)

Based on that I think it would be good to keep these metrics: There is a very valid use case here to see whether the replication works, and https://kafka.apache.org/documentation/#rep-throttle ("Ensuring Progress") explicitly refers to using the `ConsumerLag` metric.

brian-brazil: Okay, still a potential cardinality issue but a reasonable thing to have as you can't monitor replication otherwise.

brian-brazil:
Thanks!

ankon:
Thank you very much for the review questions, these helped me a lot!

cricket007:
This is just Kafka Broker, right? 

Zookeeper, Kafka Streams (clients) and Kafka Connect are not included?

cricket007:
Thanks! Then, is the controller collected under the `kafka.(\w+)` ones?

ankon:
Yes, that was the idea. Note the list of metrics I added as an example in the initial post: https://github.com/prometheus/jmx_exporter/files/2257502/example-metrics.txt

PR #318, opened 2018-09-12
Update README: use shaded jar for standalone mode

spike01:
When starting in standalone mode, using the current instructions results in "Error: Could not find or load main class io.prometheus.jmx.JmxScraper"


spike01:
Interesting - this was when run against the jar at https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/0.3.1/jmx_prometheus_javaagent-0.3.1.jar

PR #330, opened 2018-10-25
Update the Apache Cassandra example configuration

mattrussi:
@brian-brazil 

Inspired by the kafka-2_0_0.yml configuration
* Latency histograms are exported as emulated Summaries.
* Otherwise, histograms exported as Summaries without the _sum.

List of example metrics: [cassandra-example-metrics.txt](https://github.com/prometheus/jmx_exporter/files/2516302/cassandra-example-metrics.txt)

There a ***lot*** of metrics exposed by default. In our implementation, we utilize several whitelistObjectNames values to only capture the specific metrics that we need. Otherwise, it presents a fairly heavy load on the exporter. I didn't want to restrict that outright. Instead, let the users determine which metrics they want to see.

### DIFF
--- a/example_configs/cassandra.yml
+++ b/example_configs/cassandra.yml
@@ -5,81 +5,54 @@ whitelistObjectNames: ["org.apache.cassandra.metrics:*"]
 blacklistObjectNames: ["org.apache.cassandra.metrics:type=ColumnFamily,*"]
 rules:
 # Generic gauges with 0-2 labels
-- pattern: org.apache.cassandra.metrics<type=(.+), (.+)=(.+), (.+)=(.+), name=(.+)><>Value
-  name: cassandra_$1_$6
+- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(\S*)><>Value
+  name: cassandra_$1_$5
   type: GAUGE
   labels:
+    "$1": "$4"
     "$2": "$3"
-    "$4": "$5"
-- pattern: org.apache.cassandra.metrics<type=(.+), (.+)=(.+), name=(.+)><>Value
-  name: cassandra_$1_$4
-  type: GAUGE
-  labels:
-    "$2": "$3"
-- pattern: org.apache.cassandra.metrics<type=(.+), name=(.+)><>Value
-  name: cassandra_$1_$2
-  type: GAUGE
 
 #
 # Emulate Prometheus 'Summary' metrics for the exported 'Histogram's.
 # TotalLatency is the sum of all latencies since server start
 #
-- pattern: org.apache.cassandra.metrics<type=(.+), (.+)=(.+), (.+)=(.+), name=(.+)?(?:Total)(Latency)><>Count
-  name: cassandra_$1_$6$7_sum
+- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(.+)?(?:Total)(Latency)><>Count
+  name: cassandra_$1_$5$6_seconds_sum
   type: COUNTER
   labels:
+    "$1": "$4"
     "$2": "$3"
-    "$4": "$5"
-
-- pattern: org.apache.cassandra.metrics<type=(.+), (.+)=(.+), (.+)=(.+), name=(.+)><>Count
-  name: cassandra_$1_$6_count
-  type: COUNTER
-  labels:
-    "$2": "$3"
-    "$4": "$5"
-- pattern: org.apache.cassandra.metrics<type=(.+), (.+)=(.+), (.+)=(.+), name=(.+)><>(\d+)thPercentile
-  name: cassandra_$1_$6_seconds
-  type: GAUGE
-  labels:
-    "$2": "$3"
-    "$4": "$5"
-    quantile: "0.$7"
   # Convert microseconds to seconds
   valueFactor: 0.000001
 
-- pattern: org.apache.cassandra.metrics<type=(.+), (.+)=(.+), name=(.+)?(?:Total)(Latency)><>Count
-  name: cassandra_$1_$4$5_sum
+- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=((?:.+)?(?:Latency))><>Count
+  name: cassandra_$1_$5_seconds_count
   type: COUNTER
   labels:
+    "$1": "$4"
     "$2": "$3"
 
-- pattern: org.apache.cassandra.metrics<type=(.+), (.+)=(.+), name=(.+)><>Count
-  name: cassandra_$1_$4_count
+- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(.+)><>Count
+  name: cassandra_$1_$5_count
   type: COUNTER
   labels:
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks. Is this losing us any metrics?

example_configs/cassandra.yml
```diff
@@ -1,14 +1,58 @@
...
+
+- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=((?:.+)?(?:Latency))><>Count
+  name: cassandra_$1_$5_seconds_count
+  type: COUNTER
+  labels:
+    "$1": "$4"
+    "$2": "$3"
+
+- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(.+)><>Count
+  name: cassandra_$1_$5_count
```
brian-brazil: This is picking up counters too, which would be _total usually.

mattrussi: I went back and forth on this. There doesn't appear to be a reliable way to determine whether the `<>Count` is part of a DropWizard Histogram/Timer or if it's an individual counter. 

The compromise I came to was choosing the naming scheme of a Summary which matches the majority of the metrics, and it still clearly denotes a counter (despite not matching best-practices for naming of counters with `_total`).
> the count of events that have been observed, exposed as `<basename>_count`

brian-brazil: Looking through the file, I think you can do that if it has histogram in the name it's a count otherwise total.

mattrussi: There are some other Histograms (Timers included) that don't have "histogram" in the name.

Would you prefer that they are defined each of these in the configuration for `_count`? 
We could then set all other remaining matches for `<>Count` to `<basename>_total`
Example:
```
- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(\S+Histogram|ViewReadTime|ViewLockAcquireTime|...)><>Count
  name: cassandra_$1_$5_count
  type: COUNTER
  labels:
    "$1": "$4"
    "$2": "$3"

- pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(.+)><>Count
  name: cassandra_$1_$5_total
  type: COUNTER
  labels:
    "$1": "$4"
    "$2": "$3"
```
Reference: http://cassandra.apache.org/doc/latest/operating/metrics.html

Name | Type | Description
 --- | --- | ---
AnticompactionTime | Timer | Time spent anticompacting before a consistent repair.
BytesValidated | Histogram | Histogram over the amount of bytes read during validation.
Hints_delays | Histogram | Histogram of hint delivery delays (in milliseconds)
Hints_delays-\<PeerIP\> | Histogram | Histogram of hint delivery delays (in milliseconds) per peer
IndexedEntrySize | Histogram | Histogram of the on-heap size, in bytes, of the index across all SSTables.
IndexInfoCount | Histogram | Histogram of the number of on-heap index entries managed across all SSTables.
IndexInfoGets | Histogram | Histogram of the number index seeks performed per SSTable.
PartitionsPerCounterBatch | Histogram | Distribution of the number of partitions processed per counter batch
PartitionsPerLoggedBatch | Histogram | Distribution of the number of partitions processed per logged batch
PartitionsPerUnloggedBatch | Histogram | Distribution of the number of partitions processed per unlogged batch
PartitionsValidated | Histogram | Histogram over the number of partitions read during validation.
RepairPrepareTime | Timer | Total time spent preparing for repair.
RepairTime | Timer | Total time spent as repair coordinator.
SyncTime | Timer | Time spent doing streaming during repair.
ValidationTime | Timer | Time spent doing validation compaction during repair.
ViewLockAcquireTime | Timer | Time taken acquiring a partition lock for materialized view updates on this table.
ViewReadTime | Timer | Time taken during the local read of a materialized view update.
WaitingOnCommit | Timer | The time spent waiting on CL fsync; for Periodic this is only occurs when the sync is lagging its sync interval.
WaitingOnFreeMemtableSpace | Histogram | Histogram of time spent waiting for free memtable space, either on- or off-heap.
WaitingOnSegmentAllocation | Timer | Time spent waiting for a CommitLogSegment to be allocated - under normal conditions this should be zero.

brian-brazil: It's probably easiest to do leave them all as _count, though several of the metrics in that list aren't in your output.

mattrussi: I've noticed that some of the metrics won't always show in the mBeans until they're called, unfortunately. My output was taken from a Cassandra 3.11 server in a development environment. There may be some functions which have not been called since Cassandra had been restarted.

There are some differences from the "latest" (4.0) docs, that I referenced above, compared to 3.11. That explains why some of them are not showing in my output. Sorry for the confusion.

3.11 docs: https://github.com/apache/cassandra/blob/cassandra-3.11/doc/source/operating/metrics.rst

brian-brazil: Actually thinking about this we should make these _total so that when the java client supports openmetrics it doesn't become _count_total as it's marked as a COUNTER.

mattrussi: Would that also apply for the `_sum` metrics as well? Or only for the `_count` metrics?
```
 - pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=((?:.+)?(?:Latency))><>Count
-  name: cassandra_$1_$5_seconds_count
+  name: cassandra_$1_$5_seconds_total
   type: COUNTER

 - pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(.+)><>Count
-  name: cassandra_$1_$5_count
+  name: cassandra_$1_$5_total
   type: COUNTER
```

brian-brazil: We could list them as UNTYPED

mattrussi: How's this?
```

Comment (brian-brazil) [reply example_configs/cassandra.yml]: If you're doing this, you can go back to _count.

Comment (mattrussi) [reply example_configs/cassandra.yml]: Let's go with UNTYPED so that we'll follow the Summary naming standards and since we cannot guarantee that the metrics under the `<>Count` catch-all is necessarily a Counter.

Does that work for you?
```diff
 - pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(.+)?(?:Total)(Latency)><>Count
   name: cassandra_$1_$5$6_seconds_sum
-  type: COUNTER
+  type: UNTYPED
   labels:
     "$1": "$4"
     "$2": "$3"

 - pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=((?:.+)?(?:Latency))><>Count
   name: cassandra_$1_$5_seconds_count
-  type: COUNTER
+  type: UNTYPED
   labels:
     "$1": "$4"
     "$2": "$3"

 - pattern: org.apache.cassandra.metrics<type=(\S*)(?:, ((?!scope)\S*)=(\S*))?(?:, scope=(\S*))?, name=(.+)><>Count
   name: cassandra_$1_$5_count
-  type: COUNTER
+  type: UNTYPED
   labels:
```

mattrussi: Thanks, @brian-brazil! Updated.

brian-brazil:
Thanks!

PR #335, opened 2018-11-12
safeName: handle numerica label case

### LINKED ISSUE #334: can pass on invalid label names
Example: `looker_cache_averagehitrate{001="Local Cache",name="General",} 0.0` the 001 violates label naming rules as it starts with a number.

Perhaps this exporter should automatically fix up these labels?

tmc:
Fixes #334


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -248,6 +248,9 @@ static String safeName(String name) {
       }
       boolean prevCharIsUnderscore = false;
       StringBuilder safeNameBuilder = new StringBuilder(name.length());
+      if (!name.isEmpty() && Character.isDigit(name.charAt(0))) {
+        safeNameBuilder.append("l");
```
brian-brazil: I don't think this is what you meant

tmc: @brian-brazil can you elaborate?

tmc: this puts a non-numeric 'l' in front of labels that start with a number. Should I do something longer/different/more-clear?

brian-brazil: I'd suggest putting an underscore in front

tmc: @brian-brazil done.

brian-brazil: Did you forget to push your change?

tmc:
@brian-brazil changed to underscore.

PR #358, opened 2019-03-04
fix objectName pattern for messaging queues

ndimitry:
Without this fix only exact "jboss.as:subsystem=messaging-activemq,server=*" objects were found, all having one more identifier (like jms-queue or jms-topic) were not found.
@brian-brazil please, take a look (I'm tagging you according to the process described in Contributing file for trivial fixes)

I've checked this fix on Java8 + WildFly13

### DIFF
--- a/example_configs/wildfly-10.yaml
+++ b/example_configs/wildfly-10.yaml
@@ -6,11 +6,13 @@ whitelistObjectNames:
  # see https://github.com/prometheus/jmx_exporter/issues/246#issuecomment-367573931
  # Each object in the rules below has to be added to whitelistObjectNames too !
  # note that rules use regex (like "foo.*", whereas the whitelist use globbing expressions (like "foo*")
- - "jboss.as:subsystem=messaging-activemq,server=*"
+ - "jboss.as:subsystem=messaging-activemq,server=*,jms-queue=*"
+ - "jboss.as:subsystem=messaging-activemq,server=*,jms-topic=*"
  - "jboss.as:subsystem=datasources,data-source=*,statistics=*"
  - "jboss.as:subsystem=datasources,xa-data-source=*,statistics=*"
  - "jboss.as:subsystem=transactions*"
  - "jboss.as:subsystem=undertow,server=*,http-listener=*"
+ - "jboss.as:subsystem=undertow,server=*,https-listener=*"
  # - "java.lang:*"
 rules:
   - pattern: "^jboss.as<subsystem=messaging-activemq, server=.+, jms-(queue|topic)=(.+)><>(.+):"
@@ -29,10 +31,9 @@ rules:
     attrNameSnakeCase: true
     name: wildfly_transaction_$1
 
-  - pattern: "^jboss.as<subsystem=undertow, server=(.+), http-listener=(.+)><>(bytes_.+|error_count|processing_time|request_count):"
+  - pattern: "^jboss.as<subsystem=undertow, server=(.+), (http[s]?-listener)=(.+)><>(bytes_.+|error_count|processing_time|request_count):"
     attrNameSnakeCase: true
-    name: wildfly_undertow_$3
+    name: wildfly_undertow_$4
     labels:
       server: $1
-      http_listener: $2
-
+      listener: $3


example_configs/wildfly-10.yaml
```diff
@@ -6,7 +6,8 @@ whitelistObjectNames:
  # see https://github.com/prometheus/jmx_exporter/issues/246#issuecomment-367573931
  # Each object in the rules below has to be added to whitelistObjectNames too !
  # note that rules use regex (like "foo.*", whereas the whitelist use globbing expressions (like "foo*")
- - "jboss.as:subsystem=messaging-activemq,server=*"
+ - "jboss.as:subsystem=messaging-activemq,server=*,jms-queue=*"
```
brian-brazil: `messaging-activemq,server=*,*` is what the docs indicate

ndimitry: Then I have a suggestion to replace also 2 strings "jboss.as:subsystem=datasources,data-source=*,statistics=*" and "jboss.as:subsystem=datasources,xa-data-source=*,statistics=*" with "jboss.as:subsystem=datasources,statistics=*,*" (it will also find all required datasources).

What do you think of this change? (If you approve I will commit changes an add them to this pull request)

brian-brazil: Assuming that doesn't pull in something else we don't want, that sounds good

ndimitry: I've checked, `jboss.as:subsystem=messaging-activemq,server=*,*` gets some more objects, not only the `jms-queue` and `jms-topic`
So it seems to me logical either (1) to keep both jms and datasources with `*` (your suggestion and my later suggestion). This will result in less amount of jmx calls (i.e. fewer `beanConn.queryMBeans` calls). but more obsolete objects will be received and thus more work with the pattern filters will occur. Or (2) to keep both jms and datasourses with more explicit  whitelist (which I suggested in the beginning). This will result in more jmx calls but since less MBeans will be received, less work will be done in filtering.

What seems to be more sufficient? In other words what is more "expensive": jmx call or (in my case) about 10 additional objects to filter? (being not an expert in this area I really don't know, I suspect they have pretty a similar "time price", hope you know more)

brian-brazil: Can you time both in your setup and see if there's a difference? If not, go with the simpler one.

ndimitry: Sorry, haven't perform additional checks... So let's have the simpler variant.

ndimitry: I will perform checks on WidlFly-15 (and some JBoss version) and then perform a merge here

brian-brazil: What was the outcome?

ndimitry: Found out that monitoring on our docker image was turned off...
Sorry, once I have any news I will write it.
P.S. I've also added a pattern for java.lang.* MBeans

brian-brazil:
No need for java.lang, that's already covered elsewhere.

example_configs/wildfly-10.yaml
ndimitry: Ok, this works for WildFly 15. Finally determined it (:
As for java.lang — in the white list there's a commented out line
` # - "java.lang:*"`
So I decided to add a commented line to the patterns
` # - pattern: "java.lang.*"`

brian-brazil: Can you remove the java.lang references? You also need to add a DCO

ndimitry: Ok, I'm removing pattern line which I added (if you want I may remove both java lines, what would you prefer?).

Sorry for having to spend so much time with me... This is the first time I face the DCO. After googling I found that by this I state that changed are made by me and I have right to commit them (something like this).
Does need to add DCO means adding a comment to the git commit a line alike:
Signed-off-by: Joe Smith <joe.smith@email.com>

ndimitry: Checked one of old commits, found similar line (Signed-off-by ...) as a DCO. I will add it on the last commit after a decision to remove both java.lang mentions or only one of them (which was added my be, the `pattern` one).

brian-brazil:
You need to add it to all commits, squashing is generally easiest.

example_configs/wildfly-10.yaml
ndimitry: After some investigations have found some more interesting details to commit. For example http-listener helps to track http communication, https-listener also helps to check https traffic. In my case this was a really important thing.

brian-brazil:
Are you still working on this?

ndimitry:
I've made changes, but need some time to commit it correctly.
Finally I haven't made any performance testing... So maybe it is worth to use the simple (i.e. shortest) way

brian-brazil:
I believe I just need you to add the DCO so that I can merge.

ndimitry:
Let me do this today. Sorry that left it without any action for such long time.

ndimitry:
P.S. I have one more change for the https listener. Will add it.

ndimitry:
FYI I've checked 2 configurations with jmx http server on wildfly 15. The longer config seems to work faster (scrapes were about 0.3 comparing to 1 sec), so I'm adding the longer varient

ndimitry:
Done

example_configs/wildfly-10.yaml
```diff
@@ -29,10 +31,11 @@ rules:
...
-  - pattern: "^jboss.as<subsystem=undertow, server=(.+), http-listener=(.+)><>(bytes_.+|error_count|processing_time|request_count):"
+  - pattern: "^jboss.as<subsystem=undertow, server=(.+), (http[s]?-listener)=(.+)><>(bytes_.+|error_count|processing_time|request_count):"
     attrNameSnakeCase: true
-    name: wildfly_undertow_$3
+    name: wildfly_undertow_$4
     labels:
       server: $1
-      http_listener: $2
-
+      $2: $3
```
brian-brazil: The label names shouldn't vary here.

ndimitry: According to the pattern there are only 2 choices for this label name: either `http_listener` or `https_listener.`
Bit if you prefer 2 separate pattern with fixed label names I will separate them.

brian-brazil: Why do you want two different label names here?

ndimitry: I might be wrong but it looks for me as it is possible to have both http_listener and https_listener. If this is so we will have a clash with the same metric name and labels names.

ndimitry: On my installation there was no http_listener at all, there was only an https_listener one

brian-brazil: That's something to distinguish with label values rather than names, and I'd presume both can't be on the same port.

ndimitry: That's true, same names are not possible.
I've checked Wildfly config, the undertow system can contain 3 types of listeners: ajp listeners, http listeners and https listeners. That's why I see it as 3 separated (though living together) things. In this config only 2 of them are queried for metrics. That was a reason I added 2 different labels (via the pattern).
If you'd still prefer only `http_listener` as label, lets make it so. I've just shared my understanding :)

ndimitry: Please, make final desision and I'll commit last changes

brian-brazil: Oh sorry, missed your response. It sounds like `listener` would be most correct here.

example_configs/wildfly-10.yaml
```diff
@@ -29,10 +31,11 @@ rules:
...
+      
+  # - pattern: "java.lang.*"
```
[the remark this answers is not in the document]
ndimitry: done

example_configs/wildfly-10.yaml
ndimitry: changed `$2: $3` to `listener: $3`
So now all seems to be fine (:

brian-brazil:
Thanks!

ndimitry:
You are welcome!
And thanks for patience!

PR #363, opened 2019-03-06
Update JmxScraper.java

AviNachman:
add null check for attributes of bean to avoid NPE


collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -149,6 +149,9 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) {
         final AttributeList attributes;
         try {
             attributes = beanConn.getAttributes(mbeanName, name2AttrInfo.keySet().toArray(new String[0]));
+            if(attributes == null){
```
brian-brazil: Can you add a debug  log line when this happens, and fix the formatting?

brian-brazil:
This is obsolete.

Issue #312, opened 2018-09-06
Argument parsing fails on Windows with absolute configuration file path


enyone:
Reason is `string.split(":")` is used blindly when parsing arguments.

```
java.exe "-javaagent:./jmx_prometheus_javaagent-0.3.1.jar=8080:C:\config.yaml" -jar "C:\yourJar.jar"
```

brian-brazil:
Currently you can workaround this using relative paths

mvdobrinin:
You can also work around this by using absolute paths from the current drive. 

```\Program Files\Custom Utilities\StringFinder.exe```

See the MS docs - https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats.

mvdobrinin:
@brian-brazil Let me know your thoughts. 😄

mvdobrinin:
#371

PR #371, opened 2019-04-01
Fix handling of Windows paths in -javaagent config arguments (#312)

mvdobrinin:
This fixes Windows path handling.


mvdobrinin:
@brian-brazil

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -20,37 +22,71 @@ public static void premain(String agentArgument, Instrumentation instrumentation
...
+            this.host = host;
+            this.port = port;
+            this.file = file;
+            this.socket = socket;
+        }
+    }
+
+    private static Pattern pattern = Pattern.compile(
+            "^(([\\w.]+):)?" +  // optional host portion (not ipv6)
+            "(\\d{1,5})" +      // port
```
brian-brazil: Ports can be letters.

I'd suggest counting the colons, and if there's > 2, parsing them from left to right and what's left is the filename.

mvdobrinin: Hmm? Can you clarify when a port can be letters? I'm not familiar with this concept.

brian-brazil: `http` is the same as 80, per /etc/services.

mvdobrinin: Ah yes, I see. However I'm not sure that's possible in this code base since we are in Java. Java does not have an API to find ports by name. All the APIs that are used in this code base require a port to be an integer. For example all constructor calls to `InetSocketAddress`. Maybe that part can be added as a new feature down the road?

brian-brazil: So it is.

This code still isn't quite right. Someone can do [ipv6]:host:1234:path

jmx_prometheus_javaagent/src/test/java/io/prometheus/jmx/JavaAgentIT.java
```diff
@@ -112,4 +113,66 @@ private String resolveRelativePathToResource(String resource) {
...
+    }
+
+    /**
+     * Exercise the existing Ipv6 parsing logic. The current logic leaves the brackets on the host.
+     */
+    @Test
+    public void testIpv6AddressParsing() {
+        final String DEFAULT_HOST = "0.0.0.0";
+
+        JavaAgent.Config config = JavaAgent.parseConfig("[1:2:3:4]:8080:config.yaml", DEFAULT_HOST);
```
mvdobrinin: See here.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
mvdobrinin: If you are talking about ipv6 addresses, I believe that functionality is captured. I left a Github comment on the unit test where this is demonstrated. That functionality was introduced by a previous pull request, and I left that logic as-is. That logic remains at the top of the `parseConfig` method. I tried to make these changes non-invasive.

brian-brazil: I'm talking specifically about the invalid input `[ipv6]:host:1234:path`, which should be rejected.

> I tried to make these changes non-invasive.

It's more important that it works, and the end code is clean.

mvdobrinin: Having `[ipv6]:host:1234:path` as accepted input was an existing issue in the codebase, but I did go ahead and fix it 😁. I also cleaned up the solution to use a single regex, instead of having separate logic for the ipv6 address.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -3,54 +3,88 @@
...
+
+        Config(String host, int port, String file, InetSocketAddress socket) {
+            this.host = host;
+            this.port = port;
+            this.file = file;
+            this.socket = socket;
+        }
+    }
+
+    private static Pattern pattern = Pattern.compile(
```
brian-brazil: I'd put this in the function for clarity. Efficiency doesn't matter here.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -3,54 +3,88 @@
...
+            "^(?:((?:[\\w.]+)|(?:\\[.+])):)" +  // host name, or ipv4, or ipv6 address in brackets
+            "?(\\d{1,5})" +                     // port
```
brian-brazil: Question mark should be on the line above

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -3,54 +3,88 @@
...
+        Config(String host, int port, String file, InetSocketAddress socket) {
+            this.host = host;
+            this.port = port;
+            this.file = file;
+            this.socket = socket;
+        }
+    }
+
+    private static Pattern pattern = Pattern.compile(
+            "^(?:((?:[\\w.]+)|(?:\\[.+])):)" +  // host name, or ipv4, or ipv6 address in brackets
```
brian-brazil: Escape the `]` too

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -3,54 +3,88 @@
...
+
+    static class Config {
+        String host;
+        int port;
+        String file;
+        InetSocketAddress socket;
+
+        Config(String host, int port, String file, InetSocketAddress socket) {
+            this.host = host;
+            this.port = port;
```
brian-brazil: Port and host aren't used.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
mvdobrinin: This is considered a redundant escape in Java.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
mvdobrinin: Good point.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
mvdobrinin: These are used for the unit tests.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
mvdobrinin: Done.

brian-brazil:
Hmm, this change is making `mvn verify` fail.

mvdobrinin:
It is passing for me locally. Can you share the error?

brian-brazil:
https://travis-ci.org/prometheus/jmx_exporter/builds/526844396

PR #381, opened 2019-04-23
Add example of Apache Ignite config

char16t:


example_configs/ignite.yml
```diff
@@ -0,0 +1,37 @@
+---
+# This is example of getting Apache Ignite cache metrics.
+#
+# This metrics will be available after enabling JMX metrics gathering
+# Instructions available on https://apacheignite.readme.io/v2.0/docs/memory-and-cache-metrics
+# or by search query "ignite jmx metrics"
+#
+# 1. Enable statistic for your cache (example for Scala)
```
brian-brazil: If you have access to change the code, then using client_java directly is preferred to using JMX.

char16t: Complex question. Metrics registration needs to be explicitly call. Some metrics are not directly dependent on application actions. For Example, "EntryProcessorHits", "KeysToRebalanceLeft". I think that it is undesirable to register these metrics yourself

char16t: Part of available Ignite metrics: screenshot-1

brian-brazil: The code should be registering the metrics, and storing them as static class methods.

PR #382, opened 2019-04-30
Stricter safename

eliasson:
Only characters `a-zA-Z0-9_:` is allowed for names and labels.

The current use of `Character.isLetterOrDigit` supports Unicode characters (within the Basic Multilingual Plane) which lets names using invalid characters through (e.g. use of Swedish letters åäö).

I find this change trivial one, so according to `CONTRIBUTING.md` I went straight ahead and made a pull request. @brian-brazil please review.


brian-brazil:
Can you add the DCO please?

This bug might be lurking in other repos too.

eliasson:
DCO added

brian-brazil:
Thanks!

PR #383, opened 2019-04-30
Create Docker image

srueg:
Set up a Dockerfile and entrypoint script to build a Docker image.

Signed-off-by: Simon Rüegg <simon.ruegg@vshn.ch>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git/
+Dockerfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM docker.io/library/maven:3.6-jdk-11-slim
+
+WORKDIR /build
+
+COPY ./ /build
+
+RUN mvn clean package
+
+WORKDIR /build/jmx_prometheus_httpserver
+RUN mvn clean package
+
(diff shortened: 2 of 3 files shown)


brian-brazil:
Thanks for your PR. It does not make sense to run this in Docker, as it's intended to be used as a java agent.

srueg:
I am actually running it in Kubernetes as a side car, using the http server mode.

brian-brazil:
I'd suggest using the java agent then.

srueg:
Unfortunately that's not possible as I don't control the other image and it doesn't contain the agent.

srueg:
I might have found a possible solution though: https://github.com/banzaicloud/prometheus-jmx-exporter-operator

srueg:
Unfortunately the JMX operator doesn't work as expected and it makes some assumptions about the targeted environment we can't fulfill.

So I have a valid use case to run the JMX exporter in a Docker image. There are also several other images which package it, but they're all outdated and I think it would make most sense to have the Dockerfile in this project. How do you feel about this?

brian-brazil:
I continue to believe that the agent is the best way to do things.

PR #388, opened 2019-06-12
Add jmxremote_optional dependency into collector to support to scrap remote JMX server metrics

imasahiro:
Motivation:
- When using <= JDK 8, we can scrap the JMX metrics via JMXMP protocol
  with installing jmxremote_optional.jar to JRE lib/ext dir (that is
  used for JVM extension mechanism).
- But for JDK 10+, this JVM extension mechanism is removed. So we cannot
  support JMXMP protocol anymore

Changes:
- Add jmxremote_optional dependency to collector module

Result:
- Can support JMXMP protocol again

### DIFF
--- a/collector/pom.xml
+++ b/collector/pom.xml
@@ -31,6 +31,13 @@
       <artifactId>snakeyaml</artifactId>
       <version>1.16</version>
     </dependency>
+    <!-- Since javax.management:jmxremote_optional:1.0.1_04 jar file does not
+         exist on maven central, use substitute one -->
+    <dependency>
+        <groupId>org.glassfish.external</groupId>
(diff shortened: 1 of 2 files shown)


brian-brazil:
This isn't something that we currently support. What are the implications?

imasahiro:
Oh, I didn't notice that running this exporter as a java agent is recommended way these days until recieve your comment. (sorry, I'm now reading latest README)

When I started to write this PR, I thought it is needed to add this but after some try, I realized that if I run jmx_prometheus_httpserver with `-Xbootclasspath/a:/path/to/jmxremote_optional.jar`, it works well. So it seems I jumped to conclusions and there is no point to handle this PR for now.

PR #395, opened 2019-07-02
Add kafka.coordinator.* metrics

### LINKED ISSUE #393: Fix kafka-2_0_0.yml to include kafka.coordinator metrics
Current sample is pretty comprehensive but misses out on `kafka.coordinator.group` and `kafka.coordinator.transaction` metrics.

Walliee:
Resolves #393

### DIFF
--- a/example_configs/kafka-2_0_0.yml
+++ b/example_configs/kafka-2_0_0.yml
@@ -15,6 +15,9 @@ rules:
   labels:
     clientId: "$3"
     broker: "$4:$5"
+- pattern : kafka.coordinator.(\w+)<type=(.+), name=(.+)><>Value
+  name: kafka_coordinator_$1_$2_$3
+  type: GAUGE
 
 # Generic per-second counters with 0-2 key/value pairs
 - pattern: kafka.(\w+)<type=(.+), name=(.+)PerSec\w*, (.+)=(.+), (.+)=(.+)><>Count


brian-brazil:
What do the new metrics look like?

Walliee:
I was unable to find any documentation for these (which is why they were probably missed in the first place) but here they are from the source:

**kafka.coordinator.transaction**
https://github.com/apache/kafka/blob/050fdd6537fcfb640277ff5e1607e44d18c0d95f/core/src/main/scala/kafka/coordinator/transaction/TransactionMarkerChannelManager.scala#L146-L158

**kafka.coordinator.group**
https://github.com/apache/kafka/blob/6dd4ebcea78f9c6fb8a2921cab3fd92be4978063/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala#L92-L137

brian-brazil:
I meant more what does the jmx exporter expose for these now with this config?

Walliee:
Using the current version of the config, kafka.coordinator.* are ignored as they are not captured by even the most generic pattern:
```
- pattern: kafka.(\w+)<type=(.+), name=(.+)><>Value
  name: kafka_$1_$2_$3
  type: GAUGE
```

With the new addition to the config the following gauges will be exposed:

- kafka_coordinator_group_groupmetadatamanager_numgroups
- kafka_coordinator_group_groupmetadatamanager_numgroupscompletingrebalance
- kafka_coordinator_group_groupmetadatamanager_numgroupsdead
- kafka_coordinator_group_groupmetadatamanager_numgroupsempty
- kafka_coordinator_group_groupmetadatamanager_numgroupspreparingrebalance
- kafka_coordinator_group_groupmetadatamanager_numgroupsstable
- kafka_coordinator_transaction_transactionmarkerchannelmanager_logappendretryqueuesize
- kafka_coordinator_transaction_transactionmarkerchannelmanager_unknowndestinationqueuesize

brian-brazil:
Thanks!

PR #399, opened 2019-07-08
include error message with exception

rustyrazorblade:
When the agent starts up, you get the same output no matter what is wrong:

```
Usage: -javaagent:/path/to/JavaAgent.jar=[host:]<port>:<yaml configuration file>```
```

This patch adds the exception to the output:

```
Usage: -javaagent:/path/to/JavaAgent.jar=[host:]<port>:<yaml configuration file> At most one of hostPort and jmxUrl must be provided
```

cc @brian-brazil,


brian-brazil:
Can you rebase against master, and add the DCO?

rustyrazorblade:
You got it.

brian-brazil:
Looks like Travis has broken jdk8 installs, so I'll just merge this.

Issue #400, opened 2019-07-17
Request to change module version in pom.xml


nakyo7:
There is an item I want to see, but it isn't there. (e.g, jvm_threads_state)
The simpleclient_hotspot version is too low to see. 

https://github.com/prometheus/jmx_exporter/blob/6f9f74ce892374bc917377f8bb00016244c4e7e9/jmx_prometheus_javaagent/pom.xml#L26

PR #401, opened 2019-07-18
Update simpleclient_hotspot version (#400)

nakyo7:
Hi @brian-brazil 

There is an item I want to see, but it isn't there. (e.g, jvm_threads_state)
I found the simpleclient_hotspot version is too low to see in javaagent.


brian-brazil:
Can you add the DCO please?

nakyo7:
@brian-brazil  Ok. I did it.

brian-brazil:
Thanks!

Issue #405, opened 2019-07-31
Agent Test has bad check for empty javaHome


swpalmer:
It looks like this line:
https://github.com/prometheus/jmx_exporter/blob/cae9f262557f4460a3de920fa9542f8ed75f5cbe/jmx_prometheus_javaagent/src/test/java/io/prometheus/jmx/JavaAgentIT.java#L62

Should be:

`if (javaHome != null && !javaHome.isEmpty()) {`

brian-brazil:
Would you like to send a PR?

bristy:
@brian-brazil  Do you want me to send PR for this?

brian-brazil:
Please.

PR #406, opened 2019-08-01
Fixed empty check of JAVA_HOME (#405)

swpalmer:
@brian-brazil Trivial fix for testing when to use JAVA_HOME in a test


brian-brazil:
Can you add the DCO please?

brian-brazil:
This has gone stale, let me know if you're still working on it.

PR #425, opened 2019-08-29
Update example Kafka config to omit per-partition metrics

nickbp:
Per the comment in the change, these metrics scale with Ntopics * Npartitions. In a non-trivial cluster this will lead to payloads getting dropped because they take so long to generate/collect. Therefore it makes sense to omit them as a default unless they're specifically needed.

Tested against Kafka 2.1.

@brian-brazil


brian-brazil:
This is true, however they're also the most interesting metrics generally so should be there by default.

PR #430, opened 2019-09-10
Allow collector configuration from InputStream

mindcrime-ilab:
@brian-brazil adding support for reading the collector's configuration from an ``InputStream``. This is handy if you want to read the config from an archive in the classpath, eg. using Spring's ``ResourceLoader``:

```java
try(InputStream is = loader.getResource("classpath:/collect.yml").getInputStream()) {
  collectorRegistry.register(new JmxCollector(is));
}
```


brian-brazil:
Does String not cover this?

mindcrime-ilab:
I feel it is a bit more convenient and make the API easier to use than reading it into a String to be read out again. The YAML parser itself allows parsing a stream directly too, so it would be even closer to it.

brian-brazil:
Thanks!

PR #440, opened 2019-10-16
Allow metrics server to be configured with TLS.

senthilkumarkj:
Add a new server config file. Accept keystore path and password.


senthilkumarkj:
Created by mistake. Please ignore.

PR #450, opened 2019-11-26
Possibility to add IPv6 as a parameter for http server

ndimitry:
Small change that allows to add IPv6 hosts


brian-brazil:
Do we need any changes on the java agent side?

ndimitry:
No, no changes are required in the agent.
JavaAgent class contains method parseConfig which parses data in an even more sophisticated and enhanced way (but there 3 parameters are possible [host:]<port>:<config> and regular expressions are used for parsing)

brian-brazil:
Thanks!

ndimitry:
Welcome (: not that big change.
And I made a typo in my previous comment, there should be: "3 parameters are possible [host:]\<port\>:\<config\>" (didn't know I should escape \< and \>, now I do)

PR #451, opened 2019-12-02
Update example kafka config to pull in MeanRate, OneMinuteRate, FiveM…

mitchell-h:
…inuteRate, FifteenMinuteRate

Pulling in the various Mean/One/Five/fifteen minute rates is especially necessary when debugging resource utilization issues with kafka.  For example kafka.server:type=KafkaRequestHandlerPool,name=RequestHandlerAvgIdlePercent

### DIFF
--- a/example_configs/kafka-2_0_0.yml
+++ b/example_configs/kafka-2_0_0.yml
@@ -96,6 +96,18 @@ rules:
 - pattern: kafka.(\w+)<type=(.+), name=(.+)><>Count
   name: kafka_$1_$2_$3_count
   type: COUNTER
+- pattern: kafka.(\w+)<type=(.+), name=(.+)><>MeanRate
+  name: kafka_$1_$2_$3_MeanRate
+  type: COUNTER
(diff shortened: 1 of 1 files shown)


brian-brazil:
This is on purpose, see https://prometheus.io/docs/instrumenting/writing_exporters/#drop-less-useful-statistics

mitchell-h:
Thanks for the comment @brian-brazil .   The reason I submitted this for the kafka exporter alone is because many of the count metrics aren't what is need to support kafka.  for example take a look at the count value for kafka.server:type=KafkaRequestHandlerPool,name=RequestHandlerAvgIdlePercent it is not what's needed to monitor that metric.  the COUNT value is just the number of submitted metrics. 

If I refine the change down to only the most relevant metrics would that be better?

brian-brazil:
You never need these with Prometheus, rate(Count) provides the same.

brian-brazil:
Closing, per above these are explicitly excluded.

PR #457, opened 2020-01-06
Fix config for zookeeper standalone

rafitadiaz:
The metrics from the JMX exporter have this format: 
```
org.apache.ZooKeeperService<name0=StandaloneServer_port-1><>MinRequestLatency)
```
So the regexp needs a `-` after `port`.

### DIFF
--- a/example_configs/zookeeper.yaml
+++ b/example_configs/zookeeper.yaml
@@ -17,7 +17,7 @@ rules:
       replicaId: "$2"
       memberType: "$3"
   # standalone Zookeeper
-  - pattern: "org.apache.ZooKeeperService<name0=StandaloneServer_port(\\d+)><>(\\w+)"
+  - pattern: "org.apache.ZooKeeperService<name0=StandaloneServer_port-(\\d+)><>(\\w+)"
     name: "zookeeper_$2"
-  - pattern: "org.apache.ZooKeeperService<name0=StandaloneServer_port(\\d+), name1=InMemoryDataTree><>(\\w+)"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Hmm, did this change at some point? It'd seem odd that you'd be the first person to notice such an obvious breakage.

rafitadiaz:
Not sure, I just did this setup in my staging like 3 weeks ago and I found this. 
In my case the versions that I'm using are:

    zookeeper-3.4.6-1.el7.x86_64
    jmx_prometheus_javaagent-0.12.0.jar

And debugging the Mbeans with jconsole I found a tree like this
```
> org.apache.ZooKeeperService
  > StandaloneServer_port-1
    > ....
```

I'm not familiar with Mbeans or Java processes so maybe it's a misconfiguration in the zookeeper server for Centos7 (I'm using the RPM package with the default conf)

brian-brazil:
No, this shouldn't be affected by configuration. Hmm could it be that the port in this case is -1, that is negative?

rafitadiaz:
Oh, I understood that the port was `1`... like `port-(\\d)`
I don't think that I've never seen a negative port before.  :thinking: and it sounds like a really weird default value but it could be.

If for another people is working without the `-` we also can go for a more generic implementation for the regexp with:

```regexp
    org.apache.ZooKeeperService<name0=StandaloneServer_port-*(\\d*)><>AvgRequestLatency)
```
That should work for metrics with and without `-`

brian-brazil:
If it's part of the number we should capture it. We need to know where this is coming from though, as this is not what is documented.

PR #459, opened 2020-01-24
hbase.yaml

tgunkel:
For the HBase database I needed a config file. Found one here https://blog.godatadriven.com/hbase-prometheus-monitoring but it could not handle _ in table names.

### DIFF
--- a/example_configs/hbase.yaml
+++ b/example_configs/hbase.yaml
@@ -1,0 +1,13 @@
+# The metrics look this this
+# Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_foo_table_my_table_name_region_1234567xy_metric_scanNext_75th_percentile: 0.0
+# So we have many many different metrics and it is hard to aggregate per namespace or per table
+# Idea is to have it under one metric and the details in the labels
+# Idea came from https://blog.godatadriven.com/hbase-prometheus-monitoring
+rules:
+  - pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_(.+)_table_(.+)_region_(.+)_metric_(.+):\ (.+)
+    name: HBase_metric_$4
+    labels:
+      namespace: "$1"
+      table: "$2"
+      region: "$3"
+    value: "$5"


example_configs/hbase.yaml
```diff
@@ -0,0 +1,13 @@
...
+# Idea is to have it under one metric and the details in the labels
+# Idea came from https://blog.godatadriven.com/hbase-prometheus-monitoring
+rules:
+  - pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_(.+)_table_(.+)_region_(.+)_metric_(.+):\ (.+)
+    name: HBase_metric_$4
+    labels:
+      namespace: "$1"
+      table: "$2"
+      region: "$3"
+    value: "$5"
```
brian-brazil: Why are you not leaving this to the default?

example_configs/hbase.yaml
```diff
@@ -0,0 +1,13 @@
+# The metrics look this this
+# Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_foo_table_my_table_name_region_1234567xy_metric_scanNext_75th_percentile: 0.0
+# So we have many many different metrics and it is hard to aggregate per namespace or per table
+# Idea is to have it under one metric and the details in the labels
+# Idea came from https://blog.godatadriven.com/hbase-prometheus-monitoring
+rules:
+  - pattern: Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_(.+)_table_(.+)_region_(.+)_metric_(.+):\ (.+)
+    name: HBase_metric_$4
```
brian-brazil: metric is redundant here

we also go with lowercase, so hbase

example_configs/hbase.yaml
```diff
@@ -0,0 +1,13 @@
+# The metrics look this this
+# Hadoop<service=HBase, name=RegionServer, sub=Regions><>Namespace_foo_table_my_table_name_region_1234567xy_metric_scanNext_75th_percentile: 0.0
+# So we have many many different metrics and it is hard to aggregate per namespace or per table
```
brian-brazil: No need for this comment, this is the usual way to do things.

example_configs/hbase.yaml
nvtkaszpir: It may be added so that it does not overwrite other metrics which may overlap or have different type.

for example in HBase you can have already generated metric named `Hbase_active`: 1.0

Then there is some hbase data with below structure (abbreviated)
...`Namespace_test_table_meta_region_abc_metric_get`: 20
and which would produce `Hbase_active` with labels namespace=test,table=meta,region=abc with value 20.0

In that case it would cause metric clash.

I would keep lowercase, as suggested.

dhoard:
Closed due to inactivity.

PR #463, opened 2020-02-08
Add all avg metrics of kafka broker

alvarobacelar:
Hi @brian-brazil. In this PR I added a regex for to get the avg type metrics (Mean).

The regexp bellow get only metrics of the type thPercentile.
```
- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.*)><>(\d+)thPercentile
  name: kafka_$1_$2_$3
  type: GAUGE
  labels:
    "$4": "$5"
    quantile: "0.$6"
```

But for the Apache Kafka is very import know the metrics of the type avg as well. 

Now with the metrics of this PR it is possible filter the quantile with 0.50, 0.75, 0.99, 0.999 or avg.

### DIFF
--- a/example_configs/kafka-2_0_0.yml
+++ b/example_configs/kafka-2_0_0.yml
@@ -87,6 +87,12 @@ rules:
   type: COUNTER
   labels:
     "$4": "$5"
+- pattern: kafka.(\w+)<type=(.+), name=(.+), (.+)=(.+)><>Mean
+  name: kafka_$1_$2_$3
+  type: GAUGE
+  labels:
+    "$4": "$5"
+    quantile: "avg"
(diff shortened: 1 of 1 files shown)


brian-brazil:
These are explicitly excluded, as they're not useful. You can calculate latency via PromQL.

alvarobacelar:
@brian-brazil , but some cases I want filter a request total time per 99th, 999th or avg, on the Grafana for exemple. So I create variables with quantilites and can make available in Grafana. 
So it gets easier more than Using PromQL.

brian-brazil:
This sounds more like a usage query, the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) is a better place for that.

Issue #469, opened 2020-02-28
Support custom JMX connector properties


bmesserer:
We're trying to connect to WebLogic via remote t3s protocol, but unfortunately, like quite a few other application servers, its remote "JMX over RMI" implementation uses a proprietary protocol called "t3" (or t3s over TLS) - see e.g. https://docs.oracle.com/cd/E13222_01/wls/docs90/jmx/accessWLS.html.
While it also supports IIOP, and newer versions have partially rectified this situation, many enterprises are still running lots of these app servers and enabling the possibility to scrape them is actually pretty simple - just allow to configure a few custom properties for the JMX environment.
These might also come in handy for more detailed customization as vendors (incl. Sun) use these for all configuration of the JMX connector.

brian-brazil:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

bmesserer:
> It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

This isn't a question. The JMX exporter currently does not support this because it doesn't support setting `jmx.remote.protocol.provider.pkgs` (and other necessary properties).
The prometheus-users mailing list [doesn't help](https://groups.google.com/forum/#!searchin/prometheus-users/jmx|sort:date/prometheus-users/nrEH0FLis3c/PL8TgA2nFwAJ) as it's not concerned with JMX/RMI protocol issues.

sgoeschl:
@brian-brazil I actually asked on the user mailing list some time ago to no avail - please see

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/prometheus-users/nrEH0FLis3c

PR #470, opened 2020-02-28
Added support for custom JMX connector properties (#469)

bmesserer:
Adds ability to configure custom JMX connector properties in YAML config (e.g. for more involved remote connections), fixes issue #469 (full disclosure: created by me).

@brian-brazil 
@VladimirYushkevich

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ username:
 password: 
 jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:1234/jmxrmi
 ssl: false
+jmxEnvironment:
+  "some.useful.property": "any string value"
+  "another.jmx.connector.property": "customValue"
 lowercaseOutputName: false
 lowercaseOutputLabelNames: false
 whitelistObjectNames: ["org.apache.cassandra.metrics:*"]
@@ -58,10 +61,11 @@ Name     | Description
 ---------|------------
 startDelaySeconds | start delay before serving requests. Any requests within the delay period will result in an empty metrics set.
 hostPort | The host and port to connect to via remote JMX. If neither this nor jmxUrl is specified, will talk to the local JVM.
-username | The username to be used in remote JMX password authentication.
-password | The password to be used in remote JMX password authentication.
+username | The username to be used in remote JMX password authentication (used for `jmx.remote.credentials` JMX-connector property).
+password | The password to be used in remote JMX password authentication (used for `jmx.remote.credentials` JMX-connector property).
 jmxUrl   | A full JMX URL to connect to. Should not be specified if hostPort is.
-ssl      | Whether JMX connection should be done over SSL. To configure certificates you have to set following system properties:<br/>`-Djavax.net.ssl.keyStore=/home/user/.keystore`<br/>`-Djavax.net.ssl.keyStorePassword=changeit`<br/>`-Djavax.net.ssl.trustStore=/home/user/.truststore`<br/>`-Djavax.net.ssl.trustStorePassword=changeit`
+ssl      | Whether JMX connection should be done over SSL. Note this only sets RMI/JRMP standard properties, other RMI implementations (e.g. WebLogic, JBoss, IIOP) may not be affected! To configure certificates you have to set following system properties:<br/>`-Djavax.net.ssl.keyStore=/home/user/.keystore`<br/>`-Djavax.net.ssl.keyStorePassword=changeit`<br/>`-Djavax.net.ssl.trustStore=/home/user/.truststore`<br/>`-Djavax.net.ssl.trustStorePassword=changeit`
+jmxEnvironment | Properties passed directly to JMX-connector. Can be used to configure proprietary RMI protocol implementations and more nuanced properties. Typically not needed for "standard" use cases. 
 lowercaseOutputName | Lowercase the output metric name. Applies to default format and `name`. Defaults to false.
 lowercaseOutputLabelNames | Lowercase the output metric label names. Applies to default format and `labels`. Defaults to false.
 whitelistObjectNames | A list of [ObjectNames](http://docs.oracle.com/javase/6/docs/api/javax/management/ObjectName.html) to query. Defaults to all mBeans.
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.Collections;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -56,6 +57,7 @@ public class JmxCollector extends Collector implements Collector.Describable {
       String username = "";
       String password = "";
       boolean ssl = false;
+      Map<String, String> jmxEnvironment=Collections.emptyMap();
       boolean lowercaseOutputName;
       boolean lowercaseOutputLabelNames;
       List<ObjectName> whitelistObjectNames = new ArrayList<ObjectName>();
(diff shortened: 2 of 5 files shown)


brian-brazil:
I don't see how this PR is tied to this issue, you can already set the full URI.

In general you should use the java agent, which avoids all this.

bmesserer:
It allows to set custom JMX properties. 
Setting the full URL is unfortunately not enough (see `weblogic-t3s.yml`), because scraping will just result in "unsupported protocol", because T3/T3s requires `jmx.remote.protocol.provider.pkgs` to be set to `weblogic.management.remote` (not to mention the separate authentication via `java.naming.security.`[`principal`|`credentials`] vs. the currently used `jmx.remote.credentials`) - see https://docs.oracle.com/cd/E13222_01/wls/docs90/jmx/accessWLS.html .
The only standard supported protocol is JRMP (since IIOP has been removed from the JDK; the default is to take sub-packages of `com.sun.remote.protocol` - unless one sets `jmx.remote.protocol.provider.pkgs` property in the JMX connection environment properties).
The recommendation to use the javagent is all good and well, and we do in most cases, but in more restricted settings it's often difficult to "just load this java agent".

brian-brazil:
That sounds like you'd need a custom build to pull in whatever propriety extensions you need, which is getting a bit beyond the scope of JMX.

bmesserer:
> That sounds like you'd need a custom build to pull in whatever propriety extensions you need, which is getting a bit beyond the scope of JMX.

Not sure why you think a custom build would be needed. This patch was tested to connect with weblogic remotely, because it just exposes the standard configuration mechanism of `JMXConnectorFactory.connect`, which takes a URL and properties (why would anyone pass properties if the URL was enough for configuration?).
This patch should be enough to connect to all "proprietary" JMX/RMI protocol implementations and fine-tune any standard connections - I wouldn't have asked to merge back if this weren't the case, i.e. if I didn't think this was a generally useful extension.
All one needs to do is: add the jars to the classpath and set the properties accordingly, so JMX loads the protocol classes. This is the "standard" extension mechanism of JMX/JNDI. What the Java creators were unfortunately missing back at the time was a proper, standardized "service provider locator", so they solved this via the "environment". The same should be usable for other servers, just with different properties and jars (for reference, see e.g. [here](http://docs.testplant.com/epp/9.0.0/ePP/jmx_monitor.htm), bottom).

As mentioned, I really wanted to make this more generally useful, not just satisfy my usecase like e.g. the [WebLogic fork](https://github.com/javafuns/weblogic_prometheus_jmx_exporter/blob/master/WEBLOGIC_README.md), but if you're opposed to adding this for some other reason, please mention, because I don't want to put work into futile explanations.

brian-brazil:
If this is required for WebLogic, then why have none of the other people using WebLogic asked about it?

bmesserer:
> If this is required for WebLogic, then why have none of the other people using WebLogic asked about it?

It's required for using WebLogic remotely via t3(s) (or JBoss/WildFly via jboss-remoting without JBoss classloading tricks or on a modularized JDK, which prohibits them, i.e. JDK9+ AFAIK), not for using the java agent. Maybe because there's a [fork](https://github.com/javafuns/weblogic_prometheus_jmx_exporter/blob/master/WEBLOGIC_README.md) which can be used or it requires really quite a bit of energy (see the discussion) to get a patch merged back which simply exposes the standard JMX configuration mechanism, used for a lot of stuff, just the more "obscure" one many people don't even know about? Or you're quick to redirect people with [these problems](https://github.com/prometheus/jmx_exporter/issues/249) to the mailing list which then doesn't [help](https://groups.google.com/forum/#!msg/prometheus-users/-uIGD2qWI5o/DcaVyidvAQAJ) and they'll just keep their private fork?
Maybe also because not many people use WebLogic and prometheus at the same time? I don't know, I can just tell you: neither we nor many others seem to have found a way to access WebLogic remotely via t3(s) without customizing the JMX connection. And `url` and `environment` *are* the parameters to configure the JMX connection. Why would it take two if one was always constant? See the WebLogic sample code I linked to.

As said, I don't know how else to "prove" this isn't just required for the use case, but has been crafted to be generally useful as far as possible as there are more properties (even standard ones) one can use there such as `jmx.remote.default.class.loader` (which could remove the problems people face e.g. [here](https://stackoverflow.com/questions/59521451/cannot-run-jmx-exporter-in-wildfly-17-running-on-java-11) (which needs jars included due to wrong remote classloader which won't work in Java 11 because of modules) because it's just the mechanism used to configure `JMXConnectorFactory.connect`.

Finally, there's *always* a chance this can be accomplished somehow differently, but after many have tried for quite some time (see all the links provided) and we researched and asked around, it doesn't seem like it's possible - why else would I have invested the effort to create this patch?

brian-brazil:
I'm trying to understand here what this t3 thing is, it sounds like it's not JMX. As I mentioned elsewhere it's not sane to try to support every single non-standard protocol that vendors come up with.

sgoeschl:
This "t3" thing is the following: "T3 is Oracle WebLogic Server’s proprietary RMI protocol. Basically, t3 protocol is a protocol that sits on top of Java RMI.". And in comes also with "t3s" to support secure communication.

bmesserer:
> I'm trying to understand here what this t3 thing is, it sounds like it's not JMX. As I mentioned elsewhere it's not sane to try to support every single non-standard protocol that vendors come up with.

It *is* JMX. As with many Java standards, JMX specifies the API, but can use a multitude of underlying protocols for the wire transfer, like er.g. JRMP, IIOP, T3, T3 over SSL/TLS, JBoss remoting, JBoss remoting over http(s) and so on. And the means to configure this are the environment properties (and the JMX URL).
The relationship is the same as with JMS: JMS is the API and can run over a lot of protocols, which are configured via properties passed to the connection factory. Or JNDI, which can use LDAP(s), AD, local, ... - again configured via connection factory properties.
Or, the most prominent example: JDBC and its DataSources.
This is a rather standard mechanism in "Java world", providing SPIs for standard interfaces to "plug into", which are then configured via properties, so the "standard interface" knows what service provider to "load".

brian-brazil:
Looking at the docs, you can set `jmx.remote.protocol.provider.pkgs` as a system property.

bmesserer:
> Looking at the docs, you can set `jmx.remote.protocol.provider.pkgs` as a system property.

That's correct. Unfortunately not `java.naming.security.`[`principal`|`credentials`].

brian-brazil:
That's a bit of a problem there, even if this were to be accepted. We already have code to set credentials per it's standard semantics.

bmesserer:
> That's a bit of a problem there, even if this were to be accepted. We already have code to set credentials per it's standard semantics.

Could you elaborate on tht please? Yes, I know this (see above), but what should I do there? Ofc, the `jmx.remote.credentials` are most used, as it's the JRMP config. Would it have been better to expose the jmx environment from the start vs. adding special properties? Sure. But I cannot just remove that code now, as that will introduce incompatibilities. And I cannot "detect conflicting" credentials, because how do I know which of these are supposed to be credential properties? 
Should I add the `java.naming.security` stuff as "conflicting" and fail startup (like with `jmxUrl` and `hostPort`)? Or any property which has `security` in its name?

brian-brazil:
> Yes, I know this (see above), but what should I do there? 

That's the problem. One is a list, the other a string so doing this generically isn't really practical.

bmesserer:
> That's the problem. One is a list, the other a string so doing this generically isn't really practical.

I'm at a loss here - can you please be more specific? Are you trying to say "you should support arrays/lists in generic properties"? Or "We don't want to support generic configuration because we already support the most common special case"? Or something entirely different?

brian-brazil:
I'm saying we already support the standard case. Trying to support configuring arbitrary data structures, which overlap with things we already have first class support for, doesn't seem sane to me.

If you're in a situation where you're already off the beaten track due to not using the agent, and the vendor inventing their own protocol and not supporting the standard protocl, I don't think it's completely unreasonable to have to do some extra work.

ahmgeek:
@brian-brazil Hey Brian, thank you very much for your work 🤗 
Big fan here especially of the amazing blog posts you write 😻 


What did you mean by:
>  don't think it's completely unreasonable to have to do some extra work

- Did you mean that whoever in such situation needs to fork the exporter and tailor it and move on?
Or
- They can do some effort and get it generalized enough so it could be merged back in the exporter?

And if you think this PR could be cleaned enough to be merged, what you suggest to do there?
Clear points would help us contribute back so other people would gain the profit.

Thanks again for the effort and time you put in this ❤️

brian-brazil:
The former as it stands.

These sorts of infinitely pluggable things work when you work with code, not when you only have configuration. We see the same sort of issues with AWS auth for example. It's fine if you know which one you want, not so much if you have to sanely support all of them.

brian-brazil:
I don't think there's anything we can do here unfortunately.

Issue #477, opened 2020-03-27
Wrong regex in zookeeper example config. 


rolatsch:
Hi there,

the current pattern for extracting the packet counts appends the current value of the counter to the measurement name.

```
# HELP zookeeper_PacketsReceived:_50 PacketsReceived (org.apache.ZooKeeperService<name0=ReplicatedServer_id3, name1=replica.3, name2=Leader><>PacketsReceived)
# TYPE zookeeper_PacketsReceived:_50 counter
zookeeper_PacketsReceived:_50{memberType="Leader",replicaId="3",} 50.0
```

This can be fixed by changing the pattern from
`org.apache.ZooKeeperService<name0=ReplicatedServer_id(\\d+), name1=replica.(\\d+), name2=(\\w+)><>(Packets.*)`

to `org.apache.ZooKeeperService<name0=ReplicatedServer_id(\\d+), name1=replica.(\\d+), name2=(\\w+)><>(Packets\\w+)`

brian-brazil:
Would you like to send a PR?

tomklino:
Hi @rolatsch , @brian-brazil !
This actually effects our setup too, and crashes the grafana explorer by exploding the response for the labels request from prometheus (https://our-dev-domain.com/grafana/api/datasources/proxy/1/api/v1/label/__name__/values) with millions of labels starting with zookeeper_PacketsReceived:NUMBER:

```

curl 'https://our-dev-domain.com/grafana/api/datasources/proxy/1/api/v1/label/__name__/values' -H 'Accept: application/json, text/plain, */*' --compressed -H 'Cookie: grafana_session=xxxx' | jq '.' | grep zookeeper_PacketsReceived | wc -l

1047216 # over a million labels!!!
```

Will your suggestion fix this issue? Is a PR coming soon or should I apply your fix manually for now?

brian-brazil:
There's already a PR, Github hadn't sent a notification that it had been force pushed to add the DCO but I've just merged it now.

PR #478, opened 2020-03-27
fix packet counter pattern in zookeeper sample config

rolatsch:
@brian-brazil fix for #477


brian-brazil:
You'll probably find it easiest to handle the DCO by squashing down to one commit.

brian-brazil:
Thanks!

PR #481, opened 2020-03-31
Fix incorrect 'java' path when JAVA_HOME isn't empty

roldevg:
Hi guys! 
You build the important staff, we are using your tool for our ActiveMq instance.

I made a little fix for this:
https://github.com/prometheus/jmx_exporter/issues/405


brian-brazil:
Thanks!

Issue #498, opened 2020-05-19
Multiple exporters on single JVM


jkimixl:
What would it take to allow for multiple javaagents to export the JMX info on different ports with different config files?  Use-case would be that you'd want to poll these ports at different intervals.

Current difficulty seems to come from #279 causing something like #290 .

If there's a workaround, then a pointer in that direction would be wonderful.  Thanks!

brian-brazil:
That's not supported, but I guess you could do a custom build. Generally you should try to keep things to one interval per application.

PR #499, opened 2020-05-20
Remove the use of the default, static CollectorRegistry

jkimixl:
It looks like #498 can be resolved with this simple change so that the CollectorRegistrys from multiple JavaAgents don't interfere with each other.  The change in the WebServer isn't strictly necessary, I think, but done to keep things parallel.

Submitted for your consideration.

### DIFF
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -3,6 +3,8 @@ package io.prometheus.jmx;
 import java.io.File;
 import java.lang.instrument.Instrumentation;
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'm not sure we should be complicating things like this, exporters should only use custom registries in very specific circumstances which don't apply to the jmx exporter.

jkimixl:
Maybe it doesn't factor into the WebServer as much, but the JavaAgent's CollectorRegistry.defaultRegistry is already not the "default" one in the sense that it's at io.prometheus.jmx.shaded.io.prometheus.client.CollectorRegistry rather than io.prometheus.client.CollectorRegistry.

I haven't tried, but I assume if you started the WebServer with a JavaAgent attached, the two registries will happily ignore each other.  Having a single CollectorRegistry per exporter seems like a perfectly good rule of thumb.  I don't know what specific circumstances you're thinking of, but JavaAgent already seems to be using a custom registry.

brian-brazil:
The java agent uses the default registry, that shading subsequently happens is orthogonal.

jkimixl:
Updated the PR to reflect my understanding of the best implementation of this change possible.
 
I can't believe I'm the first one to want different polling intervals for vm-level info as the application-level info.  That seems so useful to me, and the cost seems so minimal.

brian-brazil:
I believe you are the first, and I'm not at all convinced that it's wise to allow users to over-engineer their monitoring like this.

brian-brazil:
Per above I don't think this makes sense. If you really need this you can always do additional shading.

PR #504, opened 2020-06-18
Compile using java 11 and upgrade dependencies

amuraru:
- compile using java-11
- enforce min version of maven to 3.0.5 and upgraded plugin and dependencies versions
- added simple Dockerfile to build a container using jdk-11


brian-brazil:
Java 1.6 is explicitly supported, and using the http agent is not recommended so a Docker version of it should not be provided.

PR #518, opened 2020-08-11
Allow caching regular expression matching in rules

flavray:
This improves performance for some use cases.

The exporter is configured with a set of rules to match regular
expressions against bean names. Regular expressions can be CPU-intensive,
and matching the same bean names over the same patterns is
time-consuming when there are tens of thousands of bean names (e.g:
exposed by Kafka). Using a single `ConcurrentHashMap` to store the result
of pattern matching here.

The `MatchedRule` class was introduced here to "store" this result.
Each rule in the configuration leads to a `MatchedRule` which helps
build the set of metrics to expose to Prometheus.
The mapping between a bean name + attributes to a `MatchedRule` is
done once. Some (in Kafka, most) bean names may not match any rule,
so caching of `MatchedRule.unmatched()` helps prevent re-computation
of bean names to non-existent rules.

The logic to map a bean name to a Prometheus MetricFamilySample is
unchanged from the upstream, the `MatchedRule` was introduced only to
be able to cache the results.

Caching is disabled by default (behaviour unchanged), and can be enabled
by setting `cacheRules: true` in the configuration.

-----

In our case, this lead to reducing collection time from ~70 sec to ~3 sec on
some of our applications (Kafka brokers). I opened [a thread](https://groups.google.com/g/prometheus-developers/c/RNpcNOPwyeE) on the mailing list
a couple of weeks ago and haven't heard back, but definitely happy to discuss
and make changes if needed, if you think this is worth merging. 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,6 @@ value             | Value for the metric. Static values and capture groups from 
 valueFactor       | Optional number that `value` (or the scraped mBean value if `value` is not specified) is multiplied by, mainly used to convert mBean values from milliseconds to seconds.
 labels            | A map of label name to label value pairs. Capture groups from `pattern` can be used in each. `name` must be set to use this. Empty names and values are ignored. If not specified and the default format is not being used, no labels are set.
 help              | Help text for the metric. Capture groups from `pattern` can be used. `name` must be set to use this. Defaults to the mBean attribute description and the full name of the attribute.
-matchBeanValue    | Whether to append the bean value in the expression to match patterns against for this rule. (e.g: "beanName<beanType>: <beanValue>"). Defaults to `true`.
 cache             | Whether to cache bean name expressions to rule computation (match and mismatch). This can increase performance when collecting a lot of mbeans. Defaults to `false`.
 type              | The type of the metric, can be `GAUGE`, `COUNTER` or `UNTYPED`. `name` must be set to use this. Defaults to `UNTYPED`.
 
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ value             | Value for the metric. Static values and capture groups from 
 valueFactor       | Optional number that `value` (or the scraped mBean value if `value` is not specified) is multiplied by, mainly used to convert mBean values from milliseconds to seconds.
 labels            | A map of label name to label value pairs. Capture groups from `pattern` can be used in each. `name` must be set to use this. Empty names and values are ignored. If not specified and the default format is not being used, no labels are set.
 help              | Help text for the metric. Capture groups from `pattern` can be used. `name` must be set to use this. Defaults to the mBean attribute description and the full name of the attribute.
-cache             | Whether to cache bean name expressions to rule computation (match and mismatch). This can increase performance when collecting a lot of mbeans. Defaults to `false`.
+cache             | Whether to cache bean name expressions to rule computation (match and mismatch). Not recommended for rules matching on bean value only the first value will be cached and re-used. This can increase performance when collecting a lot of mbeans. Defaults to `false`.
 type              | The type of the metric, can be `GAUGE`, `COUNTER` or `UNTYPED`. `name` must be set to use this. Defaults to `UNTYPED`.
 
 Metric names and label names are sanitized. All characters other than `[a-zA-Z0-9:_]` are replaced with underscores,
--- a/README.md
+++ b/README.md
@@ -51,7 +51,6 @@ rules:
     valueFactor: 0.001
     labels: {}
     help: "Cassandra metric $1 $2"
-    matchBeanValue: true
     cache: false
     type: GAUGE
     attrNameSnakeCase: false
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ lowercaseOutputName: false
 lowercaseOutputLabelNames: false
 whitelistObjectNames: ["org.apache.cassandra.metrics:*"]
 blacklistObjectNames: ["org.apache.cassandra.metrics:type=ColumnFamily,*"]
+cacheRules: false
 rules:
   - pattern: 'org.apache.cassandra.metrics<type=(\w+), name=(\w+)><>Value: (\d+)'
     name: cassandra_$1_$2
@@ -57,16 +58,17 @@ rules:
 Name     | Description
 ---------|------------
 startDelaySeconds | start delay before serving requests. Any requests within the delay period will result in an empty metrics set.
-hostPort | The host and port to connect to via remote JMX. If neither this nor jmxUrl is specified, will talk to the local JVM.
-username | The username to be used in remote JMX password authentication.
-password | The password to be used in remote JMX password authentication.
-jmxUrl   | A full JMX URL to connect to. Should not be specified if hostPort is.
-ssl      | Whether JMX connection should be done over SSL. To configure certificates you have to set following system properties:<br/>`-Djavax.net.ssl.keyStore=/home/user/.keystore`<br/>`-Djavax.net.ssl.keyStorePassword=changeit`<br/>`-Djavax.net.ssl.trustStore=/home/user/.truststore`<br/>`-Djavax.net.ssl.trustStorePassword=changeit`
+hostPort   | The host and port to connect to via remote JMX. If neither this nor jmxUrl is specified, will talk to the local JVM.
+username   | The username to be used in remote JMX password authentication.
+password   | The password to be used in remote JMX password authentication.
+jmxUrl     | A full JMX URL to connect to. Should not be specified if hostPort is.
+ssl        | Whether JMX connection should be done over SSL. To configure certificates you have to set following system properties:<br/>`-Djavax.net.ssl.keyStore=/home/user/.keystore`<br/>`-Djavax.net.ssl.keyStorePassword=changeit`<br/>`-Djavax.net.ssl.trustStore=/home/user/.truststore`<br/>`-Djavax.net.ssl.trustStorePassword=changeit`
 lowercaseOutputName | Lowercase the output metric name. Applies to default format and `name`. Defaults to false.
 lowercaseOutputLabelNames | Lowercase the output metric label names. Applies to default format and `labels`. Defaults to false.
 whitelistObjectNames | A list of [ObjectNames](http://docs.oracle.com/javase/6/docs/api/javax/management/ObjectName.html) to query. Defaults to all mBeans.
 blacklistObjectNames | A list of [ObjectNames](http://docs.oracle.com/javase/6/docs/api/javax/management/ObjectName.html) to not query. Takes precedence over `whitelistObjectNames`. Defaults to none.
-rules    | A list of rules to apply in order, processing stops at the first matching rule. Attributes that aren't matched aren't collected. If not specified, defaults to collecting everything in the default format.
+cacheRules | Whether to cach bean to rule computation. If true, bean values are ommitted from patterns when matching.
+rules      | A list of rules to apply in order, processing stops at the first matching rule. Attributes that aren't matched aren't collected. If not specified, defaults to collecting everything in the default format.
 pattern  | Regex pattern to match against each bean attribute. The pattern is not anchored. Capture groups can be used in other options. Defaults to matching everything.
 attrNameSnakeCase | Converts the attribute name to snake case. This is seen in the names matched by the pattern and the default format. For example, anAttrName to an\_attr\_name. Defaults to false.
 name     | The metric name to set. Capture groups from the `pattern` can be used. If not specified, the default format will be used. If it evaluates to empty, processing of this attribute stops with no output.
--- a/README.md
+++ b/README.md
@@ -44,7 +44,6 @@ lowercaseOutputName: false
 lowercaseOutputLabelNames: false
 whitelistObjectNames: ["org.apache.cassandra.metrics:*"]
 blacklistObjectNames: ["org.apache.cassandra.metrics:type=ColumnFamily,*"]
-cacheRules: false
 rules:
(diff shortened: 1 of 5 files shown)


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -368,99 +380,134 @@ public void recordBean(
...
+          matchedRule = cachedRules.get(cacheName);
+        }
+
+        if (matchedRule == null) {
+          for (Rule rule : config.rules) {
+            Matcher matcher = null;
+            String matchName = beanName + (rule.attrNameSnakeCase ? attrNameSnakeCase : attrName);
+
+            // Using bean value in caching is not possible (caching is only done on bean name, values can change)
+            if (!config.cacheRules) {
```
brian-brazil, line 395: This creates two subtly different formats to match against, which is going to cause confusion.

flavray:
Good catch, I've added some logic to only retain beans that have just been scraped in the cache.

Regarding the per-rule caching, I think it makes sense to try to have per-rule caching. However I'm not able to find a decent way to support all of: per-rule caching, caching "unmatched rules" and allowing bean value in patterns.

If we allow patterns with the bean value in them, caching a miss for a given value will lead to false negatives; on the other hand, not caching unmatched rules will limit the performance of the cache. I could try to "detect" rules that have a bean value in them and treat them differently, but that seems really prone to errors too.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -368,99 +383,145 @@ public void recordBean(
...
+        LOGGER.fine("add metric sample: " + matchedRule.name + " " + matchedRule.labelNames + " " + matchedRule.labelValues + " " + value.doubleValue());
+        addSample(new MetricFamilySamples.Sample(matchedRule.name, matchedRule.labelNames, matchedRule.labelValues, value.doubleValue()), matchedRule.type, help);
       }
 
     }
 
-    public List<MetricFamilySamples> collect() {
-      if (configFile != null) {
+  // Remove stale rules (in the cache but not collected in the last run of the collector)
+  private void evictStaleCachedRules(Set<String> lastCachedRules) {
```
brian-brazil, line 505: This is never called.

flavray: 👍

flavray:
> > not caching unmatched rules will limit the performance of the cache.
> 
> This is what I'd suggest. If it matters to a user, the impact can be minimised by having those rules last.

Are you suggesting having a catch-all rule at the end of the list of rules, that would effectively cache unmatched bean names (at the expense of actually exposing them)? If so, that sounds reasonable to me. If not, happy to hear more about the idea!

In the meantime, I had switched to having a per-rule flag to cache regular expression matching. The code looks quite different from before (with the introduction of `MatchedRulesCache`) and I am not entirely conviced with the approach, still around dealing with rule matching against bean value, which has a different behaviour now (albeit limited to the given rule rather than every rule). Happy to have more feedback on this.

-----

Commit message

> Cache regular expression matching per rule
> 
> Instead of having a global `cacheRules` flag, have a per-rule `cache`
> flag (default: false). Rules that do not have the `cache` flag set will have the default
> behaviour.
> 
> In order to allow caching of rules with values that change over time
> (e.g: gauges), a new per-rule `matchBeanValue` flag has also been added
> (default: true). Disabling the flag will not add the bean value to the
> expression when matching against the list of rules (slightly different
> behaviour than before, but toggleable per rule) and allow caching of
> just the bean name (no matter the value).

brian-brazil:
> Are you suggesting having a catch-all rule at the end of the list of rules, that would effectively cache unmatched bean names (at the expense of actually exposing them)? 

I'm not, that'd be potentially incorrect. I'm proposing that the setting be per-rule, and if this is a concern for the user that they put the non-cached rules last so that they're not applied to everything that has already been matched. There's also already a way to halt processing of a given attribute.

README.md
```diff
@@ -51,30 +51,34 @@ rules:
     valueFactor: 0.001
     labels: {}
     help: "Cassandra metric $1 $2"
+    matchBeanValue: true
```
brian-brazil, line 54: We don't need two settings, they amount to the same thing.

flavray: Makes sense, thanks!

flavray:
> I'm not, that'd be potentially incorrect. I'm proposing that the setting be per-rule, and if this is a concern for the user that they put the non-cached rules last so that they're not applied to everything that has already been matched. There's also already a way to halt processing of a given attribute.

Gotcha. I've had another go at this, and removed the `matchBeanValue` flag.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -368,40 +390,52 @@ public void recordBean(
...
+
         for (Rule rule : config.rules) {
+          String cacheName = beanName + (rule.attrNameSnakeCase ? attrNameSnakeCase : attrName);
+          String matchName = cacheName + ": " + beanValue;
+
+          if (rule.cache) {
+            MatchedRule cachedRule = cachedRules.get(rule, cacheName);
+            if (cachedRule != null) {
+              addToCache(rule, cacheName, cachedRule);
+              if (!cachedRule.isUnmatched()) {
```
brian-brazil, line 403: The double negative is a bit confusing here

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -368,40 +390,52 @@ public void recordBean(
         String help = attrDescription + " (" + beanName + attrName + ")";
         String attrNameSnakeCase = toSnakeAndLowerCase(attrName);
 
+        MatchedRule matchedRule = MatchedRule.unmatched();
+
         for (Rule rule : config.rules) {
+          String cacheName = beanName + (rule.attrNameSnakeCase ? attrNameSnakeCase : attrName);
+          String matchName = cacheName + ": " + beanValue;
```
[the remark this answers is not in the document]
flavray: I've used `beanName + attrName`. Let me know if you'd like to see something else used.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -368,40 +390,52 @@ public void recordBean(
...
+
+          if (rule.cache) {
+            MatchedRule cachedRule = cachedRules.get(rule, cacheName);
+            if (cachedRule != null) {
+              addToCache(rule, cacheName, cachedRule);
```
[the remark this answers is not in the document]
flavray: Good point, I was adding it to make sure it was not considered stale. I've removed the call and just made a call to the staleness tracker directly.

README.md
```diff
@@ -51,30 +51,32 @@ rules:
...
+value             | Value for the metric. Static values and capture groups from the `pattern` can be used. If not specified the scraped mBean value will be used.
+valueFactor       | Optional number that `value` (or the scraped mBean value if `value` is not specified) is multiplied by, mainly used to convert mBean values from milliseconds to seconds.
+labels            | A map of label name to label value pairs. Capture groups from `pattern` can be used in each. `name` must be set to use this. Empty names and values are ignored. If not specified and the default format is not being used, no labels are set.
+help              | Help text for the metric. Capture groups from `pattern` can be used. `name` must be set to use this. Defaults to the mBean attribute description and the full name of the attribute.
+cache             | Whether to cache bean name expressions to rule computation (match and mismatch). This can increase performance when collecting a lot of mbeans. Defaults to `false`.
```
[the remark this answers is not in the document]
flavray: Good point, I've updated the README. Let me know if you'd like to see more/fewer details.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -437,30 +471,53 @@ public void recordBean(
...
 
-    public List<MetricFamilySamples> collect() {
-      if (configFile != null) {
+  public List<MetricFamilySamples> collect() {
+    if (configFile != null) {
         long mtime = configFile.lastModified();
         if (mtime > config.lastUpdate) {
           LOGGER.fine("Configuration file changed, reloading...");
           reloadConfig();
+          cachedRules.clear();  // rules may have changed with the configuration, clear the rule cache
```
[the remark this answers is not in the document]
flavray: Good catch! I've tried to get a decent thread-safe implementation.

collector/src/main/java/io/prometheus/jmx/MatchedRulesCache.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * MatchedRulesCache is a cache for bean name to configured rule mapping (See JmxCollector.Receiver).
+ * The cache also retains unmatched entries (a bean name not matching a rule pattern) to avoid
+ * matching against the same pattern in later bean collections.
+ */
+public class MatchedRulesCache {
+    private final Map<Entry, MatchedRule> cachedRules;
```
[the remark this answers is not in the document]
flavray: Done. It makes the logic inside MatchedRulesCache a little bit more tricky to be thread-safe, but it should be fine now.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -368,40 +389,52 @@ public void recordBean(
         String help = attrDescription + " (" + beanName + attrName + ")";
         String attrNameSnakeCase = toSnakeAndLowerCase(attrName);
 
+        MatchedRule matchedRule = MatchedRule.unmatched();
+
         for (Rule rule : config.rules) {
+          String cacheKey = beanName + attrName;
+          String matchName = beanName + (rule.attrNameSnakeCase ? attrNameSnakeCase : attrName) + ": " + beanValue;
```
brian-brazil, line 396: If we're using the cache, can we put in a dummy value such as cacheEnabled for the bean value so things will obviously break for anyone who tries to extract something from it?

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -437,30 +470,62 @@ public void recordBean(
...
-      if (configFile != null) {
+  public List<MetricFamilySamples> collect() {
+    MatchedRulesCache cachedRules = this.cachedRules;
+
+    if (configFile != null) {
         long mtime = configFile.lastModified();
         if (mtime > config.lastUpdate) {
           LOGGER.fine("Configuration file changed, reloading...");
           reloadConfig();
+          synchronized (this) {
```
brian-brazil, line 516: It'd be better to do all of this inside reloadConfig, under a lock.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -437,30 +470,62 @@ public void recordBean(
...
+        LOGGER.fine("add metric sample: " + matchedRule.name + " " + matchedRule.labelNames + " " + matchedRule.labelValues + " " + value.doubleValue());
+        addSample(new MetricFamilySamples.Sample(matchedRule.name, matchedRule.labelNames, matchedRule.labelValues, value.doubleValue()), matchedRule.type, help);
       }
 
     }
 
-    public List<MetricFamilySamples> collect() {
-      if (configFile != null) {
+  public List<MetricFamilySamples> collect() {
+    MatchedRulesCache cachedRules = this.cachedRules;
```
brian-brazil, line 509: This should be under a lock.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -109,19 +109,18 @@ private void reloadConfig() {
...
+        synchronized (this) {
+          long mtime = configFile.lastModified();
+          if (mtime > config.lastUpdate) {
+            LOGGER.fine("Configuration file changed, reloading...");
+            reloadConfig();
           }
         }
       }
+
+      return config;
```
brian-brazil, line 123: This reference to config need to come from inside the lock.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -109,19 +109,18 @@ private void reloadConfig() {
       }
     }
 
-    private void maybeReloadConfig() {
+    private Config maybeReloadConfig() {
```
brian-brazil, line 112: getLatestConfig would be a better name now

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
flavray: Good point. I've made the whole method `synchronized` now.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -109,15 +109,13 @@ private void reloadConfig() {
       }
     }
 
-    private Config maybeReloadConfig() {
+    private synchronized Config getLastConfig() {
```
brian-brazil, line 112: Latest, not Last

flavray: Good catch, updated. 🙂

brian-brazil:
Thanks!

flavray:
Awesome! Thank you for all the constructive feedback!

amuraru:
@qinghui-xu thanks for this improvement. Would it make sense to add an example of how to use cache functionality in https://github.com/prometheus/jmx_exporter/blob/master/example_configs/kafka-2_0_0.yml ?

brian-brazil:
PRs welcome to update the existing example configs to use this.

PR #527, opened 2020-09-15
Change white/blacklist naming convention

richardbadman:
This is just a small PR to change the naming convention of two variables that are used within this repository, coming from a diversity and inclusion point of view, the proposed changes in this PR don't have negative racial historic connotations attached with them.

I have also tested this working on a tiny HBase cluster of 4 machines

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:1234/jmxrmi
 ssl: false
 lowercaseOutputName: false
 lowercaseOutputLabelNames: false
-whitelistObjectNames: ["org.apache.cassandra.metrics:*"]
-blacklistObjectNames: ["org.apache.cassandra.metrics:type=ColumnFamily,*"]
+allowedlistObjectNames: ["org.apache.cassandra.metrics:*"]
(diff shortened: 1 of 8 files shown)


brian-brazil:
Thanks for your PR. Changing these names is on the cards, however it is a breaking change in a widely used tool. There is interest and someone in the community who said they were working on expanding these particular options to work on not just the object names but also the attribute names, which would also be a breaking change. Accordingly to minimise disruption to end users, I want to combine breaking changes as much as is possible, and have this altered as part of that functionality expansion rather than breaking the same config fields twice in a row.


Also I think processed/ignored would make more sense here, as we're not allowing/denying anything.

richardbadman:
Thanks for the feedback, with that in mind, should I close this PR in favour of word going on elsewhere that someone is working on?

brian-brazil:
Let's leave it open in case anyone else is wondering, as while it has always been the plan it wasn't written down until now.

dhoard:
I have a draft PR https://github.com/prometheus/jmx_exporter/pull/852 in progress for this change.

Existing configurations using `whitelistObjectNames` and `blacklistObjects` will continue to work, but we should consider them deprecated.

dhoard:
Closing. Implemented and merged changes.

PR #533, opened 2020-09-22
Performance improvements

massamany:
Hi @brian-brazil,

This PR regroups 4 commits, each of them allowing to optimize a little the exporter code.
The first 3 can not change the behavior.
The last commit, about MBeanInfo caching, could have impact. I don't think a MBeanInfo should change, but if you think this can happen may be a configuration param could allow to deactivate.

Please tell me what you think.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -11,6 +11,7 @@ import java.io.InputStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -73,20 +74,25 @@ public class JmxCollector extends Collector implements Collector.Describable {
     private File configFile;
     private long createTimeNanoSecs = System.nanoTime();
 
-    private final JmxMBeanPropertyCache jmxMBeanPropertyCache = new JmxMBeanPropertyCache();
+    private JmxMBeanPropertyCache jmxMBeanPropertyCache = new JmxMBeanPropertyCache();
 
     public JmxCollector(File in) throws IOException, MalformedObjectNameException {
+        this((Map<String, Object>)new Yaml().load(new FileReader(in)));
         configFile = in;
-        config = loadConfig((Map<String, Object>)new Yaml().load(new FileReader(in)));
         config.lastUpdate = configFile.lastModified();
     }
 
     public JmxCollector(String yamlConfig) throws MalformedObjectNameException {
-      config = loadConfig((Map<String, Object>)new Yaml().load(yamlConfig));
+        this((Map<String, Object>)new Yaml().load(yamlConfig));
     }
 
     public JmxCollector(InputStream inputStream) throws MalformedObjectNameException {
-      config = loadConfig((Map<String, Object>)new Yaml().load(inputStream));
+        this((Map<String, Object>)new Yaml().load(inputStream));
+    }
+
+    private JmxCollector(Map<String, Object> load) throws MalformedObjectNameException {
+        config = loadConfig(load);
+        createScraper();
     }
 
     private void reloadConfig() {
@@ -98,6 +104,7 @@ public class JmxCollector extends Collector implements Collector.Describable {
           config = loadConfig(newYamlConfig);
           config.lastUpdate = configFile.lastModified();
           configReloadSuccess.inc();
+          createScraper();
         } catch (Exception e) {
           LOGGER.severe("Configuration reload failed: " + e.toString());
           configReloadFailure.inc();
@@ -248,6 +255,15 @@ public class JmxCollector extends Collector implements Collector.Describable {
 
         return cfg;
 
+    }
+
+    /**
+     * Creates JMX Scraper
+     */
+    private void createScraper() {
+        jmxMBeanPropertyCache.onlyKeepMBeans(Collections.<ObjectName>emptySet());
+        scraper = new JmxScraper(config.jmxUrl, config.username, config.password, config.ssl,
+                config.whitelistObjectNames, config.blacklistObjectNames, config.detectUnusedJmxAttributes, jmxMBeanPropertyCache);
     }
 
     static String toSnakeAndLowerCase(String attrName) {
@@ -529,6 +545,7 @@ public class JmxCollector extends Collector implements Collector.Describable {
       }
     }
 
+    JmxScraper scraper;
   public List<MetricFamilySamples> collect() {
       // Take a reference to the current config and collect with this one
       // (to avoid race conditions in case another thread reloads the config in the meantime)
@@ -536,8 +553,7 @@ public class JmxCollector extends Collector implements Collector.Describable {
 
       MatchedRulesCache.StalenessTracker stalenessTracker = new MatchedRulesCache.StalenessTracker();
       Receiver receiver = new Receiver(config, stalenessTracker);
-      JmxScraper scraper = new JmxScraper(config.jmxUrl, config.username, config.password, config.ssl,
-              config.whitelistObjectNames, config.blacklistObjectNames, config.detectUnusedJmxAttributes, receiver, jmxMBeanPropertyCache);
+
       long start = System.nanoTime();
       double error = 0;
       if ((config.startDelaySeconds > 0) &&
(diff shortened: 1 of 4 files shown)


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -244,22 +246,28 @@ private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObject
 
     }
 
+    private static final ConcurrentHashMap<String, String> snakeCache = new ConcurrentHashMap<String, String>();
```
brian-brazil, line 249: This can leak memory in certain (unwise) configurations, so we shouldn't do this.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -124,42 +125,52 @@ public void doScrape() throws Exception {
         }
     }
 
+    private static final ConcurrentHashMap<ObjectName, Map<String, MBeanAttributeInfo>> cache =
```
brian-brazil, line 128: This is also a memory leak.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
massamany: In fact it *could* be if MBeans attributes names were changed at each scrape. This won't be the case : the list of "snake cased" attributes will be stable over time I guess.
What's true is the map would hold lots of Entries. But as they're Strings, the memory footprint is not heavy.
Again, may be this can be disabled via configuration if necessary.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
massamany: Yes indeed this is possible. So may be a system of purge can be put in place. Even though I think monitored Beans should be stable more or less and the Map content won't change a lot after initialization.

massamany:
Just a general comment on this PR : it might not be taken "as is" and certainly should be reworked. I'm just proposing concept to optimize scraping, especially in remote JMX access cases, when sometimes response times can be really long.
For instance, in my own case, I had 8 seconds of response time, now I'm below 5.5 seconds. I think many users could appreciate that.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
brian-brazil: It's probably best not to make that assumption, there's a lot of weird stuff out there. Anything like this should work correctly out of the box, without any configuration options.

brian-brazil (commented):
I think the first two commits are fine, the others don't look safe as-is.

massamany:
I agree with you that the statics are not so safe.
So I see several options :

- the impact of caching the snake cased strings is not so big, and can be simply suppressed.
- but the impact of caching the MBeansinfoAttributes is very visible. I propose to avoid the static map, which has other side effect than memory leaks (for instance, in case of several collectors scraping different versions of the same software, collision is possible). With the Map as an instance field, it's safer. As it will be provided by the constructor, a null Map can mean "no cache".

Also, I agree with you that if everything works out of the box it's better. But a parameter to activate cache (meaning it would be deactivated by default to avoid regressions on existing configurations) would not be so annoying don't you think ?

brian-brazil:
> but the impact of caching the MBeansinfoAttributes is very visible.

The question isn't if it's useful, but can it leak memory? Do you have a proposal that avoids that?

> But a parameter to activate cache (meaning it would be deactivated by default to avoid regressions on existing configurations) would not be so annoying don't you think ?

I'd prefer to have things correct, rather than a potentially dangerous option that helps performance but is effectively hidden in documentation. None of this should be configurable, as it shouldn't affect semantics.

massamany:
> Do you have a proposal that avoids that?

The proposal I think about would be to implement a cleanup of the cache at each scrape. The steps would be :
- Retrieve the candidate MBeans from the MBean server (_already done_)
- For each MBean candidate :
  - Check the cache to retrieve MBeanAttributeInfo if present (_already submitted in PR_)
  - If not present, retrieve MBeanAttributeInfo from the MBean server (_already done_)
  - Update the cache with new MBeanAttributeInfo (_already submitted in PR_)
- At the end, clean the cache to remove entries which were not used in the current run

With this implementation, the same amount of memory would used by the cache that the one which would be used by the objects retrieved from the MBean server. So, no configuration would be needed as it would not change the behaviour, except that the object would be resident in memory instead of beeing re-created at every scrape.

massamany:
By the way, we did not discuss about the last commit, which aims to detect if some attributes should not be retrieved, with eventual recursive calls to MBean Servers in case of composite objects or tabular data.

The concept is simple : if an attribute has not been matched by any rule at the first scrape, it won't be retrieve at the next ones.

This has a great impact also on memory, time, etc ; because it can avoid some recursive calls. But may be it can be improved. I essentially wonder if some rules with particular regex on the value could be affected. What's your opinion on this one ?

brian-brazil:
> The steps would be :

That's basically what we do elsewhere.

> The concept is simple : if an attribute has not been matched by any rule at the first scrape, it won't be retrieve at the next ones.

That's not a safe assumption, as whether something can match can depend on the value.

massamany:
> That's basically what we do elsewhere

OK so if I understand well, we're good with this solution. I'll make the implementation and update the commits.

> That's not a safe assumption, as whether something can match can depend on the value.

That's what I said in the comment, and the reason why I opened the discussion. The concept is proven to have positive impact, but is not perfect because there is one case when it can have undesirable side effect, so is there a solution to avoid this side effect (for instance detect that a match is performed on value and could change the match) ? My idea is to share and improve the system globally for everyone, not only patching the source on my own to satisfy my own use case.

brian-brazil:
> so is there a solution to avoid this side effect (for instance detect that a match is performed on value and could change the match)

I don't believe there's a correct way to do it, even detecting rules where it may happen is probably impossible.

massamany:
Hi,

I just finished integrating jmx_exporter in my project, and therefore to investigate for optimizations.

In order to share these researches and help the community, I provide all this in this PR.

Main optimization ways are :
- Caching Attribute Info (very strong effect)
- Excluding unused attributes from fetching (strong effect - on this one indeed I can't auto detect in a 100% sure way, so it's deactivated by defaut with a param - maybe somebody would find a great algorithm and manage to do it)
- Caching JMX connections (strong effect, with auto detection of closed / failed connections)
- Other little optimizations with little effect.

At the end, I had a case which took more or less 6 seconds to fetch. I'm under 2 seconds now.

I read, on lots of forums, that this exporter is good, but very slow. I think these modifications could be helpful for lots of users. With all the points we discussed (thanks for your time by the way), it appears that behavior should not change, so no risks of regression. I admit that the "all works out of the box" thing is a little disturbed by a new param :-) , but I think it's worth it.

So, thanks again for your time and work on all this open source projects which are great. And keep me updated if you need some modifications in this PR.

brian-brazil:
I'm happy to take the first two commits, and the third if the memory leak is dealt with.

massamany:
Hi.

The memory leak is dealt with, as discussed together.

You should take a look at the JMX connection caching too, which avoids open / close at each scrape. I added this one yesterday.

Cheers,

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -93,82 +111,132 @@ public void doScrape() throws Exception {
...
+                mBeanNames.remove(instance.getObjectName());
             }
-        } finally {
-          if (jmxc != null) {
-            jmxc.close();
-          }
+        }
+
+        // Now that we have *only* the whitelisted mBeans, remove any old ones from the cache:
+        jmxMBeanPropertyCache.onlyKeepMBeans(mBeanNames);
```
brian-brazil, line 166: Can we combine these caches?

brian-brazil:
> You should take a look at the JMX connection caching too, which avoids open / close at each scrape. I added this one yesterday.

I don't think it's wise to add so much complexity for a operational mode that is not recommended in the first place. If you care about performance, you'll use the agent.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
massamany: Hi.

Sorry for the delay, had no time these last days.
Good point. Done.

massamany:
> If you care about performance, you'll use the agent.

Don't agree. This mode is a great advantage if you can't modify remote environment. I've read in various forums that lots of people use this mode.
Of course, all optims brought in the PR are more or less dedicated to remote mode, because on a local JVM these latencies can not be seen.

So, I insist on the vision that all tuning that technical people may apply to optimize are good to take, but I agree with you that "out of the box" the product should run and provide correct result, even if not in an optimized way. That's why I implemented the new param in this way.

brian-brazil:
As I said I'll take the first 3 commits, but not the others. I'm not planning on removing the option of remote, but neither am I willing to make this already complex exporter even more complex for the sake of a performance improvement for a method that has fundamental performance issues. Performance improvements need to be justified based on the agent, which is the recommended way to use this.

PR #540, opened 2020-10-07
Roll back snakeyaml version from 1.26 to 1.16 to fix java 6 support

YangMuye:
Snakeyaml dropped support for java 6 from 2.14+, so we revert back to an earlier version to maintain java 6 support.

### DIFF
--- a/collector/pom.xml
+++ b/collector/pom.xml
@@ -29,7 +29,8 @@
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
-      <version>1.26</version>
+      <!-- uses an older version because 2.24+ requires Java 7 -->
+      <version>1.16</version>
     </dependency>
   </dependencies>
 


collector/pom.xml
```diff
@@ -29,7 +29,8 @@
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
-      <version>1.26</version>
+      <!-- uses an older version because 2.24+ requires Java 7 -->
+      <version>1.16</version>
```
brian-brazil, line 33: Can we use .23?

YangMuye: In theory, yes, but I don't know if there will be any gotchas.
I'm stuck with some last century systems, and satisfied with 1.16 so far. But I can try 1.23 later this week.

brian-brazil: .18 is what the previous bump was trying to avoid, so let's see if we can make everyone happy.

brian-brazil:
Superseded by #575

PR #543, opened 2020-10-12
updated spark example config for spark 3.0

wobu:
- added "type" value to Regex so resulting prometheus metrics are identical to spark 2.4 version

@brian-brazil 

i don't know the details, but it seems that a "type=counters" or "type=gauge" has been added to the JMX metrics of spark in 3.0 version.
So i added this type strings to the regex to emit this value from the resulting prometheus metric name so we have a config for spark 2.4 and 3.0 resulting in the same metric names.

### DIFF
--- a/example_configs/spark-3-0.yml
+++ b/example_configs/spark-3-0.yml
@@ -2,17 +2,17 @@ rules:
 
   # These come from the master
   # Example: master.aliveWorkers
-  - pattern: "metrics<name=master\\.(.*)><>Value"
+  - pattern: "metrics<name=master\\.(.*), type=counters><>Value"
     name: spark_master_$1
 
   # These come from the worker
   # Example: worker.coresFree
-  - pattern: "metrics<name=worker\\.(.*)><>Value"
+  - pattern: "metrics<name=worker\\.(.*), type=counters><>Value"
     name: spark_worker_$1
 
   # These come from the application driver
   # Example: app-20160809000059-0000.driver.DAGScheduler.stage.failedStages
-  - pattern: "metrics<name=(.*)\\.driver\\.(DAGScheduler|BlockManager|jvm)\\.(.*)><>Value"
+  - pattern: "metrics<name=(.*)\\.driver\\.(DAGScheduler|BlockManager|jvm)\\.(.*), type=gauges><>Value"
     name: spark_driver_$2_$3
     type: GAUGE
     labels:
@@ -20,13 +20,13 @@ rules:
 
   # These come from the application driver
   # Emulate timers for DAGScheduler like messagePRocessingTime
-  - pattern: "metrics<name=(.*)\\.driver\\.DAGScheduler\\.(.*)><>Count"
+  - pattern: "metrics<name=(.*)\\.driver\\.DAGScheduler\\.(.*), type=counters><>Count"
     name: spark_driver_DAGScheduler_$2_total
     type: COUNTER
     labels:
       app_id: "$1"
 
-  - pattern: "metrics<name=(.*)\\.driver\\.HiveExternalCatalog\\.(.*)><>Count"
+  - pattern: "metrics<name=(.*)\\.driver\\.HiveExternalCatalog\\.(.*), type=counters><>Count"
     name: spark_driver_HiveExternalCatalog_$2_total
     type: COUNTER
     labels:
@@ -34,7 +34,7 @@ rules:
 
   # These come from the application driver
   # Emulate histograms for CodeGenerator
-  - pattern: "metrics<name=(.*)\\.driver\\.CodeGenerator\\.(.*)><>Count"
+  - pattern: "metrics<name=(.*)\\.driver\\.CodeGenerator\\.(.*), type=counters><>Count"
     name: spark_driver_CodeGenerator_$2_total
     type: COUNTER
     labels:
@@ -42,14 +42,14 @@ rules:
 
   # These come from the application driver
   # Emulate timer (keep only count attribute) plus counters for LiveListenerBus
-  - pattern: "metrics<name=(.*)\\.driver\\.LiveListenerBus\\.(.*)><>Count"
+  - pattern: "metrics<name=(.*)\\.driver\\.LiveListenerBus\\.(.*), type=counters><>Count"
     name: spark_driver_LiveListenerBus_$2_total
     type: COUNTER
     labels:
       app_id: "$1"
 
   # Get Gauge type metrics for LiveListenerBus
-  - pattern: "metrics<name=(.*)\\.driver\\.LiveListenerBus\\.(.*)><>Value"
+  - pattern: "metrics<name=(.*)\\.driver\\.LiveListenerBus\\.(.*), type=gauges><>Value"
     name: spark_driver_LiveListenerBus_$2
     type: GAUGE
     labels:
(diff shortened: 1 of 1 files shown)


example_configs/spark-3-0.yml
```diff
@@ -0,0 +1,116 @@
...
+  - pattern: "metrics<name=(.*)\\.driver\\.HiveExternalCatalog\\.(.*), type=counters><>Count"
+    name: spark_driver_HiveExternalCatalog_$2_total
+    type: COUNTER
+    labels:
+      app_id: "$1"
+
+  # These come from the application driver
+  # Emulate histograms for CodeGenerator
+  - pattern: "metrics<name=(.*)\\.driver\\.CodeGenerator\\.(.*), type=counters><>Count"
+    name: spark_driver_CodeGenerator_$2_count
```
brian-brazil, line 39: Counters should end in _total

brian-brazil (commented):
Can you share example output metrics?

wobu:
These are output metrics of an spark executor

[spark 3.0 with new example_configs/spark-3-0.yml](https://github.com/prometheus/jmx_exporter/files/5363975/spark_3_0_ExecutorPrometheusMetrics.txt)
[spark 2.4 with example_configs/spark.yml](https://github.com/prometheus/jmx_exporter/files/5363976/sparkExecutorPrometheusMetrics.txt)

brian-brazil:
Other than the above the output looks largely okay. spark_executor_filesystem_hdfs_read_ops, spark_executor_filesystem_hdfs_read_ops and spark_executor_filesystem_hdfs_read_bytes smell more like counters than gauges though - is this just bad tagging on their end?

wobu:
yes, i guess the hdfs_metrics are all gauges, at least to the documentation in spark 3.0, see: https://spark.apache.org/docs/3.0.0/monitoring.html#component-instance--executor

for me they seem to be tagged correctly. Spark Uses ".Value" for gauges, and they are additionally tagged with "type=gauges" in spark 3.0.

brian-brazil:
What do the values mean though?

wobu:
https://spark.apache.org/docs/3.0.0/monitoring.html#list-of-available-metrics-providers

They say:
- metrics which end to "count" are counters
- everything else are gauges

brian-brazil:
That doesn't mean that they're semantically counters or gauges, as Dropwizard definitions are based on which API you chose to use rather than the meaning of the metric. What do the metrics actually mean?

wobu:
i don't know yet whats this metrics mean, i just wanted to have for both versions of spark (2.4 or 3.0) to 
get the same metrics for prometheus.

brian-brazil:
> "spark_executor_filesystem_hdfs_read_bytes" is declared as a gauge, so i expect that this value displays the bytes read at this moment when seen. (I still hope this is a mistake, i hoped for a counter type actually...)

This is why I suspect it's actually a counter in Prometheus terms. Can you check the code?

wobu:
in the corresponding spark implementation they are using the types Gauge or Counter accordingly to their documention. 
When they use the Counter type, they are using the DropWizard API and call there increase method. In case of a Gauge type, they are just returning the scalar value.

In case of the filesystem metrics, they are using the integrated statistics of the [Apache Hadoop Common Filesystem API](https://github.com/apache/hadoop-common/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2801)
They handle the metric increments internally.

So in my opinion it seems indeed to be a counter metric, but nothing which i can change easily.
What would happen in Prometheus when having semantically a counter metric but declaring it as Gauge?

Code snippets from the Spark Filesystem metric integration ([Full SourceCode](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/executor/ExecutorSource.scala)):
```
  private def fileStats(scheme: String) : Option[FileSystem.Statistics] =
        
    FileSystem.getAllStatistics.asScala.find(s => s.getScheme.equals
```

```
  private def registerFileSystemStat[T](
        scheme: String, name: String, f: FileSystem.Statistics => T, defaultValue: T) = {
    metricRegistry.register(MetricRegistry.name("filesystem", scheme, name), new Gauge[T] {
      override def getValue: T = fileStats(scheme).map(f).getOrElse(defaultValue)
    })
  }
```

```
  for (scheme <- Array("hdfs", "file")) {
    registerFileSystemStat(scheme, "read_bytes", _.getBytesRead(), 0L)
    registerFileSystemStat(scheme, "write_bytes", _.getBytesWritten(), 0L)
    registerFileSystemStat(scheme, "read_ops", _.getReadOps(), 0)
    registerFileSystemStat(scheme, "largeRead_ops", _.getLargeReadOps(), 0)
    registerFileSystemStat(scheme, "write_ops", _.getWriteOps(), 0)
  }
```

brian-brazil:
> What would happen in Prometheus when having semantically a counter metric but declaring it as Gauge?

Nothing, but it's confusing for users and may cause issues downstream.

From the standpoint of this exporter, if there's a mix of types and you can't practically distinguish you should use untyped rather than gauge.

wobu:
okay, i'll have look on this metrics when i am testing in detail and update the exporter config accordingly

wobu:
i guess this is a counter metric, see:

grafik

i will change it in both configuration files and update the pull reuqest

wobu:
i updated the config files and added sample outputs.
Also created an Issue @ Apache Spark for the wrong metric type: https://issues.apache.org/jira/browse/SPARK-33127

example_configs/spark-3-0.yml
brian-brazil: This still applies.

example_configs/spark-3-0.yml
```diff
@@ -0,0 +1,131 @@
...
+
+  # These come from the application driver
+  # Emulate timers for DAGScheduler like messagePRocessingTime
+  - pattern: "metrics<name=(.*)\\.driver\\.DAGScheduler\\.(.*), type=counters><>Count"
+    name: spark_driver_DAGScheduler_$2_count
+    type: COUNTER
+    labels:
+      app_id: "$1"
+
+  # HiveExternalCatalog is of type counter
```
brian-brazil, line 29: This is redundant.

example_configs/spark-3-0.yml
```diff
@@ -0,0 +1,131 @@
...
+  - pattern: "metrics<name=(.*)\\.driver\\.spark\\.streaming\\.(.*)\\.(.*), type=gauges><>Value"
+    name: spark_driver_structured_streaming_$3
+    labels:
+      app_id: "$1"
+      query_name: "$2"
+
+  # These come from the application executors
+  # Example: app-20160809000059-0000.0.executor.threadpool.activeTasks (value)
+  #  app-20160809000059-0000.0.executor.JvmGCtime (counter)
+  - pattern: "metrics<name=(.*)\\.(.*)\\.executor\\.(?!filesystem)(.*), type=gauges><>Value"
```
brian-brazil, line 78: Generally switch the order of the rules rather than using negative lookaheads.

brian-brazil (commented):
The output shouldn't be in the PR.

example_configs/spark-3-0.yml
wobu: Didn't know the rules have an order and are mutually exclusive. Will test and change accordingly.

example_configs/spark-3-0.yml
wobu: Could you be more specific? This is a spark driver metric, i didn't change there anything except for adding the metric type

brian-brazil: The comment isn't needed.

brian-brazil:
Can you add the DCO? Squashing is generally easiest.

wobu:
squashed and signed commit

brian-brazil:
Thanks!

PR #553, opened 2020-11-04
better scraper for getting multiple attributes

damonYuan:
`attributes = beanConn.getAttributes(mbeanName, name2AttrInfo.keySet().toArray(new String[0]));` 

In the line above, if one of the Attributes for some reasons cannot be fetched (eg., java.rmi.UnmarshalException in my case) from remote server, the whole batch of attributes will be discarded.

This PR is to accommodate for this case so that valuable data will be kept when some of them might throw unexpected exceptions.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -20,14 +20,7 @@ import javax.naming.Context;
 import javax.rmi.ssl.SslRMIClientSocketFactory;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -146,7 +139,7 @@ class JmxScraper {
             }
             name2AttrInfo.put(attr.getName(), attr);
         }
-        final AttributeList attributes;
+        AttributeList attributes;
         try {
             attributes = beanConn.getAttributes(mbeanName, name2AttrInfo.keySet().toArray(new String[0]));
             if (attributes == null) {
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. I don't think this makes sense overall, and the correct approach is to fix the application in question to not throw exceptions when queried via JMX. Technically we should be failing the entire scrape when there's a single mBean that has an exception.This could also lead to unpredictable performance.

damonYuan:
Thanks Brain,

> the correct approach is to fix the application in question to not throw exceptions when queried via JMX

As a jmx client you don't always have the access to fix the issue happens in jmx server. If you try to connect some java application (in my case, Confluence) using JMC, you will find some fields are Null. This is how JMC handles the exception - it just ignores those cannot be parsed and still collect those that valuable for that bean. Can I fix a JMX exception for a non-opensource Confluence App?

> Technically we should be failing the entire scrape when there's a single mBean that has an exception.

As the example of JMC above, I think we shouldn't discard all the treasures when only one rock inside.

> This could also lead to unpredictable performance.

Shall we consider the performance before we know its impact? If most of the beans don't have any problems in their attributes as you imply, then there should be no performance impact; if many beans is problematic in their attributes, the current implementation will cause lots of troubles to developers like me - I spent several days to work on the rules try to catch the missing attributes compared with the JMC with no luck, and I hope it won't cost troubles to others again. And you might miss lots of metrics if anyone are careless / lazy to compare to the JMC / check the logs.

brian-brazil:
> Can I fix a JMX exception for a non-opensource Confluence App?

It sounds like this is a bug you should take up with your vendor then, and ask them to implement JMX properly.

> I think we shouldn't discard all the treasures when only one bug inside.

Think about what happens if this is intermittent. That would be effectively impossible for users to deal with correctly, as there's so many ways partial data could happen - plus the PromQL to deal with that isn't the simplest either.


jconsole is not the jmx_exporter, it's a debugging tool for ad-hoc manual use - not a tool that is trying to regularly extract data with specific semantics on a strict deadline. Comparisons with it aren't very useful here as the use case is quite different.

As I've indicated I think think this makes sense, this is a matter between you and your vendor. In all previous cases for things like this it turns out that fixing the software was easy, once someone actually tried so I expect it's the same here.

PR #557, opened 2020-11-18
Code to reset the MBean metrics after each scrape

NiallWard:
We had a project requirement to reset the metrics on the MBeans each time we scrape in order to avoid any smoothing affect of the averages reported over time. This is the code we used to achieve that, the ability to turn the functionality on/off as well as the list of MBeans to enable this functionality for is controllable via the config yml file

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -59,6 +59,8 @@ public class JmxCollector extends Collector implements Collector.Describable {
       boolean ssl = false;
       boolean lowercaseOutputName;
       boolean lowercaseOutputLabelNames;
+      boolean resetMetrics;      
+      List<ObjectName> MBeansToReset = new ArrayList<ObjectName>();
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks for your PR. This would breaks the ability to have multiple scrapers, which is an important part of the Prometheus design. It would also have the jmx exporter perform mutating operations, which would be a security concern. You're also making assumptions about how mBeans are being laid out. Accordingly this feature does not belong in the exporter.

Reporting averages since the process started is not workable in general, I suggest improving the software you're using to expose raw counters so this can be calculated properly.

NiallWard:
Thanks for the review. Just so I can report onwards, the PR was rejected because it was deemed undesirable to allow the exporter to have the ability to change the underlying metrics (in this case by resetting them to 0 after each scrape ) as other clients (to your multiple scraper point) would see unexpected results. Would this be a fair summary?

brian-brazil:
Yes, one reason is that it's fundamentally in conflict with the Prometheus architecture.

PR #580, opened 2021-02-04
Process JMX beans with an ArrayList value

sdarlington:
Based on the discussion in other pull requests, I'm creating this "for debate" rather than with the expectation that will be merged in its current state!

Basically, Apache Ignite (and others as far as I can tell) export certain JMX metrics as an ArrayList. It's not possible to directly bring this into Prometheus but returning the number of entries works reasonably well in this case. Are there better alternatives?

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -20,14 +20,7 @@ import javax.naming.Context;
 import javax.rmi.ssl.SslRMIClientSocketFactory;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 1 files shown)


fstab:
Hi Stephen, sorry for responding so late.

Reading Apache Ignite's documentation I see that they have a JMX bean `Connections` with a list of connections. In that case the size gives you the number of connections, which is a valuable metric.

However, I fear that this might be an exception. I suspect that in almost all cases the size of the list is not a meaningful metric at all. Are there other examples of List beans where the size is a meaningful metric?

sdarlington:
Fundamentally, there's a difference between how OpenCensus and JMX see the world. I see three options:

1. If there's no direct mapping, we just ignore the values (the current behaviour)
2. A hack, like this PR. It's not going to work for some (many? most?) use cases, but it may be better than nothing
3. A general solution for mapping non-compliant JMX metrics to _something_ in OpenCensus

I'm not sure what 3 would look like, so I went for 2. Without knowing what's in the array, I'm not sure what else we could do but count the number of entries?

I can also see the logic in 1, though I'd prefer to see it documented if that's the direction the project takes.

PR #584, opened 2021-02-18
Bump snakeyaml to 1.27

7vs:
A vulnerability scan of the prom/jmx_exporter image finds this.
CVE | SEVERITY | CVSS | PACKAGE | VERSION | STATUS | PUBLISHED | DISCOVERED | DESCRIPTION
-- | -- | -- | -- | -- | -- | -- | -- | --
CVE-2017-18640 | high | 7.50 | org.yaml_snakeyaml | 1.17 | fixed in 1.26 | > 10 months | < 1 hour | The Alias feature in SnakeYAML 1.18 allows entity expansion during a load operation, a related issue to CVE-2003-1564.

@brian.brazil@robustperception.io

### DIFF
--- a/collector/pom.xml
+++ b/collector/pom.xml
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
-      <version>1.23</version>
+      <version>1.27</version>
     </dependency>
   </dependencies>
 


fstab (approved):
LGTM, Thanks 7vs!

brian-brazil:
This PR broke Java 6 support, and thus I think it should be reverted. The listed CVE is not a concern for this exporter, as the YAML is from a trusted source.

fstab:
Ah, thanks for the heads-up @brian-brazil. I reverted it. I guess I'll extend our circle ci tests to cover Java 6, so that it's easier to check this in the future.

bliof-fc:
@fstab @brian-brazil How about releasing a dedicated jar for java 6?

e.g.
- jmx_prometheus_javaagent-0.16.0.jar
- jmx_prometheus_javaagent-0.16.0-jdk6.jar  - tbo I don't really get how these are named but here is how other do it http://www.bouncycastle.org/latest_releases.html

The problem we are having is that we do security scans our docker images and have processes to force people to fix their issues. And we have to log exceptions and keep explaining that this is not really a vulnerability. 

The main benefit of fixing this is that you'll reduce the amount of communication in random other companies :D and this may be masking a real problem in the actual jars of people.

ps.  Happy to help with something of the sort but I am not very familiar with the java world.

PR #585, opened 2021-03-04
Update snakeyaml version to fix CVE-2017-18640

gurbuzali:
Update snakeyaml version to 1.26 to remediate the [CVE-2017-18640](https://nvd.nist.gov/vuln/detail/CVE-2017-18640) vulnerability.

### DIFF
--- a/collector/pom.xml
+++ b/collector/pom.xml
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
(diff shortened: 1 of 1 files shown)


fstab:
Thanks for the PR @gurbuzali. We had a similar PR already, but reverted it because it breaks Java 6 compatibility #584. The CVE is not a concern for this exporter, as the YAML is from a trusted source.

fstab:
I added a comment in `pom.xml`.

PR #587, opened 2021-03-06
Add support for java.util.Optional MBean properties

thomasgl-orange:
The SonarQube maintainers had this weird idea of an `Optional<Long>` property in an mbean:
https://github.com/SonarSource/sonarqube/blob/8.7.0.41497/server/sonar-ce/src/main/java/org/sonar/ce/monitoring/CeTasksMBean.java#L37
(IMHO, having a regular `long` property, and returning `0` when no task is currently pending, would have been fine too, but that's how it is...)

This currently can't be scraped via the `jmx_exporter` JVM agent. So here we go, a modest proposal to support `java.util.Optional`, of course only when the agent runs on Java 8+. This implementation uses Java reflection to find the `java.util.Optional` class and its `orElse(Object)` method. If found, then when an mbean property `value` is indeed an `Optional`, it will be processed as `value.orElse(null)`.

Caveats:
- don't feed it a self-referencing `Optional`, I've put no safeguard against infinite recursion on `JmxScraper.processBeanValue(...)` calls :-)
- there is no unit test. I've checked that this code works as expected on the SonarQube CE process, but I can't figure how to write a test in a Java 1.6 project. One option could be to add a Java 8 Maven submodule just to hold this test (via a profile activated only when building with Java 8+, or something like that), but honestly, that would be ugly.

Ping: @fstab and @tomwilkie

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -54,6 +54,7 @@ class JmxScraper {
     private final boolean ssl;
     private final List<ObjectName> whitelistObjectNames, blacklistObjectNames;
     private final JmxMBeanPropertyCache jmxMBeanPropertyCache;
+    private final OptionalValueExtractor optionalValueExtractor = new OptionalValueExtractor();
 
     public JmxScraper(String jmxUrl, String username, String password, boolean ssl,
                       List<ObjectName> whitelistObjectNames, List<ObjectName> blacklistObjectNames,
@@ -282,6 +283,16 @@ class JmxScraper {
             }
         } else if (value.getClass().isArray()) {
             logScrape(domain, "arrays are unsupported");
+        } else if (optionalValueExtractor.isOptional(value)) {
+            logScrape(domain + beanProperties + attrName, "java.util.Optional");
+            processBeanValue(
+                    domain,
+                    beanProperties,
+                    attrKeys,
+                    attrName,
+                    attrType,
+                    attrDescription,
+                    optionalValueExtractor.getOptionalValueOrNull(value));
         } else {
             logScrape(domain + beanProperties, attrType + " is not exported");
         }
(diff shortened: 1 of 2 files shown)


fstab:
Thanks for the PR @thomasgl-orange. I will have a closer look in the next couple of days.

fstab (approved):
Thanks for the PR. The code looks good, and does not break Java 6 support. I'm happy to merge it. 

A few notes:

SonarCube clearly violate the JMX specification, as `Optional` is not one of the allowed data types. However, in the spirit of the [Robustness Principle](https://en.wikipedia.org/wiki/Robustn
ess_principle) it's good that `jmx_exporter` still deals with `Optional` if possible.

One thing to note is that `Optional` as an `MBean` value can only occur if `jmx_exporter` is attached as an agent, it can never occur when `jmx_exporter` connects remotely. The reason is that
 remote JMX uses the RMI protocol, and transmitting an `Optional` over RMI yields a `NotSerializableException`:

```
java.rmi.UnmarshalException: error unmarshalling return; nested exception is:
        java.io.WriteAbortedException: writing aborted; java.io.NotSerializableException: java.util.Optional
        at sun.rmi.server.UnicastRef.invoke(UnicastRef.java:197)
        at com.sun.jmx.remote.internal.PRef.invoke(Unknown Source)
        at javax.management.remote.rmi.RMIConnectionImpl_Stub.getAttributes(Unknown Source)
        at javax.management.remote.rmi.RMIConnector$RemoteMBeanServerConnection.getAttributes(RMIConnector.java:928)
        at io.prometheus.jmx.JmxScraper.scrapeBean(JmxScraper.java:153)
        at io.prometheus.jmx.JmxScraper.doScrape(JmxScraper.java:119)
        at io.prometheus.jmx.JmxCollector.collect(JmxCollector.java:547)
        ...
```

This is one of the reasons why `Optional` and other non-serializable types are not supported by the JMX specification.

In agent mode, the `jmx_exporter` runs in the same JVM as the application, so we will never see the combination where an application uses `Optional` but the `jmx_exporter` runs in Java 6.

tobias-trabelsi-sonarsource:
FYI @fstab and @thomasgl-orange we hear you and created a [bug ticket](https://jira.sonarsource.com/browse/SONAR-15057) to fix this on our side as well.
thanks for taking care of this here :)

Issue #286, opened 2018-06-21
using env variables in conf file


azimovMichael:
Is there any way to inject environment variables as placeholders in the conf file?
If not, do you think its something to think about? wouldnt mind doing it myself if it will seem needed

brian-brazil:
Prometheus configuration file are intended to be dumb, and in non-trivial setups to come from configuration management. In addition we like to have only one way to do things and one place a given bit of configuration can come from, so environment variables are not supported. There are tools like `envsubst` which you can use if you need to hack something together.

May I ask why you want to substitute with the jmx exporter? I'd expect the entire file to be hardcoded, with no variation between instances.

brian-brazil:
That should be handled with a target label on the Prometheus end from service discovery/relabelling.

azimovMichael:
To be extra specific, I need it for kafka consumer, I need to add consumer group id as a label, but kafka doesnt export it as jmx so i need to add it manually, but I want to share the config file between all my services, SO i dont want to hardcode the label value but instead to inject it from a env var.

azimovMichael:
Ok  that actually sounds like a giod solution.  I will check it out. thank you!

PR #594, opened 2021-03-29
Enable substituting env vars in the config

wshihadeh:
Enable using environment variables for defined labels, as shown below 

```
     labels:
          application: "${APPLICATION}"
          host: "${HOSTNAME}"
          simplified_host: "${SIMPLIFIEDHOSTNAME}"
```


wshihadeh:
@fstab  can you take a look here ?

mindw:
Not a maintainer, but the idea was rejected before - #286 . Maintainers have changed since though.

dhoard:
For an application on a machine/VM, you could do this 1) via your CI/CD process as part of the deployment or 2) as a pre-execution step/process.  (i.e. build the `config.yaml` then install or execute the application.)

(Bash examples https://stackoverflow.com/questions/9928492/search-and-replace-variables-in-a-file-using-bash-sed)

For a K8s application, see my response in https://github.com/prometheus/jmx_exporter/issues/636 for a solution.

(I'm not a maintainer)

PR #604, opened 2021-05-07
Logging level to SEVERE for Exception

shaipan:
Currently, all exceptions are logged as FINE level. The FINE level gives too much log. We should be having SEVERE level for Exceptions.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -130,10 +130,10 @@ class JmxScraper {
         try {
           info = beanConn.getMBeanInfo(mbeanName);
         } catch (IOException e) {
-          logScrape(mbeanName.toString(), "getMBeanInfo Fail: " + e);
+          logScrape(mbeanName.toString(), "getMBeanInfo Fail: " , e);
           return;
         } catch (JMException e) {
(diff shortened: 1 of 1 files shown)


fstab:
Thanks for the PR. Did you get these exceptions in a real world scenario? If so, it would be interesting to hear what happened. I'd like to avoid spamming user's log files with SEVERE messages, and it's hard to decide what is severe without an actual example.

shaipan:
> Thanks for the PR. Did you get these exceptions in a real world scenario? If so, it would be interesting to hear what happened. I'd like to avoid spamming user's log files with SEVERE messages, and it's hard to decide what is severe without an actual example.

We are setting an VM -Dsoption un.rmi.transport.tcp.responseTimeout=30000. And our scrapping interval is 25s. So if some bean is taking more time to scrap than the scrapping interval, we wanted to capture it in the logs so that the problem bean can be identified. But due to FINE logging, the logs generated are too much and we cant keep all the logs in the production environment. So wanted to keep only SEVERE logs. On seeing java.net.SocketTimeoutException we can know that this particular bean took more time than 30s.

dhoard:
@shaipan what is the status of the PR? Can you resolve the changes as well as complete the checks (DCO)?

dhoard:
Closing due to lack of activity.

Issue #633, opened 2021-09-13
Kafka connect, how to expose kafka connector status ?


ivalexa:
Hi,

We are using jmx_exporter to export Kafka Connect JMX metrics to Prometheus. In the example config files there:
https://github.com/prometheus/jmx_exporter/blob/master/example_configs/kafka-connect.yml
There is a metric stat that exposes the connector task status:
```
  #kafka.connect:type=connector-task-metrics,connector="{connector}",task="{task}<> status"
  - pattern: 'kafka.connect<type=connector-task-metrics, connector=(.+), task=(.+)><>status: ([a-z-]+)'
    name: kafka_connect_connector_status
    value: 1
    labels:
      connector: "$1"
      task: "$2"
      status: "$3"
    help: "Kafka Connect JMX Connector status"
    type: GAUGE
```
However when the connector fails to start none of the metrics being exposed, in Kafka Connect rest API you can expose connector status that returns connector state:

```
GET /connectors/my-connector/status 
```
https://docs.confluent.io/platform/current/connect/references/restapi.html#get--connectors-(string-name)-status
`"connector": {
        "state": "RUNNING",
        "worker_id": "fakehost:8083"
    }`

I'm wondering if it's possible to get the same metric from the JMX exporter, looking in confluent official docs(https://docs.confluent.io/home/connect/monitoring.html#connector-metrics), looks like that there is MBean `kafka.connect:type=connector-metrics,connector=”{([-.w]+)` which should export it, can anyone help me with correct pattern to explore it, please, thnx!.

ivalexa:
Pull request created to expose the correct MBean https://github.com/prometheus/jmx_exporter/pull/635

PR #635, opened 2021-09-15
Align kafka connect metrics

ialexivy:
Currently, the Kafka connect task status is exposed under kafka_connect_connector_status which is not the connector status, and if the connector failed to start nothing gets exposed, I have added the Kafka connect connector metrics and renamed the existing status to kafka_connect_connector_task_status, more details are provided here:
#633
Signed-off-by: alexivsn <alex@securenative.com>


dhoard:
Closed due to inactivity.

PR #650, opened 2021-11-08
Use artifactId to generate debian package name

Skunnyk:
- Since 77a5b8fe17520a097429cf5f47021990c0eb6438, "name" is "Prometheus
  JMX Exporter - Http Server", which break debian package generation
- use artifactId to get a correct name

### DIFF
--- a/jmx_prometheus_httpserver/src/deb/control/control
+++ b/jmx_prometheus_httpserver/src/deb/control/control
@@ -1,4 +1,4 @@
-Package: [[name]]
+Package: [[artifactId]]
 Version: [[version]]
 Section: misc
 Priority: optional


Skunnyk:
This restore previous behavior:

godvoigt:
It would bee great to get this merged. We prefer installing debian packages over hand crafting installations, but this bug is preventing us from using newer versions of jmx exporter.

Skunnyk:
Hi :)

Is it possible for this PR to be merged ? This is really a small change affecting debian package build (which is advertised in the [README](https://github.com/prometheus/jmx_exporter#installing)), and it would be super awesome to have it fixed for the next release  :pray:

fstab:
Thanks a lot for the reminder and for the PR, it's merged now.

PR #655, opened 2021-11-23
Cassandra: disable two particularly slow metrics

russss:
Disabling these two metrics speeds up the scrape time dramatically on our Cassandra cluster - from around 20 seconds to 2-3 seconds.

I think it's worth disabling these by default because it makes the JMX exporter much more usable with Cassandra and this trick isn't particularly obvious.


dhoard:
@russss Thanks for the PR!!!

PR #671, opened 2022-01-10
Added more documentation to README

davidmpaz:
Hi,

could we please add this to main stream, for dummies like me? In my case it was not straight clear what to do in order to get things up and running. I think this could contribute to newbies to get onboard better.

Thanks and cheers.
David

-----

Changes:

Added suggestions on running the agent when getting Exceptions:

io.prometheus.jmx.shaded.io.prometheus.jmx.JmxCollector.collect JMX scrape failed:
java.io.IOException: Failed to retrieve RMIServer stub: javax.naming.NoInitialContextException:
Can not instantiate class: org.apache.naming.java.javaURLContextFactory

Added more steps to use it as http server.

PR #674, opened 2022-01-16
Updated README to add beginner hints

dhoard:
Signed-off-by: Doug Hoard <doug.hoard@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,17 +19,37 @@ The Java agent is available in two versions with identical functionality:
 
 The only difference between these versions is the version of the bundled snakeyaml dependency. See [release notes](https://github.com/prometheus/jmx_exporter/releases/tag/parent-0.16.1) for more info.
 
+Please note that due to the nature of JMX the `/metrics` endpoint might exceed Prometheus default scrape timeout of 10 seconds.
+
+### Agent mode
+
 To run as a Java agent, download one of the JARs and run:
 ```
-java -javaagent:./jmx_prometheus_javaagent-0.16.1.jar=8080:config.yaml -jar yourJar.jar
+java -javaagent:./jmx_prometheus_javaagent-0.16.1.jar=8080:config.yaml -jar <your_application.jar>
 ```
+
 Metrics will now be accessible at http://localhost:8080/metrics
 
 To bind the java agent to a specific IP change the port number to `host:port`.
 
-See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
+__Note:__ The configuration file **must not** specify _hostPort_ or _jmxUrl_ when running in agent mode.
 
-Please note that due to the nature of JMX the `/metrics` endpoint might exceed Prometheus default scrape timeout of 10 seconds.
+### Standalone mode
+
+To run in standalone mode, run your application with external JMX access:
+
+```
+java -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=5555 -jar <your_application.jar>
+```
+
+Download one of the JARs and run:
+```
+java -jar ./jmx_prometheus_javaagent-0.16.1.jar 8080 config.yml
+```
+
+Metrics will now be accessible at http://localhost:8080/metrics
+
+__Note:__ The configuration file **must** specify _hostPort_ or _jmxUrl_ when running in standalone mode
 
 ## Building
 


dhoard:
@fstab please review

README.md
```diff
@@ -27,6 +29,17 @@ Metrics will now be accessible at http://localhost:8080/metrics
...
 To bind the java agent to a specific IP change the port number to `host:port`.
 
+__Note:__ Within the config.yaml...
+
+```
+hostPort: localhost:5555
+username: 
+password:
+```
+... is typically not required.
```
fstab, line 39: I just created #675 which would make this config illegal in agent mode. So I guess we can remove this note now.

README.md
```diff
@@ -27,6 +29,17 @@ Metrics will now be accessible at http://localhost:8080/metrics
...
+
+### As an HTTP Server
```
fstab, line 41: As I'm thinking about it, _HTTP Server_ seems to be a bad name, as it implies the agent does not have an HTTP server (see #671).

It's hard to change that consistently, as the `jmx_prometheus_httpserver` module literally has `httpserver` in its name. However, maybe we should call this section "Standalone mode" as opposed to "agent mode"?

README.md
```diff
@@ -27,6 +29,17 @@ Metrics will now be accessible at http://localhost:8080/metrics
...
+
 See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
```
fstab, line 43: Not directly related to this PR, but since we have a `jmx_example_application` in our integration tests, the `./run_sample_httpserver.sh` seems like a weird example. Maybe it's less confusing to exemplify both modes using the example application and to remove `./run_sample_httpserver.sh`:

Agent mode

```
java -javaagent:jmx_prometheus_javaagent/target/jmx_prometheus_javaagent-0.16.2-SNAPSHOT.jar=8080:example_configs/agent_sample_config.yml -jar integration_tests/jmx_example_application/target/jmx_example_application.jar
```

Standalone mode

```
java -javaagent:jmx_prometheus_javaagent/target/jmx_prometheus_javaagent-0.16.2-SNAPSHOT.jar=8080:example_configs/httpserver_sample_config.yml -jar integration_tests/jmx_example_application/target/jmx_example_application.jar

java -jar jmx_prometheus_httpserver/target/jmx_prometheus_httpserver-0.16.2-SNAPSHOT-jar-with-dependencies.jar 5556 example_configs/httpserver_sample_config.yml
```

What do you think?
```

**Review by fstab — CHANGES_REQUESTED** [review]
Thanks for the PR. I added a few thoughts. What do you think?

File: README.md (hunk 2 above)
Comment (dhoard) [reply README.md:41]: `Standalone mode` sounds great.

File: README.md (hunk 1 above)
Comment (dhoard) [reply README.md:39]: I agree, https://github.com/prometheus/jmx_exporter/pull/675 makes it obsolete, but it _may_ be worth keeping for a new user since this seems like a common configuration error.

I have no strong opinion either way, but I know that @davidmpaz ran into this issue since the README wasn't clear.

File: README.md (hunk 3 above)
Comment (dhoard) [reply README.md:43]: I agree, remove `./run_sample_httpserver.sh` and update the documentation.

It might make sense for your to merge my branch into your work so we can deliver it as a single commit?

File: README.md (hunk 1 above)
Comment (fstab) [reply README.md:39]: If we keep it we could replace it with a sentence saying _Note that the configuration file must not specify hostPort or jmxUrl when running in agent mode_.

Then we should also add the opposite note to the standalone section: _Note that the configuration file must specify hostPort or jmxUrl when running in standalone mode_.

The current _is typically not required_ is no longer true, because it will explicitly result in an error.

File: README.md (hunk 3 above)
Comment (fstab) [reply README.md:43]: I could as well merge #675 right away and then we have your PR for the README.

Comment (fstab) [reply README.md:43]: I merged it :)

**Comment by dhoard** [comment]
@fstab please review. (I had to update/fix my branch)

File: README.md [hunk 4]
```diff
@@ -27,9 +31,22 @@ Metrics will now be accessible at http://localhost:8080/metrics
...
 
-See `./run_sample_httpserver.sh` for a sample script that runs the httpserver against itself.
+__Note:__ The configuration file **must not** specify _hostPort_ or _jmxUrl_ when running in agent mode.
 
-Please note that due to the nature of JMX the `/metrics` endpoint might exceed Prometheus default scrape timeout of 10 seconds.
+### Standalone mode
+
+To run in standalone mode, download one of the JARs and run:
+```
+java -javaagent:jmx_prometheus_javaagent/target/jmx_prometheus_javaagent-0.16.2-SNAPSHOT.jar=8080:example_configs/httpserver_sample_config.yml -jar integration_tests/jmx_example_application/target/jmx_example_application.jar
```
fstab, line 40: The `-javaagent` flag is for running it in agent mode, i.e. attaching the agent directly to the JVM at startup. In standalone mode you would run the application without the agent, but with a JMX remote port exposed, and then connect the `jmx_exporter` to that port.

fstab (changes requested):

README.md
dhoard: Good catch. Mistake on my part. Will try to fix it this weekend.

Issue #673, opened 2022-01-16
Terminate with an error if started as in agent mode but configured with a remote JMX url


fstab:
`jmx_exporter` can run in an agent mode (attaching directly to the JVM to be motitored) or in a standalone mode (connecting to a remote JVM to be monitored).

If you want to connect to a remote JVM you have to configure either the `jmxUrl` or the `hostPort`.

Here's an overview of the possible configurations:

### 1 Running in agent mode without configuring jmxUrl or hostPort

This is as expected, if you run `jmx_exporter` in agent mode you don't need to connect to a remote JVM.

### 2 Running in standalone mode and configuring jmxUrl or hostPort

This is as expected, if you run `jmx_exporter` in standalone mode you want to configure the remote connection to the JVM to be monitored.

### 3 Running in standalone mode without configuring jmxUrl or hostPort

This is now prevented and will result in a startup error. https://github.com/prometheus/jmx_exporter/blob/5c2c9216622342116b0068d4cc47f71a37093d3a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java#L101-L112

Previously, what happened was that `jmx_exporter` starts and monitors its own JVM, which is very likely not what you want.

### 4 Running in agent mode and configuring jmxUrl or hostPort

This is currently possible. What happens is that `jmx_exporter` attaches to a JVM, but instead of monitoring that JVM it tries to connect to a remote JVM to monitor.

# Proposal

Modify the behavior in case 4 and make `jmx_exporter` terminate with an error. If you run `jmx_exporter` in  agent mode attaching to a JVM you certainly don't want to use it to monitor a remote JVM.

dhoard:
I agree... case 4 doesn't seem like something we would want to support.

If a user wants to run Tomcat/Jetty (or another JEE servlet container) to monitor an application (as opposed to the built-in `HTTPServer`), they really should be using `client_java` and `MetricsServlet` directly.

dhoard:
Closed as resolved.

PR #675, opened 2022-01-16
Prevent remote JMX monitoring when started in agent mode

fstab:
This PR implements the proposal from #673.


fstab:
I added the single quotes. No strong opinion on that.

Issue #682, opened 2022-02-15
[config] Kafka percent metrics loss 


maixiaohai:
Some important metrics (kafka.server<type=KafkaRequestHandlerPool, name=RequestHandlerAvgIdlePercent><>MeanRate) do not show.

- Add MeanRate Percent metrics for kafka in example config

PR #692, opened 2022-03-07
[ISSUE #682] Add MeanRate Percent metrics for kafka in example config

maixiaohai:
Signed-off-by: maixiaohai <maixiaohai00@gmail.com>

#682 
Fix some percent metrics loss

### DIFF
--- a/example_configs/kafka-2_0_0.yml
+++ b/example_configs/kafka-2_0_0.yml
@@ -101,3 +101,17 @@ rules:
   type: GAUGE
   labels:
     quantile: "0.$4"
+
+# Generic gauges for MeanRate Percent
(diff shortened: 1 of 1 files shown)


fstab (approved):

fstab:
Thank you!

brian-brazil:
MeanRate should not be exposed as it is the average rate since application start and is thus not useful, see https://prometheus.io/docs/instrumenting/writing_exporters/#drop-less-useful-statistics

I'm also unclear on why the other metrics are being called `_percent` considering that's a) not the base unit of a ratio b) that raw denominator/numerator metrics are preferred and c) that a DropWizard `Value` a counter and thus can't be a ratio/percentage gauge.

To do this correctly, you want to expose the Value/Count values as counters and then take a rate in Prometheus.

Issue #696, opened 2022-03-25
Allow blacklisting of jvm metrics


udesaiitrs:
Hello, I was hoping to understand why certain JVM metrics (java.lang:*) cannot be blacklisted/removed via configuration of the javaagent? The exporter should include these as there can be instances like in our case where we do not want the duplicate JVM metrics because they are collected by another application.

udesaiitrs:
I have found how there are many default exports registered: https://github.com/prometheus/jmx_exporter/blob/ca972bfe693cca03d3a0fac802968e33828a1dd9/jmx_prometheus_javaagent_java6/src/main/java/io/prometheus/jmx/JavaAgent.java#L30

I feel as though it would be a good option to implement a configuration option to allow these to be turned off (on by default), especially with the growing scenarios such as Kubernetes environments where all the extra metrics may not be needed. 

Is this something that sounds plausible?

gmariette:
will appreciate such feature to, to not spawn our Prom cluster with the jvm container metrics

dhoard:
This sounds like a great enhancement. It would require changes to ...

https://github.com/prometheus/client_java/blob/main/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/DefaultExports.java

... to allow unregistering the collectors to support dynamic configuration reloading.

dhoard:
Or as an alternative not use `DefaultExports.initialize();` but add configuration-specific options and call the underlying exports classes directly.

```
new BufferPoolsExports().register(registry);
new ClassLoadingExports().register(registry);
new CompilationExports().register(registry);
new GarbageCollectorExports().register(registry);
new MemoryAllocationExports().register(registry);
new MemoryPoolsExports().register(registry);
new StandardExports().register(registry);
new ThreadExports().register(registry);
new VersionInfoExports().register(registry);
```

dhoard:
@udesaiitrs @gmariette, I am working on the Java agent feature. A few usage questions...

I can disable the Prometheus-generated JVM metrics (`JvmMetrics.builder().register(DEFAULT_REGISTRY);`) easy enough via an `exporter.yaml` configuration value.

Do you all want/expect JVM-generated metrics from JVM MBeans to also be filtered?

i.e., automatically append the JVM MBean names to `excludeObjectNames`/`blacklistObject` during runtime? (... the `exporter.yaml` would not be affected (no-rewrite.)

Thoughts?

udesaiitrs:
[...]

PR #698, opened 2022-03-25
Optionally disable all default exports for JavaAgent

udesaiitrs:
Signed-off-by: udesaiitrs <udesai@itrsgroup.com>

### DIFF
--- a/jmx_prometheus_javaagent_java6/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent_java6/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -26,8 +26,12 @@ public class JavaAgent {
             Config config = parseConfig(agentArgument, host);
 
             new BuildInfoCollector().register();
-            new JmxCollector(new File(config.file), JmxCollector.Mode.AGENT).register();
-            DefaultExports.initialize();
+            new JmxCollector(new File(config.file)).register();
+
+            if (!config.disableDefaultExports) {
+                DefaultExports.initialize();
+            }
+
             server = new HTTPServer(config.socket, CollectorRegistry.defaultRegistry, true);
         }
         catch (IllegalArgumentException e) {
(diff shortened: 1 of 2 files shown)


udesaiitrs:
Fix #696

fstab:
Sorry for the delay. As @dhoard mentioned, there is a branch for a new YAML config file format that we are working on. It was originally started for extended SSL configuration options, but it also introduces advanced filtering which will achieve what you need.

Here's an example: https://github.com/prometheus/jmx_exporter/blob/new-config/integration_tests/ssl_tests/src/test/resources/test-config-new.yaml

It's still work in progress. Do you think this would solve the issue?

udesaiitrs:
> Sorry for the delay. As @dhoard mentioned, there is a branch for a new YAML config file format that we are working on. It was originally started for extended SSL configuration options, but it also introduces advanced filtering which will achieve what you need.
> 
> Here's an example: https://github.com/prometheus/jmx_exporter/blob/new-config/integration_tests/ssl_tests/src/test/resources/test-config-new.yaml
> 
> It's still work in progress. Do you think this would solve the issue?

Sorry for the delay, yes something like this is exactly what we're looking for. Any eta on when it will be released?

weidox:
Seems to be stuck. And we're still waiting

udesaiitrs:
> Sorry for the delay. As @dhoard mentioned, there is a branch for a new YAML config file format that we are working on. It was originally started for extended SSL configuration options, but it also introduces advanced filtering which will achieve what you need.
> 
> Here's an example: https://github.com/prometheus/jmx_exporter/blob/new-config/integration_tests/ssl_tests/src/test/resources/test-config-new.yaml
> 
> It's still work in progress. Do you think this would solve the issue?

Yes this would solve the issue I believe, is there an ETA on merging/releasing this change?

dhoard:
Resolved with PR https://github.com/prometheus/jmx_exporter/pull/1215

PR #702, opened 2022-04-20
issue-695: revert bulk loading of attributes

### LINKED ISSUE #695: java.rmi.UnmarshalException while scraping metrics
hello all

we upgraded the jmx-exporter and collector in our application to a newer version:
- collector from 0.10 to 0.16.1
- `jmx_prometheus_javaagent` from 0.11.0 to 0.16.1

the goal of our application is to scrape metric of different servers (tomcat, weblogic). it's a standalone app and we do not use any tomcat or weblogic dependencies . 

now we are facing some issues while scraping bean metrics. if we use the old dependency, we are able to get the thread metrics from our weblogic server like:

```
[2022-03-21 15:28:25] FINE    scrape: 'com.bea:ServerRuntime=node1,Name=ThreadPoolRuntime,Type=ThreadPoolRuntime'_'StuckThreadCount': process
[2022-03-21 15:28:25] FINE    scrape: 'com.bea{ServerRuntime=node1, Name=ThreadPoolRuntime, Type=ThreadPoolRuntime}StuckThreadCount': 0
[2022-03-21 15:28:25] FINE    add metric sample: weblogic_threadpool_stuck_thread_count [] [] 0.0
```

With the new dependencies we get an error:
```
"java.rmi.UnmarshalException: error unmarshalling return; nested exception is: \n\tjava.lang.ClassNotFoundException: weblogic.management.runtime.ExecuteThread (no security manager: RMI class loader disabled)"
 detail = {ClassNotFoundException@43203} "java.lang.ClassNotFoundException: weblogic.management.runtime.ExecuteThread (no security manager: RMI class loader disabled)"
```

btw: if i use jconsole i sucessfully retrieve the metrics

Any help is very appreciated!
Regards Angela

faenschi:
@fstab closes #695

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -138,14 +138,49 @@ class JmxScraper {
         }
         MBeanAttributeInfo[] attrInfos = info.getAttributes();
 
+        Map<String, MBeanAttributeInfo> name2AttrInfo = new LinkedHashMap<String, MBeanAttributeInfo>();
         for (int idx = 0; idx < attrInfos.length; ++idx) {
             MBeanAttributeInfo attr = attrInfos[idx];
             if (!attr.isReadable()) {
                 logScrape(mbeanName, attr, "not readable");
                 continue;
             }
+            name2AttrInfo.put(attr.getName(), attr);
+        }
+        final AttributeList attributes;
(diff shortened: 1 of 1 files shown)


fstab:
Hi, I changed the code so that 1-by-1 loading of attributes is only used as a fallback if bulk loading fails.

I'm planning to include this in the next release.

Thanks a lot!

Issue #703, opened 2022-04-20
Default HELP strings produce high cardinality


SuperQ:
The default HELP text includes the full attribute information, which can be very high cardinality. For example, Kafka includes `client-id` [as attributes on the metric](https://github.com/prometheus/jmx_exporter/blob/c46cfda52447ecb341e04a6479dafca442f48717/example_configs/kafka-2_0_0.yml#L5).

> Defaults to the mBean attribute description and the full name of the attribute.

Including the full name exposes every label combo as a HELP string. In Prometheus and Thanos, this now bloats the [`/api/v1/metadata` endpoint](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata). This is used for things like type completion in the UI.

I would highly recommend changing the default help string to just the mBean attribute description.

PR #704, opened 2022-04-20
Reduce cardinality of default help strings

SuperQ:
Including the whole mBean fully qualified name can produce a huge amount
of HELP metadata cardinality, slowing down the UI and bloating the TSDB
indexes.

Fixes: https://github.com/prometheus/jmx_exporter/issues/703

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -451,8 +451,14 @@ public class JmxCollector extends Collector implements Collector.Describable {
           Object beanValue) {
 
         String beanName = domain + angleBrackets(beanProperties.toString()) + angleBrackets(attrKeys.toString());
-        // attrDescription tends not to be useful, so give the fully qualified name too.
-        String help = attrDescription + " (" + beanName + attrName + ")";
+
+        // Build the HELP string from the bean metadata.
+        String help = domain + ":name=" + beanProperties.get("name") + ",type=" + beanProperties.get("type") + ",attribute=" + attrName;
+        // Add the attrDescription to the HELP if it exists and is useful.
+        if (attrDescription != null && !attrDescription.equals(attrName)) {
+          help = attrDescription + " " + help;
+        }
+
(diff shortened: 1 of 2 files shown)


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -451,8 +451,8 @@ public void recordBean(
           Object beanValue) {
 
         String beanName = domain + angleBrackets(beanProperties.toString()) + angleBrackets(attrKeys.toString());
-        // attrDescription tends not to be useful, so give the fully qualified name too.
-        String help = attrDescription + " (" + beanName + attrName + ")";
+        // attrDescription tends not to be useful, so include domain and attrName.
+        String help = attrDescription + " (domain=" + domain + ",attrName=" attrName + ")";
```
fstab, line 455: What do you think of this?
```java
        String help = domain + ":name=" + beanProperties.get("name") + ",type=" + beanProperties.get("type") + ",attribute=" + attrName;
        if (attrDescription != null && !attrDescription.equals(attrName)) {
          help = attrDescription + " " + help;
        }
```
Having the domain, type, name, and attribute would allow users to uniquely identify where the data is coming from, while it still does not contain the `beanName` which was causing the cardinality explosion.

Example output:
```
# HELP java_lang_Metaspace_CollectionUsageThresholdSupported java.lang:name=Metaspace,type=MemoryPool,attribute=CollectionUsageThresholdSupported
# TYPE java_lang_Metaspace_CollectionUsageThresholdSupported untyped
java_lang_Metaspace_CollectionUsageThresholdSupported{type="MemoryPool",} 0.0
```

fstab (changes requested):

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
SuperQ: I like it. Let me know if those comments make sense.

fstab: :+1: thanks a lot!

PR #719, opened 2022-07-03
Improve performance of duplicate sample lookup

amuraru:
Maintain a lookup index for sample keys (name, labels)
and use that to check for duplicate sample during scraping
instead of O(n) list of samples

Also guard non-trivial computation behind Logger level check


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -335,9 +337,50 @@ private static boolean isLegalCharacter(char input) {
             (input >= '0' && input <= '9'));
   }
 
-    class Receiver implements JmxScraper.MBeanReceiver {
+    /**
+     * A sample is uniquely identified by its name, labelNames and labelValues
+     */
+    static class SampleKey {
```
fstab, line 343: Should be `private static class SampleKey`.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -335,9 +337,50 @@ private static boolean isLegalCharacter(char input) {
...
+      String name;
+      List<String> labelNames;
+      List<String> labelValues;
```
fstab, line 346: These should be `private final`.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -335,9 +337,50 @@ private static boolean isLegalCharacter(char input) {
...
+
+      SampleKey(String name, List<String> labelNames, List<String> labelValues) {
```
fstab, line 348: Can be `private` as well.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -335,9 +337,50 @@ private static boolean isLegalCharacter(char input) {
...
+        this.name = name;
+        this.labelNames = labelNames;
+        this.labelValues = labelValues;
+      }
+
+      static SampleKey of(MetricFamilySamples.Sample sample) {
```
fstab, line 354: Can be `private static`

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -335,9 +337,50 @@ private static boolean isLegalCharacter(char input) {
...
+      @Override
+      public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        SampleKey sampleKey = (SampleKey) o;
+
+        if (name != null ? !name.equals(sampleKey.name) : sampleKey.name != null) return false;
+        if (labelNames != null ? !labelNames.equals(sampleKey.labelNames) : sampleKey.labelNames != null) return false;
+        return labelValues != null ? labelValues.equals(sampleKey.labelValues) : sampleKey.labelValues == null;
```
fstab, line 367: A little micro-optimization: We could compare `labelValues` first and then `labelNames`, because `labelValues` are more likely to be different.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -335,9 +337,50 @@ private static boolean isLegalCharacter(char input) {
...
+        result = 31 * result + (labelValues != null ? labelValues.hashCode() : 0);
+        return result;
+      }
+
+    }
+
+    static class Receiver implements JmxScraper.MBeanReceiver {
       Map<String, MetricFamilySamples> metricFamilySamplesMap =
         new HashMap<String, MetricFamilySamples>();
+      Set<SampleKey> uniqueSampleKeys = new HashSet<SampleKey>();
```
fstab, line 383: Not sure about the name `uniqueSampleKeys`. As this is used to check if the sample keys exist, maybe call it `existingSampleKeys`?

fstab (changes requested):
Thanks a lot. Looks good to me, I just added a few minor remarks.

Please also remove the old `findExisting()` method as it's not used anymore.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
amuraru: good call.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
amuraru: good call.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
amuraru: good call.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
amuraru: good call.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
amuraru: good call.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
amuraru: renamed to `sampleKeys`

amuraru:
> Please also remove the old `findExisting()` method as it's not used anymore.

Good catch. Fixed.

amuraru:
@fstab - kind reminder to review this PR. thanks

dhoard:
@fstab I know you are busy, but I feel this should be prioritized. On a real Kafka broker, this change decreased collection time by ~40%.

Before the change...

```
     time_namelookup:  0.004212s
        time_connect:  0.004593s
     time_appconnect:  0.058974s
    time_pretransfer:  0.059085s
       time_redirect:  0.000000s
  time_starttransfer:  27.566052s
                     ----------
          time_total:  27.618721s
```

After the change...

```
     time_namelookup:  0.006547s
        time_connect:  0.007784s
     time_appconnect:  0.062182s
    time_pretransfer:  0.062339s
       time_redirect:  0.000000s
  time_starttransfer:  17.136378s
                     ----------
          time_total:  17.226715s
```

fstab:
Thanks a lot, and sorry for the long delay.

Issue #720, opened 2022-07-20
Add configuration for the minimum and maximum number of HTTP Server threads


atl-mk:
There's at least two problems this solves:

1. The output buffer of the HTTP threads is held by a strong reference so if there's a lot of metrics (on the order of 10,000s) the memory usage can be considerable (on the order of 100s MB which is much larger than the raw metric).
2. On smaller [hardware] machines if the exporter end point is being hammered, then the processor utilisation can also be significant (by default 5 up to hardware threads). Dropping metrics data in this case is favourable compared to the potentially instability of what's being monitored and it would help find the issue faster.

dhoard:
@atl-mk 

Regarding 1.

`client_java` https://github.com/prometheus/client_java/issues/703

The thought process around keeping the buffer around https://github.com/prometheus/client_java/issues/703#issuecomment-930470387

Regarding 2.

If the agent's metrics endpoint is being called too aggressively, then the calling system should be using a longer scraping interval.

atl-mk:
@dhoard absolutely, but that shouldn't bring down nor affect the performance of what is being monitored. Instead putting a hard and low limit allows finding that issue more quickly.

dhoard:
@atl-mk

WRT 1 - I'm in agreement, min/max threads should be configurable. The core `HTTPServer` class allows for this configuration when built using the `HTTPServer.Builder`. The JMX exporter has yet to implement this feature.

WRT 2 - I'm in agreement that the exporter shouldn't adversely affect the performance of the application being monitored. This can be difficult because the exporter has no control of JMX MBeans/metrics that are registered or how long it takes to scrape them.

You mentioned "dropping metrics" ... can you clarify?

atl-mk:
@dhoard Oh apologies, I was confused and misread that, I'm sorry.

Sorry, "dropping metrics" is not the really right, what I meant was; as the Prometheus server is polling there will be missing data points if the poll interval is shorter than the time it takes to get a response from the exporter. In my experience this has been pretty quickly identified by looking at Grafana dashboards.

dhoard:
@atl-mk I have added thread min, max, and keep-alive configurations to the HTTP server. The code is pre-release but well-tested.

See https://github.com/prometheus/jmx_exporter/issues/688

atl-mk:
Thanks @dhoard, looking forward to see this in main

dhoard:
@atl-mk This is targeted for the 0.18.0 release

[...]

PR #721, opened 2022-07-20
Add min and max thread configuration for the HTTP server

atl-mk:
To solve https://github.com/prometheus/jmx_exporter/issues/720

After some early feedback on the approach.

My thought is effectively:

- Extract and encapsulate configuration, including the loading of it. In the longer term it could be a nicety for the agent to be entirely configurable through the file, and this would be a step towards that.
- Either the `HTTPServer` takes some arguments for min and max threads, or the `NamedDaemonThreadFactory` becomes public

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/Config.java
+++ b/collector/src/main/java/io/prometheus/jmx/Config.java
@@ -1,0 +1,169 @@
+package io.prometheus.jmx;
+
+import io.prometheus.client.Collector;
+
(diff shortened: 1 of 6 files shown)


atl-mk:
@fstab and @tomwilkie I think this is ready for review

atl-mk:
Resolved via: https://github.com/prometheus/jmx_exporter/pull/837/files

PR #724, opened 2022-08-01
feat: adding apache helix as an example

wmorgan6796:
I created this configuration to allow for our helix deployment to report metrics to prometheus. Figured it would be helpful for other folks as getting it to work was non-trivial.


dhoard (approved):

dhoard:
@wmorgan6796 thanks for the contribution!

Issue #688, opened 2022-02-25
SSL configuration does not enable metrics being exposed through HTTPS


csrrmrvll:
I am trying to expose metrics via HTTPS or SSL on TCP and being scraped from a central Prometheus as there are firewall rules in place to block non-cyphered connections in our network. I have configured jmx_exporter as documented, but I can only access metrics over HTTP. This is the config:

#### version
jmx_prometheus_javaagent-0.16.1

---
#### config.yaml:
ssl: true
lowercaseOutputLabelNames: false
lowercaseOutputName: false
startDelaySeconds: 0

---
#### JAVA_OPTS
-javaagent:/opt/jmx-exporter/jmx_prometheus_javaagent-0.16.1=10001:/etc/jmx-exporter/jmx_prometheus_javaagent.yaml -Djavax.net.ssl.keyStore=/etc/jmx-exporter/keystore -Djavax.net.ssl.keyStorePassword=changeit -Djavax.net.ssl.trustStore=/etc/jmx-exporter/truststore -Djavax.net.ssl.trustStorePassword=changeit

---
What am I missing?

dhoard:
`jmx_exporter` currently does not support HTTPS for exposing metrics.

csrrmrvll:
Thanks @dhoard.
Do you know if there are plans to provide this support in the short term?

dhoard:
@csrrmrvll All the underlying pieces are in place. @fstab is working on configuration changes that would allow `jmx_exporter` to support both SSL/TLS termination and basic authentication.

srinirama:
@fstab - is the SSL configuration to allow jmx_exporter to support both SSL/TLS termination and basic authentication being scoped ? We are keeping fingers crossed for this feature

palashbhowmick:
We are also facing the same problem. We can't use JMX Exporter because it doesn't support SSL yet.
Any ETA? @dhoard @fstab 
Any work around?

fstab:
Here's a status update:
:heavy_check_mark: first implementation of a new config file format is done, it's in the `new-config` branch. This format will allow to configure SSL for scraping via JMX as well as for exporting via HTTPS.
:heavy_check_mark: first integration test for scraping with SSL via JMX is implemented and pushed to the `new-config` branch.
:black_square_button: The new config parser should be backwards compatible, we need tests to make sure current config files still work.
:black_square_button: We need more integration tests for different SSL scenarios.

Here's a bit of brainstorming what tests are missing:

[...]

PR #736, opened 2022-09-13
add simple auth for httpserver

mvpydq:
although there is auth support for collector, but httpserver is not use this auth, and I add a simple SimpleAuthenticator which will check auth which has been configured in yml with item name "username" and "password"

Signed-off-by: dqyang <dqyang@freewheel.tv>


mlhmz:
See #688, this feature and also SSL/TLS is already planned in the new-config Branch. You should fork this repo and implement the changes there. Also, are you sure that your changes work? You've just created a class which extends BasicAuthenticator, you need to implement that change too.

dhoard:
Implemented in `0.19.0`.

PR #746, opened 2022-11-07
Handled exception for non-standard attribute types in bean scraping

viruzzo:
**_`JmxScraper.scrapeBean()`_** currently fails with an `IllegalArgumentException` when calling _`AttributeList.asList()`_ if the list contains any non-`Attribute` attributes (which is permitted by the API, though heavily discouraged).

The proposed change handles the exception and logs the name of the offending MBean.

@fstab @tomwilkie


collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -160,9 +160,15 @@ private void scrapeBean(MBeanServerConnection beanConn, ObjectName mbeanName) {
             processAttributesOneByOne(beanConn, mbeanName, name2AttrInfo);
             return;
         }
-        for (Object attributeObj : attributes.asList()) {
-            if (Attribute.class.isInstance(attributeObj)) {
-                Attribute attribute = (Attribute)(attributeObj);
+        final List<Attribute> attributeList;
+        try {
+            attributeList = attributes.asList();
```
dhoard, line 165: The approach proposed is better than the current code, but still uses an "all-or-nothing" paradigm for a given MBean.

I feel a better approach would be to loop through the values in the `AttributeList`. If the Object is an instance of `Attribute`, process it... if it's not an `Attribute` log a message via a `logScrape` and ignore the object.

dhoard (changes requested):

dhoard:
@viruzzo Are still interested in finishing the PR?

dhoard:
Closing. Implemented as part of https://github.com/prometheus/jmx_exporter/pull/843

Issue #759, opened 2022-12-17
Jmx_exporter can cause program to hang if it has a lot of threads


Selikoff:
Version:  jmx_exporter 0.17.2

I noticed if an application has too many threads (15k or more), the `jmx_exporter` can cause a program to hang.   It'll hang the main thread however it takes to finish the jmx_exporter process (10+ seconds in my tests).  I wrote a simple script that can reproduce the issue:

```
public static void main(String[] args) throws Exception {
	final int count = 15_000;
	final Thread[] thread = new Thread[count];
	for(int i=0; i<thread.length; i++) {
		thread[i] = new Thread(() -> {
			while(true) {
				try {
					Thread.sleep(500);
				} catch (Exception e) {}
			}
		});
		thread[i].start();
	}

	while(true) {
		System.out.println("[time="+System.currentTimeMillis()+"]");
		Thread.sleep(100);
	}
public static void main(String[] args) throws Exception {
```

Basically if you run this with the jmx_exporter and call `curl http://localhost:123` in the background, it'll freeze the main thread intermittently (about 30% of the time).  You might have to adjust some of the timings for it to appear.

I traced the source of the delay to this class [ThreadExports.java](https://github.com/prometheus/client_java/blob/main/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java) class, lines 110-124.  There is a filter, if enabled, would disable `JVM_THREADS_STATE` / `jvm_threads_state`.  Enabling this filter prevents the issue from happening.

The problem, and the reason I'm reporting this as an issue, is there's no way to disable just the `jvm_threads_state` process in jmx_exporter.   All of the `rules` in the config gets executed *after* the collectors run, not before.  I believe the fix would be to pass down information to the [HTTPSerer.java](https://github.com/prometheus/client_java/blob/main/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java) class, which calls `metricFamilySamples()` instead of `filteredMetricFamilySamples()`.

Note:  It is possible to disable JVM metric in the curl call to the server, aka `curl http://localhost:123?name[]=my_metric` but this is extremely limited.  In particular, you have to select metrics by name.  You can't use regex or negation.  Put another way, if you have 3,000 metrics and you want to filter out 1, you would have to list 2,999 using this technique.  Ideally, the solution should be part of the jmx_exporter config.

Selikoff:
[...]

PR #760, opened 2022-12-17
Add support for collectorNamePattern in config and JavaAgent

Selikoff:
Per [Issue 759](https://github.com/prometheus/jmx_exporter/issues/759), there is no way to prevent certain metrics from being processed, such as `jvm_threads_state`, as this is run inside `CollectionRegistry` before the jmx_exporter code can apply any rules.  For applications with a lot of threads, this can temporarily cause the application to hang even if the metrics are later filtered out by a rule.

The solution is to add a new pattern value to the config that can be sent into `JavaAgent` and then passed to the `HTTPServer` as a `SampleNameFilter`.  This allows metrics to be skipped from processing and resolves issue #759 by adding a config such as:

```
collectorNamePattern : "^(?!jvm_threads_state$).*$"            
rules:
  - pattern: ".*"
  ```

### CODE REVIEW

File: jmx_prometheus_javaagent_common/src/main/java/io/prometheus/jmx/JavaAgent.java [hunk 1]
```diff
@@ -6,7 +6,10 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import com.sun.net.httpserver.HttpServer;
```
fstab, line 9: unused import

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -207,6 +208,10 @@ private Config loadConfig(Map<String, Object> yamlConfig) throws MalformedObject
           }
         }
 
+        if (yamlConfig.containsKey("collectorNamePattern")) {
+            cfg.collectorNamePattern = (String)yamlConfig.get("collectorNamePattern");
```
fstab, line 212: Please validate if collectorNamePattern is a valid regex, so that `jmx_exporter` fails immediately on config error. Otherwise it starts successfully but fails when the Prometheus server scrapes.

fstab (changes requested):
Awesome, thanks a lot for your PR. I added two minor comments, apart from that it's looking good.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
Selikoff: Added validation check as well as unit test to verify coverage.

jmx_prometheus_javaagent_common/src/main/java/io/prometheus/jmx/JavaAgent.java
Selikoff: Fixed

Selikoff:
@fstab Can we merge this CL?  I believe I addressed your concerns.

dhoard:
@Selikoff looking at the PR... I think the changes should also be made to the standalone HTTP JMX exporter as well.

https://github.com/prometheus/jmx_exporter/blob/main/jmx_prometheus_httpserver_common/src/main/java/io/prometheus/jmx/WebServer.java

dhoard:
Looking at the code more, based on...

https://github.com/prometheus/client_java/blob/9ba0c2a9376986b6b47e031e04b277e9a93c9431/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L167

... `collectorNameFilter` is an inaccurate description and should be switched to `sampleNameFilter` since the code doesn't actual filter the collector, but samples from the collector.

dhoard:
@Selikoff are you still interested in completing this PR?

dhoard:
@Selikoff following up on the PR. are you still interested in completing this PR?

dhoard:
@Selikoff following up on the PR. Are you still interested in completing this PR?

Issue #766, opened 2023-02-01
CVE issue


sreejesh-radhakrishnan-db:
Hi

on jar jmx_prometheus_javaagent-0.17.2.jar  there is a critical CVE-2022-1471 related to jmx_prometheus_javaagent-0.17.2.jar, will you be rebuilding jmx_prometheus_javaagent with 1.33 version of snakeyaml at all?

CVE-2022-1471:

A design problem in SnakeYAML leads to remote code execution when deserializing untrusted YAML data. 

Security  affected version org.yaml:snakeyaml:1.32    fix version 1.33

dhoard:
I suspect it will be upgraded in the next release.

Keep in mind that this CVE is controversial because it is related to **untrusted YAML**.

The exporter YAML is considered **trusted** because it's provided by you when you install the exporter.

fstab:
Fixed with Snakeyaml 2.0.

PR #767, opened 2023-02-04
Fix and tests for CVE-2022-1471

dhoard:
Fix and tests for CVE-2022-1471 (https://github.com/prometheus/jmx_exporter/issues/766)

Implemented a filtering SafeConstructor per https://github.com/advisories/GHSA-mjmj-j48q-9wg2

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/FilteringSafeConstructor.java
+++ b/collector/src/main/java/io/prometheus/jmx/FilteringSafeConstructor.java
@@ -1,0 +1,48 @@
+package io.prometheus.jmx;
+
+import org.yaml.snakeyaml.constructor.Constructor;
+
+import java.math.BigInteger;
+import java.sql.Timestamp;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class FilteringSafeConstructor extends Constructor {
+
+    private Set<String> allowedClassNameSet;
+
+    public FilteringSafeConstructor() {
+        super();
+
+        allowedClassNameSet = new HashSet<String>();
+        allowedClassNameSet.add(null);
+        allowedClassNameSet.add(BigInteger.class.getName());
+        allowedClassNameSet.add(Boolean.class.getName());
+        allowedClassNameSet.add(Byte.class.getName());
+        allowedClassNameSet.add(Character.class.getName());
+        allowedClassNameSet.add(Date.class.getName());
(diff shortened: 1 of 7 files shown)


dhoard:
@fstab please review.

fstab:
Hi @dhoard, sorry for the delay. I'm wondering why you closed it? Is it worthwhile to take a look, or should we just update the snakeyaml dependency as soon as an update is available?

dhoard:
@fstab my apologies. It was closed by mistake. I restored the branch.

An upgraded Snake YAML won't resolve the issue... the Snake YAML maintainers feel it's a false positive.

The code in the branch defines and uses a new `FilteringSafeConstructor` which constrains Snake YAML.

fstab:
Thanks for you reply. I had a closer look, and I'm not sure whether fixing snakeyaml bugs in jmx_exporter is the right approach.

The main issue is that security scanners find a dependency with a CVE. This annoys jmx_exporter users, they want to get rid of these alerts. This cannot be solved by wrapping the snakeyaml constructor, security scanners will still find the affected dependency and annoy our users.

Is there an actual security risk? I don't think so. In order to exploit the snakeyaml CVE, an attacker needs to be able to modify jmx_exporter's configuration file. If an attacker has write access to jmx_exporter's configuration thy system is already compromised anyway.

So I'm thinking whether it makes sense to contribute the fix upstream to snakeyaml? If they accept it and build a new release, the CVE warnings will be gone. So far I have the impression that snakeyaml CVEs do get fixed. What do you think?

dhoard:
I'm undecided.

Upstream Snake YAML states that configuration YAML is considered trusted, so it's CVE detection issue and not a bug.

But, if I have a deployment scenario where the YAML configuration is external (dynamic retrieval) and that configuration has been compromised, then the change would help with safety.

Also, In my experience, enterprise customers will be block usage because the CVE scan will fail.

Regardless of whether the PR is merged, etc. it's probably worth a statement in the README or a FAQ page.

I have seen scenarios where these type of issues keep getting reported time and time again.

joseph-skupniewicz-sp:
Yes- can you please update the FAQ or readme until this is fixed upstream?

In my organization, we are regularly testing for vulnerabilities/CVEs and this one is flagged as it is being bundled with one of our services, but it shouldn't affect us since there is no write access to the jmx_exporter configuration- being in a container in a private subnet and not exposed any other way.

It would help clarify the presence of this vulnerability as a non-issue for our whitelisting.

Issue #728, opened 2022-08-10
Version 0.17.0 is returning unknown version and name info


kaiquerass:
Docker image version 0.17.0 is returning `jmx_exporter_build_info{version="unknown",name="unknown",} 1.0`, which is breaking our scrape to this metrics.

Expected behavior: 
The same result of 0.16.1 version image `jmx_exporter_build_info{version="0.16.1",name="Prometheus JMX Exporter - Http Server",} 1.`

dhoard:
@kaiquerass @LHozzan I created a PR to resolve the issue.

However, the name will represent the artifact name `jmx_prometheus_httpserver` and not the project name `Prometheus JMX Exporter - Http Server` (The name change was apparently made post `0.0.16`... not sure why)

devopology:
@kaiquerass @LHozzan please verify against `0.18.0` and close if resolved.

PR #768, opened 2023-02-08
Fixed packaging to add version and name to build info metric

dhoard:
Fixed packaging to add version and name to build info metric for the HTTP server. (The Agent builds work correctly.)
Added integration tests to validate that the name is correct and that the version is not "unknown" for all builds.

https://github.com/prometheus/jmx_exporter/issues/728


fstab:
Thanks a lot!

PR #770, opened 2023-02-16
fix(examples): add missing kafka quota pattern

doxsch:
Quotas can be specified in tupples (user, client-id). The user and client-id can also be used isolated.

So added the missing rule.

@fstab
@tomwilkie


fstab:
Thanks @doxsch!

PR #775, opened 2023-03-02
Bump versions of source and target for maven compiler plugin.

psoo:
1.6 is not supported anymore, so bump up to 1.7. Passes all tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -237,8 +237,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.7</source>
+                    <target>1.7</target>
                 </configuration>
             </plugin>
         </plugins>


dhoard:
The project currently considers Java 6 as supported. Changing this breaks the Java 6 versions of the jars.

`mvn clean package install` ... which runs the integration tests, fails.

Java 1.6 is supported by Azul Systems until 2027 for customers that have purchased Production Support for Java 6.

https://www.azul.com/products/azul-support-roadmap/

psoo:
Hi,

I already feared there isn't an easy solution to this...

I was investigation an error on a CI pipeline here and stumpled across this when compiling the `jmx_exporter` there:

````
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project collector: Compilation failure: Compilation failure: 
[ERROR] Source option 6 is no longer supported. Use 7 or later.
[ERROR] Target option 6 is no longer supported. Use 7 or later
````

This runs on a recent Ubuntu LTS 22.04 and i couldn't get another solution to get it succeeding than in the proposed patch. 

If we still support Java 6, we need another solution.

Issue #776, opened 2023-03-02
CVE-2022-1471 fixed in snakeyaml 2.0 ... any plans to adopt it?


ppatierno:
Hey all,
following the #766 and corresponding closed (not merged) PR #767, it seems that finally snakeyaml community fixed CVE-2022-1471 in 2.0 release. See https://bitbucket.org/snakeyaml/snakeyaml/wiki/Changes
Is there any plan to update JMX exporter to the new 2.0 release to fix the problem and avoid security scanning raising the issue again?

dhoard:
@ppatierno the fix will work for the Java 8-based versions...

`jmx_prometheus_javaagent-X.jar`
`jmx_prometheus_httpserver-X.jar`

... but won't address the CVE issue for the Java 6-based versions...

`jmx_prometheus_javaagent_java6-X.jar`
`jmx_prometheus_httpserver_java6-X.jar`

... which uses SnakeYaml 1.23 (last version for Java 6 support)

It will take some investigation on how to make a workaround for Java 6... if the CVE chosen to be resolved in Java 6 versions.

ppatierno:
> the fix will work for the Java 8-based versions...

we use the agent on Java 11 and 17 as runtime in a couple of projects and directly the `JmxCollector` class in another (on Java 11), so it should be fine.
But it's not clear from your comment if you are going to have a plan about updating to 2.0 to fix it.

dhoard:
I feel that the Java 8 versions should be upgraded to snakeyaml 2.0 to mitigate the CVE (all tests pass)... and not address the CVE for Java 6 versions.

I would open a PR and tag the maintainers for a review.

ppatierno:
I saw you already opened the PR, I was going to do that this morning. Of course, thank you very much. Let's see how it will go with the maintainers review.

fstab:
Thanks a lot everyone! Update to Snakeyaml 2.0 is  pushed, for discussion on Java 6 see #777.

PR #777, opened 2023-03-02
Fix for CVE-2022-1471

dhoard:
This is a fix for CVE-2022-1471.

Due to changes in the SnakeYAML 2.0 `Constructor` between 2.0 and 1.23 (the last version to support Java 6), I had to introduce a wrapper class `FilteringSafeConstructor` which I could then implement specifically for Java 6 artifacts.

https://github.com/prometheus/jmx_exporter/issues/776

### DIFF
--- a/collector/pom.xml
+++ b/collector/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
-            <version>1.33</version> <!-- The Java 6 build overrides this with 1.23. -->
+            <version>2.0</version> <!-- The Java 6 build overrides this with 1.23. -->
         </dependency>
     </dependencies>
 
--- a/collector/src/main/java/io/prometheus/jmx/FilteringSafeConstructor.java
+++ b/collector/src/main/java/io/prometheus/jmx/FilteringSafeConstructor.java
@@ -1,0 +1,19 @@
+package io.prometheus.jmx;
+
+import org.yaml.snakeyaml.LoaderOptions;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
+
+/**
+ * Class to implement an extension to SafeConstructor to handle SnakeYAML version differences
+ *
+ * This is a SnakeYAML version 2.0.0 which provides safe filtering
+ *
+ * Java 6 versions of this class implement safe filtering of entity types
+ */
+public class FilteringSafeConstructor extends SafeConstructor {
+
+    public FilteringSafeConstructor() {
+        super(new LoaderOptions());
+    }
+}
+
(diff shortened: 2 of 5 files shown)


ppatierno (approved):
Not a maintainer but the changes look good to me.
Hopefully this will be approved and merged soon for a new release.

fstab:
Thanks a lot for the PR! The PR includes two changes:

1. Update Snakeyaml to version 2.0, which fixes the CVE.
2. Implement a wrapper that prevents the CVE for older Snakeyaml versions, as the Java 6 build still uses Snakeyaml 1.23.

Updating Snakeyaml is certainly great, I just pushed a version update. This will fix the CVE except for the Java 6 build.

For the second point I'm not sure if it's worthwhile:

* It will not fix warnings from security scanners for the Java 6 build, as the Java 6 library will still have Snakeyaml 1.23 as a dependency.
* We will drop Java 6 support with the implementation of native histograms in `client_java`. Once we have Java 8 as baseline we need to remember to remove the `FilteringSafeConstructor` classes because they'll no longer be needed.
* Snakeyaml 1.23 has other CVEs as well, so fixing one of them still leaves other vulnerabilities.
* All Snakeyaml CVEs require modifications to the YAML file, so they can be exploited only if an attacker can modify jmx_exporter's YAML config.

I think the tradeoff here should be to just bump the Snakeyaml version and ignore Java 6.

fstab:
Thanks a lot for the PR! The PR includes two changes:

1. Update Snakeyaml to version 2.0, which fixes the CVE.
2. Implement a wrapper that prevents the CVE for older Snakeyaml versions, as the Java 6 build still uses Snakeyaml 1.23.

Updating Snakeyaml is certainly great, I just pushed a version update. This will fix the CVE except for the Java 6 build.

For the second point I'm not sure if it's worthwhile:

* It will not fix warnings from security scanners for the Java 6 build, as the Java 6 library will still have Snakeyaml 1.23 as a dependency.
* We will drop Java 6 support with the implementation of native histograms in `client_java`. Once we have Java 8 as baseline we need to remember to remove the `FilteringSafeConstructor` classes because they'll no longer be needed.
* Snakeyaml 1.23 has other CVEs as well, so fixing one of them still leaves other vulnerabilities.
* All Snakeyaml CVEs require modifications to the YAML file, so they can be exploited only if an attacker can modify jmx_exporter's YAML config.

I think the tradeoff here should be to just bump the Snakeyaml version and ignore Java 6.

fstab:
Sorry, I accidentally closed it but wanted to keep it open for a while in case this needs further discussion.

fstab:
Today or tomorrow.

ppatierno:
@fstab thanks for that!

fstab:
0.18.0 released :tada:

dhoard:
Resolved in 0.18.0 for Java 8 version. No plans to address for Java 6 versions.

PR #780, opened 2023-03-16
Refactored integration test suite

dhoard:
@fstab please review.

This PR represents a refactored integration test suite using the current code base. (No new JMX exporter features.)

The integration tests were written using the AntuBLUE Test Engine (https://github.com/antublue/test-engine) which allows test class-level parameterization/multi-threaded execution/best use of resources.

(disclaimer - AntuBLUE Test Engine is a project I have been working on due to https://github.com/junit-team/junit5/issues/878)

I have left test classes/test configuration for future HTTP BASIC authentication and HTTPS support work (they may need to be changed depending on the configuration format), but they are disabled (@TestEngine.Disabled annotation)

The default Docker image names for smoke tests are:

- azul/zulu-openjdk:8
- azul/zulu-openjdk:11
- azul/zulu-openjdk:17
- azul/zulu-openjdk:19
- azul/zulu-openjdk:20
- openjdk:6
- ticketfly/java:6

Signed-off-by: dhoard <doug.hoard@gmail.com>


integration_test_suite/integration_tests/pom.xml
```diff
@@ -0,0 +1,231 @@
...
+            <id>test-engine-maven-plugin</id>
+            <url>https://repository.antublue.org/test-engine-maven-plugin</url>
+            <releases>
+                <enabled>true</enabled>
+            </releases>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+        </pluginRepository>
+    </pluginRepositories>
```
fstab, line 72: Hi Doug, I failed to run `./mvnw clean verify` locally, because the antublue dependencies cannot be retrieved:
```
Plugin org.antublue:test-engine-maven-plugin:0.0.3 or one of its dependencies could not be resolved
```

I'm wondering if the plugin is really needed? Is it used for parameterized tests? The current integration tests use  `Parameterized`, which is standard JUnit 5 functionality that does not require other dependencies.

https://github.com/prometheus/jmx_exporter/blob/9ad31dac2d0c6d49d746efd6f5380dab6c78850b/integration_tests/smoke_tests/src/test/java/io/prometheus/jmx/AgentJavaVersionsIT.java#L26

What do you think about using JUnit instead of Antublue?

dhoard: Hi Fabian, can you please re-test using `./mvnw -U clean verify` (... there was a CloudFlare/repository issue.)

`Parameterized` and `@RunWith` included with JUnit 5 are really JUnit 4. JUnit 5 has a feature gap testing parameterized test classes: https://github.com/junit-team/junit5/issues/878

I can refactor to use JUnit 4, but I feel the AntuBLUE Test Engine provides a lot of benefits...

- better console test output
- parallel execution by default (console & IntelliJ)
- test class include/exclude filters
- test method include/exclude filters
- test class tag include/exclude filters

... all of which make development/testing much easier.

dhoard: Example AntuBLUE Test Engine output...

```
...
2023-03-28 | 12:49:58.101 | test-engine-08 | INFO | ticketfly/java:6 | Container ticketfly/java:6 started in PT0.387857S
2023-03-28 | 12:50:05.108 | test-engine-08 | INFO | ticketfly/java:6 | (TEST) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testHealthy()
2023-03-28 | 12:50:05.110 | test-engine-08 | INFO | ticketfly/java:6 | (PASS) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testHealthy()
2023-03-28 | 12:50:05.110 | test-engine-08 | INFO | ticketfly/java:6 | (TEST) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testMetrics()
2023-03-28 | 12:50:05.288 | test-engine-08 | INFO | ticketfly/java:6 | (PASS) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testMetrics()
2023-03-28 | 12:50:05.288 | test-engine-08 | INFO | ticketfly/java:6 | (TEST) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testMetricsOpenMetricsFormat()
2023-03-28 | 12:50:05.349 | test-engine-08 | INFO | ticketfly/java:6 | (PASS) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testMetricsOpenMetricsFormat()
2023-03-28 | 12:50:05.349 | test-engine-08 | INFO | ticketfly/java:6 | (TEST) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testMetricsPrometheusFormat()
2023-03-28 | 12:50:05.398 | test-engine-08 | INFO | ticketfly/java:6 | (PASS) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT testMetricsPrometheusFormat()
2023-03-28 | 12:50:06.191 | test-engine-08 | INFO | ticketfly/java:6 | (PASS) io.prometheus.jmx.test.httpserver.general.StartDelaySeconds_IT
2023-03-28 | 12:50:06.574 | main | INFO | -----------------------------------
2023-03-28 | 12:50:06.574 | main | INFO | AntuBLUE Test Engine v2.0.5 Summary
2023-03-28 | 12:50:06.574 | main | INFO | -----------------------------------
2023-03-28 | 12:50:06.574 | main | INFO | 
2023-03-28 | 12:50:06.574 | main | INFO | TESTS : 595, PASSED : 595, FAILED : 0, SKIPPED : 0
2023-03-28 | 12:50:06.574 | main | INFO | 
2023-03-28 | 12:50:06.574 | main | INFO | -----------------------------------
2023-03-28 | 12:50:06.574 | main | INFO | PASSED
2023-03-28 | 12:50:06.574 | main | INFO | -----------------------------------
2023-03-28 | 12:50:06.574 | main | INFO | Total Time  : 1 minute, 34 seconds, 979 ms
2023-03-28 | 12:50:06.576 | main | INFO | Finished At : 2023-03-28T12:50:06.576-04:00
2023-03-28 | 12:50:06.576 | main | INFO | -----------------------------------
```

jmx_prometheus_httpserver_common/src/main/java/io/prometheus/jmx/WebServer.java
```diff
@@ -29,5 +35,12 @@ public static void main(String[] args) throws Exception {
      new BuildInfoCollector().register();
      new JmxCollector(new File(args[1]), JmxCollector.Mode.STANDALONE).register();
      new HTTPServer(socket, CollectorRegistry.defaultRegistry);
+
+     System.out.println(
+       String.format("%s | %s | INFO | %s | %s",
+         SIMPLE_DATE_FORMAT.format(new Date()),
+         Thread.currentThread().getName(),
+         WebServer.class.getName(),
+         "Running"));
```
fstab, line 44: Adding a message to the `WebServer` is not as critical as adding a message to the agent. However, if you remove it from the agent you should remove it here too for consistency.

jmx_prometheus_javaagent_common/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -29,6 +35,13 @@ public static void premain(String agentArgument, Instrumentation instrumentation
             new JmxCollector(new File(config.file), JmxCollector.Mode.AGENT).register();
             DefaultExports.initialize();
             server = new HTTPServer(config.socket, CollectorRegistry.defaultRegistry, true);
+
+            System.out.println(
+                    String.format("%s | %s | INFO | %s | %s",
+                            SIMPLE_DATE_FORMAT.format(new Date()),
+                            Thread.currentThread().getName(),
+                            JavaAgent.class.getName(),
+                            "Running"));
```
fstab, line 44: I don't think the agent should print anything because this will end up in the output of the instrumented application. Adding unexpected messages to an application's output might break things.

I guess you added it so that you can use it as a wait strategy in the integration test:
```
Wait.forLogMessage(".*Running.*", 1)
```

It would be better to remove the message and use a different wait strategy, like waiting for the exporter port to become reachable.

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
```diff
@@ -0,0 +1,210 @@
...
+                .withClasspathResourceMapping("common", "/temp", BindMode.READ_ONLY)
+                .withClasspathResourceMapping(getClass().getName().replace(".", "/"), "/temp", BindMode.READ_ONLY)
+                .withCommand("/bin/sh application.sh")
+                .withExposedPorts(9999)
+                .withLogConsumer(outputFrame -> System.out.print(outputFrame.getUtf8String()))
+                .withNetwork(network)
+                .withNetworkAliases("application")
+                .withStartupTimeout(Duration.ofMillis(30000))
+                .withStartupCheckStrategy(new IsRunningStartupCheckStrategy())
+                .withWorkingDirectory("/temp");
```
fstab, line 86: On my machine (arch linux), I need to increase the number of file descriptors otherwise `firefly:java6` won't start. I guess the default number of file descriptors has changed recently, as this wasn't necessary a couple of months ago.

Anyway, there's a quick fix, just add this:
```
// The firefly/java:6 container needs an increased number of file descriptors, so set the nofile ulimit here.
.withCreateContainerCmdModifier(c -> c.getHostConfig().withUlimits(new Ulimit[]{new Ulimit("nofile", 65536L, 65536L)}))
```

However, when I added this manually to run the tests on my machine, I realized that there's a lot of copy-and-paste. The `applicationContainer` and `exporterContainer` configuration code is copied to 61 files.

It would be good to refactor the tests and move copy-and-paste code to a common helper class to avoid code duplication.

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
```diff
@@ -0,0 +1,210 @@
...
+        try (Response response = httpClient.execute(requestBuilder)) {
+            assertThat(response).isNotNull();
+            assertThat(response.code()).isEqualTo(200);
+            ResponseBody responseBody = response.body();
+            assertThat(responseBody).isNotNull();
+            String content = responseBody.string();
+            assertThat(content).isNotNull();
+            assertThat(content).isEqualTo("Exporter is Healthy.");
+        }
+    }
```
fstab, line 130: This test (and others below) are copied to all test files. It would be good if the test was implemented only once, and then run with different parameters.

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @TestEngine.Test
```
fstab, line 132: Using this random code line to discuss antublue vs JUnit: Glad to hear that antublue is also available on Maven central, not depending on the antublue repository will make the build a lot more stable.

I don't want to be too picky with test dependencies. With runtime dependencies we should be very restrictive, as they might cause trouble in instrumented applications. Test dependencies are not critical.

However, I found one more disadvantage of antublue compared to JUnit: I'm using Intellij (as many Java developers), and Intellij doesn't seem to recognize antublue tests. If I click the little green play button for a test it just says "No tests were found". Being able to run a test directly in Intellij is a strong argument for JUnit. What do you think?

fstab (changes requested):

fstab:
Thanks a lot Doug! I added a few more comments. It's a lot of code, I guess after we discussed the current comment's I'll need another round of review, but I'm getting there :).

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
dhoard: I'll add the change.

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
dhoard: Definitely can refactor... though there are issues with inheritance using IntelliJ.

Example:

```
public abstract class BaseTest {

    @Test
    protected void testA() {
        // test code
    }

}
```

```
public class ConcreteTest extends BaseTest {
    // no methods, we want to use the methods in BaseTest
}
```

There are no members to run from IntelliJ in the **Project** view. The class (classname) is really a _marker_ for the configuration to use the base class.

Switch to the **Structure** view in IntelliJ for the class and enable the view to show inherited methods

image

The method is now listed and has a right-click **Run** option... but it doesn't run because it's a method in the abstract class `BaseTest`

I debugged the JUnit `engineDiscoveryRequest` which confirms IntelliJ is trying to run the method against the abstract base class (This is regardless of the test engine being used)

image

If the `BaseTest` is not abstract, it will run... but will run in the context of the `BaseTest` and not ` ConcreteTest` defeating the intent of trying to run the tests in `ConcreteTest`

---

Another approach...

Example:

```
public abstract class BaseTest {

    protected void testA() {
        // test code
    }

}
```

```
public class ConcreteTest extends BaseTest {

    @Test
    public void testA() {
        super.testA();
}
```

This will work and is less code in the test classes, but results in `BaseTest` being a "God Class" for all tests (or a set of God classes) which is an anti-pattern.

It would allow something like...

public abstract class BaseTest {

    protected void testHealthy(String username, String password, int expectedResponseCode) {
        // test code
    }

}

```
public class MyTest extends BaseTest {

    @Test
    public void testHealthy() {
        super.testHealty(String username, String password, int expectedResponseCode);
}
```

... or using functional interfaces...

```
public class UsernamePassword {

    public static Optional<UsernamePassword> of(String username, String password) {
        return Optional.of(new UsernamePassword(username, password));
    }
```

```
public abstract class BaseTest {

    protected void testHealthy(Optional<UsernamePassword> optional, Consumer<Integer> consumer) {
        // test code
    }

}
```

```
public class MyTest extends BaseTest {

    @Test
    public void testHealthy() {
        super.testHealty(
             UsernamePassword.of("prometheus", "test"),
             responseCode -> assertThat(responseCode).isEqualTo(200));
}
```

(Use of functional classes/interfaces is my preference between two above)

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
dhoard: One issue with using JUnit 4 is there are no annotations for something to be run before all tests / after all tests that are non-static.

`@BeforeClass` is/needs to be static to set up the Docker network once for performance reasons

`@Before` is run before every test.

Assuming the tests methods are run on the same thread, a `ThreadLocal` in the `@BeforeClass` could be used to set up state for `@Test` methods

---

The latest version (v2.0.6) of the AntuBLUE Test Engine supports execution via IntelliJ. (Should be on Maven Central within a few hours.)

dhoard:
> Thanks a lot Doug! These tests will definitely add value to the project, and make it a lot safer to change things in the future. I added a few more comments. It's a lot of code, I guess after we discussed the current comment's I'll need another round of review, but I'm getting there :).

Definitely has a lot of code. Sometimes the test code is bigger than the code being tested :smiley:

jmx_prometheus_javaagent_common/src/main/java/io/prometheus/jmx/JavaAgent.java
dhoard: Made the change

jmx_prometheus_httpserver_common/src/main/java/io/prometheus/jmx/WebServer.java
dhoard: Made the change to the Java agent

dhoard: Made the change to the Java agent

dhoard:
As potential refactoring approach, I changed `io.prometheus.jmx.test.httpserver.general.Minimal_IT` and `io.prometheus.jmx.test.httpserver.general.BlackListObjectNames_IT` to simplify the most common setup/teardown code.

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
fstab: There's no need to use inheritance. What do you think of a helper class with static methods like this:
```java
public class TestUtil {
    public static void runHealthTest(HttpClient httpClient, String username, String password) throws Exception {
        String path = "/-/healthy";
        Request.Builder requestBuilder = httpClient.createRequest(path);
        requestBuilder.addHeader(HttpHeader.AUTHORIZATION, basicAuthentication(username, password));
        try (Response response = httpClient.execute(requestBuilder)) {
            assertThat(response).isNotNull();
            assertThat(response.code()).isEqualTo(200);
            ResponseBody responseBody = response.body();
            assertThat(responseBody).isNotNull();
            String content = responseBody.string();
            assertThat(content).isNotNull();
            assertThat(content).isEqualTo("Exporter is Healthy.");
        }
    }
    // same for other code that's reused in multiple tests
}
```

Now you can call this method in the tests.
```java
    @TestEngine.Test
    public void testHealthy() throws Exception {
        TestUtil.runHealthTest(httpClient, USERNAME, PASSWORD);
    }
```

dhoard: This may be the best solution.

Given...

```
public interface Credentials {

    public void apply(Request.Builder requestBuilder) {
    }
}
```

```
public class BasicAuthenticationCredentials implements Credentials {
    public BasicAuthenticationCredentials(String username, String password) {
    }

    public void apply(Request.Builder requestBuilder) {
    }
}
```

We could test using valid credentials...

```
TestUtil.testHealthy(httpClient, new BasicAuthenticationCredentials(USERNAME, PASSWORD)).expect(200);
```

We could test without credentials...

```
TestUtil.testHealthy(httpClient, null).expect(401);
```

We could test using bad credentials...

```
TestUtil.testHealthy(httpClient, new BasicAuthenticationCredentials(USERNAME, INVALID_PASSWORD).expect(401);
```

---

This pattern would even work for testing the metrics response, which I feel should be test-case specific...

```
TestUtil.testMetrics(httpClient, null).expect(200, Consumer<Response> consumer);
```

```
TestUtil.testPrometheusMetrics(httpClient, null).expect(200, Consumer<Response> consumer);
```

```
TestUtil.testOpenMetrics(httpClient, null).expect(200, Consumer<Response> consumer);
```

... and make the test class implement `Consumer<Response>`

---

Thoughts?

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
dhoard: This has been primarily addressed in AntuBLUE Test Engine v3.0.0, which a minor issue that appears to be an IntelliJ bug.

https://youtrack.jetbrains.com/issue/IDEA-317561/IntelliJ-test-method-selections-fails-for-hierarchical-test-in-test-output-window

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
dhoard: Using `TestUtil` as more of a "builder" it would be something like...

```
    @TestEngine.Test
    public void testHealthyWithUsernameWithPassword() throws Exception {
        TestUtil
                .testHealthy(httpClient)
                .withCredentials(new BasicAuthenticationCredentials(USERNAME, PASSWORD))
                .expect(200);
    }

    @TestEngine.Test
    public void testHealthyWithUsernameWithoutPassword() throws Exception {
        TestUtil
                .testHealthy(httpClient)
                .withCredentials(new BasicAuthenticationCredentials(USERNAME, null))
                .expect(401);
    }
```

This feels cleaner to me.

dhoard: @fstab I refactored `io.prometheus.jmx.test.javaagent.general.Minimal_IT` and `io.prometheus.jmx.test.javaagent.general.WhiteListObjectNames_IT` using ideas discussed above.

Take a look.

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
```diff
@@ -0,0 +1,141 @@
...
+    public static void beforeClass() {
+        network = createNetwork();
+    }
+
+    @TestEngine.BeforeAll
+    public void beforeAll() {
+        applicationContainer = createApplicationContainer(this, dockerImageName, network);
+        applicationContainer.start();
+
+        exporterContainer = createExporterContainer(this, dockerImageName, network);
```
fstab, line 74: You put a lot of work into reducing copy-and-paste, thanks a lot!

I'm wondering if we can reduce copy-and-paste even more. Most tests are duplicated: One for the `agent` mode, one for the `standalone` mode. We might be able to write just a single test, and add the mode as a parameter.

Antublue supports `ParameterMap` as a wrapper for multiple parameters. So instead of having a `Parameter` with just the Docker image name, we could use a `ParameterMap` containing the image name and the mode. For example, we could add something like the following to `DockerImageNameParameters.parameters()`:

```java
        for (String mode : new String[]{"standalone", "agent"}) {
            for (String dockerImageName : dockerImageNames) {
                Parameter parameter = ParameterMap
                        .named(dockerImageName + " " + mode)
                        .put("image", dockerImageName)
                        .put("mode", mode)
                        .parameter();
                if (predicate.test(parameter)) {
                    parameters.add(parameter);
                }
            }
        }
```

Then the `parameter()` method in the tests would look something like this:

```java
    @TestEngine.Parameter
    public void parameter(Parameter parameter) {
        ParameterMap parameterMap = parameter.value(ParameterMap.class);
        dockerImageName = parameterMap.get("image");
        mode = parameterMap.get("mode");
    }
```

and in the tests initialization code we could do something like this (the initialization code can be implemented in a common utility class or base class because it will be the same for most tests):

```java
        if (mode.equals("standalone")) {
            // run the standalone http server
            // ...
        } else {
            // run the java agent
            // ...
        }
```

What do you think?

fstab (changes requested):
Hi Doug, thanks a lot for your work, it's already looking great. I left a few comments, let me know what you think.

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/Resource.java
```diff
@@ -0,0 +1,77 @@
...
+ * limitations under the License.
+ */
+
+package io.prometheus.jmx.test;
+
+import io.prometheus.jmx.test.util.Precondition;
+
+import java.util.regex.Pattern;
+
+public class Resource {
```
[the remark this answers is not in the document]
dhoard: It's not used. I'll remove it.

I also found some other files (jmx-keystore.jks, jmx-truststore.jks, jmxremote.access, jmxremote.password, keystore.jks, keystore.pkcs12, truststore.jks, truststore.pkcs12) that aren't yet being used which I'll remove as well.

integration_test_suite/integration_tests/src/main/java/io/prometheus/jmx/test/DockerImageNameParameters.java
```diff
@@ -0,0 +1,160 @@
+/*
+ * Copyright 2022-2023 Douglas Hoard
```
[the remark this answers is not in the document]
dhoard: Looks good. I'll make the change.

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/httpserver/general/BlackListAndWhiteListObjectNames_IT.java
dhoard: I refactored the tests, consolidating the Java agent and standalone tests. The code is much, much smaller, and cleaner.

The execution time has _very slightly_ increased since this approach decreases parallelism (the test engine executes a test class sequentially, but different test classes are executed in parallel.)

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/Abstract_IT.java
```diff
@@ -0,0 +1,231 @@
...
+     * @return the return value
+     */
+    @TestEngine.ParameterSupplier
+    protected static Stream<Parameter> parameters() {
+        List<Parameter> parameters = new ArrayList<>();
+
+        DockerImageNames
+                .names()
+                .forEach(dockerImageName -> {
+                    for (Mode mode : Mode.VALUES) {
```
fstab, line 52: There's no need for an explicit `VALUES` variable. Enums have a built-in `values()` method, so you could write `Mode.values()`.

fstab (approved):
Thanks a lot Doug, awesome to see the lines of code decreasing so much! So let's go and merge it! Please squash the commits before merging.

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/Abstract_IT.java
dhoard: Removed.

PR #787, opened 2023-04-19
Support Enum values

speedfl:
Hello,

Some metrics are using Enum values and are rejected. It is the case for example of the `kafka.streams:type=stream-metrics` which contains an attribute `state` which can be `RUNNING`, `REBALANCING`...

Links:

- https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java#L234-L240
- https://docs.confluent.io/platform/current/streams/monitoring.html#built-in-metrics

In a Kafka Stream environment, we may want to know to know all streams which are not in a RUNNING state. Therefore it could be useful to do `sum(kafka_streams_state_running)` and compare it with the number of instances to know if all instances are `RUNNING` and processing.

For information I built the javaagent jar and deployed it in our environment and I was able to get answer:

image


dhoard (approved):
Accepted.

@speedfl thanks for the contribution!

PR #803, opened 2023-05-12
Running multiple agents on a single machine

### LINKED ISSUE #627: Multiple JVMs per machine with mutliple Ports
Hi

I am intending to monitor Apache Spark Executor processes with the jmx_exporter. It turns out that in our setup where we configure a certain executor size in terms of CPU and memory and nodes in the cluster can have multiple of those the start of the JVM process with the agent attached failed. This is due to the static config which tries to reuse the same port over and over again on the the physical node (Kubernetes pod in our case). 

Log output when starting minimal docker-compose example:
`worker1         | 21/08/14 12:31:52 INFO ExecutorRunner: Launch command: "/usr/local/openjdk-15/bin/java" "-cp" "/opt/spark/conf/:/opt/spark/jars/*" "-Xmx1024M" "-Dspark.driver.port=43441" "-javaagent:/opt/prometheus/jmx_prometheus_javaagent-0.16.1.jar=12345:/opt/prometheus/config.yaml" "org.apache.spark.executor.CoarseGrainedExecutorBackend" "--driver-url" "spark://CoarseGrainedScheduler@192.168.50.9:43441" "--executor-id" "1" "--hostname" "172.100.100.3" "--cores" "1" "--app-id" "app-20210814123152-0000" "--worker-url" "spark://Worker@172.100.100.3:36191"
worker1         | 21/08/14 12:31:52 INFO ExecutorRunner: Launch command: "/usr/local/openjdk-15/bin/java" "-cp" "/opt/spark/conf/:/opt/spark/jars/*" "-Xmx1024M" "-Dspark.driver.port=43441" "-javaagent:/opt/prometheus/jmx_prometheus_javaagent-0.16.1.jar=12345:/opt/prometheus/config.yaml" "org.apache.spark.executor.CoarseGrainedExecutorBackend" "--driver-url" "spark://CoarseGrainedScheduler@192.168.50.9:43441" "--executor-id" "0" "--hostname" "172.100.100.3" "--cores" "1" "--app-id" "app-20210814123152-0000" "--worker-url" "spark://Worker@172.100.100.3:36191"
worker1         | 21/08/14 12:31:53 INFO Worker: Executor app-20210814123152-0000/0 finished with state EXITED message Command exited with code 134 exitStatus 134
`

From reading issues like #328 and others it sounds like such an approach is not intended or at least not supported at the time. Furthermore I understand the difficulty of dynamically scraping potentially random ports. 

So I am primarily asking for advice of how to approach this or if there is some easy setup I am missing?

dupetr:
## Why this was created
We needed a way to collect JMX metrics from Apache Spark jobs running on AWS EMR.

The way you tell Spark to expose these metrics is
```
spark submit ...
--conf "spark.executor.extraJavaOptions=-javaagent:./jmx_prometheus_javaagent-0.17.2.jar=<1 specific port>:<config file>"
```
This works fine as long as you have 1 JVM with 1 agent exposing the metrics as it will open the webserver on a specific port.
When you have 2 executors on 1 machine, one will not start (you cannot start 2 webservers on 1 port) and the job will crash.

## The enhancement

1. You can define a range of ports, in our case we use 39100-39115. Some of the ports are wasted as a headroom. Nevertheless, it needs a continuous block/range of ports.
2. It will not fail, when it cannot start a webserver. It will try to pick next port in the range. It would fail if your port range have 5 ports, but you would want to start 10 exporters on that machine.

The exporters will compete for resource (port) and they act independently. For this there are backoff times and retries.

On driver/executor stdout it looks like this
```
Backing off at server start for 1218ms
Looking up free port. Checking: 39100, remaining ports in range: 16
Port 39100 is used. Trying next one.
Looking up free port. Checking: 39101, remaining ports in range: 15
Found free port 39101
Trying to start JMX agent on 39101
Started JMX agent on 39101. (retries left: 16)
```
fixes #627


dupetr:
Closing in favour of #805. Git got better of me.

PR #858, opened 2023-07-27
Refactored logging for performance

dhoard:


pom.xml
```diff
@@ -24,7 +24,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <prometheus.simpleclient.version>0.16.0</prometheus.simpleclient.version>
+        <prometheus.simpleclient.version>0.16.1</prometheus.simpleclient.version>
```
fstab, line 27: Latest release is 0.16.0.

collector/src/main/java/io/prometheus/jmx/logger/LoggerFactory.java
```diff
@@ -0,0 +1,42 @@
...
+package io.prometheus.jmx.logger;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/** Class to implement a LoggerFactory */
+public class LoggerFactory {
+
+    private static Map<Class<?>, Logger> LOGGER_MAP = Collections.synchronizedMap(new HashMap<>());
```
fstab, line 26: This map is not needed, because you create only one `static final` instance per class anyway.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
```diff
@@ -245,13 +242,13 @@ private void processBeanValue(
...
         if (value == null) {
-            logScrape(domain + beanProperties + attrName, "null");
+            LOGGER.log(Level.FINE, "%s%s%s scrape: null", domain, beanProperties, attrName);
         } else if (value instanceof Number || value instanceof String || value instanceof Boolean || value instanceof java.util.Date) {
             if (value instanceof java.util.Date) {
                 attrType = "java.lang.Double";
                 value = ((java.util.Date) value).getTime() / 1000.0;
             }
-            logScrape(domain + beanProperties + attrName, value.toString());
+            LOGGER.log(Level.FINE, "%s%s%s scrape: %s", domain, beanProperties, attrName, value);
```
fstab, line 251: I think a static import would make it more readable, like `LOGGER.log(FINE, ...)`.

fstab (changes requested):

pom.xml
dhoard: Fixed.

collector/src/main/java/io/prometheus/jmx/logger/LoggerFactory.java
dhoard: Removed Map.

collector/src/main/java/io/prometheus/jmx/JmxScraper.java
dhoard: Changed to static import.

PR #870, opened 2023-08-05
Added MBean attribute exclusion filtering

dhoard:


collector/src/main/java/io/prometheus/jmx/ObjectNameAttributeFilter.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectName;
+
+/** Class to implement an ObjectNameAttributeFilter */
```
fstab, line 28: Nitpick: This class has JavaDoc comments that don't add any info that's not already in the class name / method name. I think we should not add comments like this. A comment is great if it has additional info, but if it just repeats the class name or method signature I think it's better to have no comment at all.

collector/src/main/java/io/prometheus/jmx/ObjectNameAttributeFilter.java
```diff
@@ -0,0 +1,133 @@
...
+            for (Map.Entry<Object, Object> entry : objectNameAttributeMap.entrySet()) {
+                ObjectName objectName = new ObjectName((String) entry.getKey());
+
+                List<String> attributeNames = (List<String>) entry.getValue();
+
+                Set<String> attributeNameSet =
+                        excludeObjectNameAttributesMap.computeIfAbsent(
+                                objectName, o -> Collections.synchronizedSet(new HashSet<>()));
+
+                for (String attribueName : attributeNames) {
```
fstab, line 68: Typo in `attribueName`, but also this loop can be replaced with `attributeNameSet.addAll(attributeNames)`.

fstab (approved):
Nice, thanks a lot Doug. I added a few minor comments. They are optional, feel free to ignore my comments and merge :)

collector/src/main/java/io/prometheus/jmx/ObjectNameAttributeFilter.java
```diff
@@ -0,0 +1,133 @@
...
+    public static final String EXCLUDE_OBJECT_NAME_ATTRIBUTES_DYNAMIC =
+            EXCLUDE_OBJECT_NAME_ATTRIBUTES + "Dynamic";
+
+    private final Map<ObjectName, Set<String>> excludeObjectNameAttributesMap;
+
+    private boolean dynamicExclusion;
+
+    /** Constructor */
+    private ObjectNameAttributeFilter() {
+        excludeObjectNameAttributesMap = Collections.synchronizedMap(new HashMap<>());
```
fstab, line 43: It might be more efficient to use `new ConcurrentHashMap<String, Set<String>>()` with `Collections.newSetFromMap(new ConcurrentHashMap<String, Boolean>());` as values, because then you won't have to access a `synchronized` data structure in `exclude()`.

collector/src/main/java/io/prometheus/jmx/ObjectNameAttributeFilter.java
dhoard: I agree.

With the addition of Spotless + Google Java Format, Javadoc build warnings occur if public / non-overridden methods don't have proper Javadoc.

I'll improve the Javadoc for these changes; improving other Javadoc as an ongoing interactive process.

collector/src/main/java/io/prometheus/jmx/ObjectNameAttributeFilter.java
dhoard: Great idea! I'll make the change.

collector/src/main/java/io/prometheus/jmx/ObjectNameAttributeFilter.java
dhoard: Good catch. I'll fix it.

PR #873, opened 2023-08-16
Fix Exceptions are not handled on premain

guhanjie:
Prev code would cause jvm to crash as Exception uncatched in javaagent premain. 
So this commit make the BindException checked and output definite error and then exit expectedly.

### DIFF
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -23,6 +23,7 @@ import io.prometheus.jmx.common.http.ConfigurationException;
 import io.prometheus.jmx.common.http.HTTPServerFactory;
 import java.io.File;
 import java.lang.instrument.Instrumentation;
+import java.net.BindException;
 import java.net.InetSocketAddress;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -62,7 +63,10 @@ public class JavaAgent {
                                     CollectorRegistry.defaultRegistry,
                                     true,
                                     new File(config.file));
-        } catch (ConfigurationException e) {
+        } catch (BindException e) {
+            System.err.println("Jmx-exporter listen port bind failed : " + e.getMessage());
+            System.exit(1);
+        }  catch (ConfigurationException e) {
             System.err.println("Configuration Exception : " + e.getMessage());
             System.exit(1);
         } catch (IllegalArgumentException e) {
--- a/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
+++ b/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
@@ -23,7 +23,6 @@ import io.prometheus.jmx.common.http.ConfigurationException;
 import io.prometheus.jmx.common.http.HTTPServerFactory;
 import java.io.File;
 import java.lang.instrument.Instrumentation;
-import java.net.BindException;
 import java.net.InetSocketAddress;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


guhanjie:
@dhoard could you please have a review?

dhoard:
@guhanjie thanks for the PR! The change looks good.

Since it appears other types of Exceptions are not handled, maybe we should have a generic catch block to catch all possible exceptions, and change the `IllegalArgumentException` when parsing configuration to a ConfigurationException`.

Thoughts?

```
try {
    // DO STUFF
} catch (Throwable t) {
    synchronized (System.err) {
        System.err.println("Failed to start Prometheus JMX Exporter");
        System.err.println();
        t.printStackTrace();
        System.err.println();
        System.err.println("Prometheus JMX Exporter exiting");
    }
    System.exit(1);
}
```

guhanjie:
> @guhanjie thanks for the PR! The change looks good.
> 
> Since it appears other types of Exceptions are not handled, maybe we should have a generic catch block to catch all possible exceptions, and change the `IllegalArgumentException` when parsing configuration to a `ConfigurationException`.
> 
> Thoughts?
> 
> ```
> try {
>     // DO STUFF
> } catch (Throwable t) {
>     synchronized (System.err) {
>         System.err.println("Failed to start Prometheus JMX Exporter");
>         System.err.println();
>         t.printStackTrace();
>         System.err.println();
>         System.err.println("Prometheus JMX Exporter exiting");
>         System.err.flush();
>     }
>     System.exit(1);
> }
> ```

It sounds good, I update it, please hava a look again.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -63,13 +63,19 @@ public static void premain(String agentArgument, Instrumentation instrumentation
                                     true,
                                     new File(config.file));
         } catch (ConfigurationException e) {
+            System.err.println(
+                    "Usage: -javaagent:/path/to/JavaAgent.jar=[host:]<port>:<yaml configuration file> ");
             System.err.println("Configuration Exception : " + e.getMessage());
             System.exit(1);
```
dhoard, line 69: Let's remove this altogether since underlying code could cause a `ConfigurationException` unrelated to usage (i.e. something wrong with the contents of the exporter YAML file.)

dhoard (changes requested):

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
dhoard, line 69: Let's remove this catch code altogether since the underlying code could cause a `ConfigurationException` unrelated to usage (i.e. something wrong with the contents of the exporter YAML file.)

dhoard (changes requested):

dhoard (approved):

dhoard:
@guhanjie Can you elaborate on how you are suffering from this issue?

These exceptions should only happen on initial setup/configuration.

guhanjie:
> @guhanjie Can you elaborate on how you are suffering from this issue?
> 
> These exceptions should only happen on initial setup/configuration.

We are deploying our service using jmx_exporter java agent, which port may be used on some busy hosts, this would happen on large cluster, incident but crashed and have a heavy core dump file(20-30GB). That's not we expected.

dhoard:
> We are deploying our service using jmx_exporter java agent, which port may be used on some busy hosts, this would happen on large cluster, incident but crashed and have a heavy core dump file(20-30GB). That's not we expected.

You should define an application deployment standard (reserved port) for applications using the exporter.

I don't feel this change is worthy of a point release to work around  what I feel is an anti-pattern for application deployment.

PR #852, opened 2023-07-22
Refactored outdated terms

dhoard:
Changed code and documentation to use `allowlistObjectNames` and `denylistObjectNames` in lieu of `whitelistObjectNames` and `blacklistObjectNames`

PR #882, opened 2023-10-23
Align example-configs to naming changes of #852

s-rwe:
A minor PR to have the example_configs make use of the updating naming, of #852

@dhoard for consideration

### DIFF
--- a/example_configs/activemq.yml
+++ b/example_configs/activemq.yml
@@ -1,8 +1,8 @@
 lowercaseOutputName: true
 lowercaseOutputLabelNames: true
-blacklistObjectNames:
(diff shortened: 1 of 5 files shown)


dhoard:
@s-rwe Thanks for the PR!

I think this is a worthwhile change, but I'm concerned that a user will use an old JMX Exporter version with the new configuration files, which doesn't work.

I am working on integrating the new `client_java` 1.0.0 library, which could introduce some breaking changes around metrics names. If this is the case, I feel we should completely remove support for the configuration values `whitelistObjectNames` and `blacklistObjectNames`.

PR #886, opened 2023-11-13
Port to client java 1.0.0

dhoard:
Ported JMX Exporter code and integration tests to use `client_java` 1.1.0.

Refactored integration tests for testing both text-based metrics (Prometheus and OpenMetrics) as well as Protobuf format metrics (counter, gauge, and untyped (unknown)).


collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -731,25 +745,94 @@ public void recordBean(
...
+            switch (matchedRule.type) {
+                case "COUNTER":
+                    {
+                        CounterSnapshot.Builder counterBuilder =
+                                countersMap.computeIfAbsent(
+                                        matchedRule.name,
+                                        name ->
+                                                CounterSnapshot.builder()
+                                                        .name(
+                                                                PrometheusNaming.sanitizeMetricName(
```
fstab, line 760: The `sanitizeMetricName()` calls here are not needed because there is a previous call to `JmxCollector.safeName()`. However, we might refactor `JmxCollector.safeName()` to something like this:
```java
    static String safeName(String name) {
        if (name == null) {
            return null;
        }
        String sanitizedName = PrometheusNaming.sanitizeMetricName(name);
        while (sanitizedName.contains("__")) {
            sanitizedName = sanitizedName.replace("__", "_");
        }
        return sanitizedName;
    }
```

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -775,83 +860,37 @@ public List<MetricFamilySamples> collect() {
...
+
+        for (UnknownSnapshot.Builder unknown : receiver.unknownMap.values()) {
+            result.metricSnapshot(unknown.build());
         }
+
+        return result.build();
+    }
+
+    @Override
+    // TODO remove once bug is fixed in client_java
```
fstab, line 888: Can this be removed? I'm not aware of a bug.

fstab (changes requested):
Thanks a lot @dhoard and sorry for taking so long. It's looking great, just a few minor comments of things that might be removed, but apart from that it's looking good.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
```diff
@@ -403,10 +449,6 @@ private SampleKey(String name, List<String> labelNames, List<String> labelValues
             this.labelValues = labelValues;
         }
 
-        private static SampleKey of(MetricFamilySamples.Sample sample) {
```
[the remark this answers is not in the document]
dhoard: I agree. Removing the code.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
dhoard: Removing the method - this was resolved in 1.1.0 with...

[ENHANCEMENT] Avoid duplicate call to the scrape() method in the default Collector implementation. This does not affect the core metrics, because the core metrics override getPrometheusName(), so they don't use the default implementation.

collector/src/main/java/io/prometheus/jmx/JmxCollector.java
dhoard: In the process of testing this change and determined that `PrometheusNaming.sanitizeMetricName` doesn't handle two scenarios defined in `SafeNameTest`

An empty string:

- `JmxCollector.safeName()` returns an empty string
- `PrometheusNaming.sanitizeMetricName` throws an `IllegalArgumentException`

I believe throwing an `IllegalArgumentException` is the correct behavior. Please confirm.

A metric prefixed with a number:

- `JmxCollector.safeName()` returns a String prefixed with `_`
- `PrometheusNaming.sanitizeMetricName` returns the string unmodified

Can you confirm the correct behavior?

fstab: * I agree throwing a an `IllegalArgumentException` is the right behavior if the metric name is empty.
* If `sanitizeMetricName()` returns metric names starting with a digit that's a bug, I'll fix it in `client_java`.

PR #917, opened 2024-02-22
Improve performance of JmxScraper

nicktelford:
`JmxScraper#scrapeBean()` uses `ObjectName#toString()` twice per-MBean
_attribute_ to check the MBean name against several special-cases.

`ObjectName#toString()` is quite expensive, because it allocates a new
`char[]` and copies the name data into it, on each call.

Consequently, when there are a large number of number of MBean attributes,
these calls compound and can create performance problems.

Caching the result of `ObjectName#toString()` outside this loop ensures
we only call it once per-MBean.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxScraper.java
@@ -212,13 +212,15 @@ class JmxScraper {
             return;
         }
 
+        final String mBeanNameString = mBeanName.toString();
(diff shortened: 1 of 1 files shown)


nicktelford:
See attached flame graph for an illustration of the allocations caused by these `ObjectName#toString()` calls:
flame_graph

dhoard:
@nicktelford Good catch! Thanks for the PR. I'll merge it

PR #918, opened 2024-02-22
Cache MBean Domain for performance

nicktelford:
As with #917, we also call `ObjectName#getDomain()` once
per-_attribute_, which is unnecessary and can be expensive when MBeans
contain a large number of attributes.

This is because `ObjectName#getDomain()` allocates a new String, copying
the domain into it.

Signed-off-by: Nick Telford <nick.telford@gmail.com>


dhoard:
@nicktelford Thanks for the PR! I'll merge it.

PR #940, opened 2024-04-13
Fix documentation links from 0.19 to 0.20

savs:
The links and references to jmx exporter were not updated between 0.19 and 0.20; this patch fixes the documentation.


dhoard:
@savs Thank you for the PR!!!

Issue #941, opened 2024-04-18
Hazelcast Metrics are not showing


Nirzak:
Hi, I am using the following rule to capture hazelcast metrics using JMX.

```
  - pattern: "^com.hazelcast<type=Metrics,instance=(.+),prefix=(.+),tag0=(.+)><>(.+):"
    name: hazelcast_$4
    attrNameSnakeCase: true
    labels:
      instance: $1
      prefix: $2
      tag0: $3
```

But the metrics are not showing on /metrics endpoint. but on the log it's showing that the metrics are scraping successfully from mbean. 

The log:
18-Apr-2024 04:50:14.448 FINE [prometheus-http-1-1] io.prometheus.jmx.logger.Logger.log com.hazelcast:type=Metrics,instance=hz-auth-inst
ance,prefix=executor.internal,tag0="name=hz:cluster:event"_completedTasks process
18-Apr-2024 04:50:14.448 FINE [prometheus-http-1-1] io.prometheus.jmx.logger.Logger.log com.hazelcast{type=Metrics, instance=hz-auth-ins
tance, prefix=executor.internal, tag0="name=hz:cluster:event"}completedTasks scrape: 12
18-Apr-2024 04:50:14.448 FINE [prometheus-http-1-1] io.prometheus.jmx.logger.Logger.log com.hazelcast:type=Metrics,instance=hz-auth-inst
ance,prefix=executor.internal,tag0="name=hz:cluster:event"_queueSize process
18-Apr-2024 04:50:14.448 FINE [prometheus-http-1-1] io.prometheus.jmx.logger.Logger.log com.hazelcast:type=Metrics,instance=hz-auth-inst
ance,prefix=executor.internal,tag0="name=hz:cluster:event"_maximumPoolSize process
18-Apr-2024 04:50:14.448 FINE [prometheus-http-1-1] io.prometheus.jmx.logger.Logger.log com.hazelcast{type=Metrics, instance=hz-auth-ins
tance, prefix=executor.internal, tag0="name=hz:cluster:event"}maximumPoolSize scrape: 1
18-Apr-2024 04:50:14.448 FINE [prometheus-http-1-1] io.prometheus.jmx.logger.Logger.log com.hazelcast:type=Metrics,instance=hz-auth-inst
ance,prefix=executor.internal,tag0="name=hz:cluster:event"_remainingQueueCapacity process
[...]

PR #944, opened 2024-04-21
Added hazelcast example config

Nirzak:
Added example config to fetch custom hazelcast metrics from jmx means.

The metrics will be like as following

hazelcast_total_max_get_latency{instance="hz-instance",prefix="map",tag0="\"name=request-trace-cache\"",} 1.0
hazelcast_priority_queue_size{instance="hz-auth-instance",prefix="operation",} 0.0

It can also handle metrics with multiple tags.

hazelcast_connection_type{instance="hazelcastCacheInstance",prefix="tcp.connection, tag0=\"endpoint=[localhost]:5703\"",tag1="\"bindAddress=[testserver]:5703\"",} 1.0

### DIFF
--- a/example_configs/hazelcast.yml
+++ b/example_configs/hazelcast.yml
@@ -1,0 +1,23 @@
+#---
+attrNameSnakeCase: true
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+whitelistObjectNames:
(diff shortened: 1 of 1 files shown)


Nirzak:
The pull request will close issue #941

dhoard:
@Nirzak Thanks for the PR!!!

Issue #727, opened 2022-08-10
Domain name in labels (weblogic)


dargow:
How can I configure the exporter to get the domain name as a label in every runtime metric?
As I can see, such information is present in JMX objects:
![image](https://user-images.githubusercontent.com/12033844/183878979-41aa9ce6-4c6a-4d33-b0be-876785153876.png)

dhoard:
You should be able use a `pattern` along with `labels` per documentation.

Nirzak:
I also want to know. have used the following pattern regex but didn't work.

```
 - pattern: "^com.bea<Name=(.+),Type=Domain>(.+):"
    name: weblogic_$2
    attrNameSnakeCase: true
    labels:
      domain: $1
```

but it didn't work.

also another thing when we see the jmx mbean we can see the mbean is like as the following,
com.bea:Name=name,Type=Domain

but in jmx config we have to mention it like as below format
com.bea<Name=name,Type=Domain><Metric> 

can someone tell me how can I dump or see all the means like as the above format so that we can easily create custom configs?

Nirzak:
OK I got this working perfectly. the whitelist object will be

`"com.bea:Name=*,Type=Domain"`

then the config will be like as below:

`
- pattern: "^com.bea<Name=(.*), Type=Domain><>(id): (.+)"
    name: weblogic_id
    attrNameSnakeCase: true
    labels:
      domain: $1
`

so basically the metrics will be like as below
weblogic_id{domain="qa05",} 0.0

then we can use prometheus rules to concat this with the other WebLogic metrics.

dhoard:
@Nirzak is it worth adding this to the example Weblogic configuration example (https://github.com/prometheus/jmx_exporter/blob/main/example_configs/weblogic.yml)?

Do you want to create a PR?

Nirzak:
> @Nirzak is it worth adding this to the example Weblogic configuration example (https://github.com/prometheus/jmx_exporter/blob/main/example_configs/weblogic.yml)?
> 
> Do you want to create a PR?

It's your call. It depends on the usage actually. People who want the domain name fetched dynamically will need this metric. If you approve I will create a PR. Thanks.

dhoard:
My thought is to create a PR with a comment and the configuration change commented out. It may help other people in the future. The files are just examples.

PR #946, opened 2024-04-23
Updated weblogic jmx config to fetch domain names as a metric

Nirzak:
Added example config to fetch domain names dynamically in metrics from JMX mbean. 

The example metric will be like as below:

weblogic_id{domain="testdomain",} 0.0

The mbean used to fetch it, is the following one:

com.bea:Name=domainName, Type=Domain


Nirzak:
It will solve the issue #727

Nirzak:
> @Nirzak Thanks for the PR!

Anytime 😀

PR #951, opened 2024-04-26
Add jdk5 jmx ssl toggle

unitsvc:


unitsvc:
@dlespiau when server remove  -Dcom.sun.management.jmxremote.registry.ssl=true , it will work,

```bash
export JDK5_RMI_REGISTRY_SSL_DISABLED=true
```

PR #959, opened 2024-05-23
Added code to deduplicate metrics

dhoard:
Added code to deduplicate metrics.

The `CounterSnapshot.builder()`, `GaugeSnapShot.builder()`, and `UnknownSnapshot.builder()` have no way to get the set of `DatapointSnapshot` list, I had to track the raw metric data (`Label` and value) in lists.

At the end of the collection object, before creating the `MetricsSnapshot` object, the code tracks the `Labels` that have been seen using a set. If the `Labels` exists in the set, a new label pair ( key = `_` + counter + `_`, value = `"id"` is added to make the metric unique. (Reverse mapping to prevent potential metric label name collision)


dhoard:
@fstab please review. The code adds a label to the second, etc. duplicate.

Issue #958, opened 2024-05-17
Collection fails for Kafka using release 1.0.0


dhoard:
Collection fails for Kafka using release 1.0.0. Investigation is underway.

dhoard:
**Background**

As part of the work in `client_java` 1.x.x to integrate with OpenMetrics/OpenTelemetry metric names are enforced.

Kafka _appears_ to create metrics with suffixes that violate the OpenMetrics/OpenTelemetry specification.

**Stacktrace**

```
An Exception occurred while scraping metrics: java.lang.IllegalArgumentException: 'kafka_server_socketservermetrics_failed_handshake_total': Illegal metric name. The metric name must not include the '_total' suffix. Call PrometheusNaming.sanitizeMetricName(name) to avoid this error.
	at io.prometheus.metrics.model.snapshots.MetricMetadata.validate(MetricMetadata.java:106)
	at io.prometheus.metrics.model.snapshots.MetricMetadata.<init>(MetricMetadata.java:63)
	at io.prometheus.metrics.model.snapshots.MetricSnapshot$Builder.buildMetadata(MetricSnapshot.java:80)
	at io.prometheus.metrics.model.snapshots.GaugeSnapshot$Builder.build(GaugeSnapshot.java:129)
	at io.prometheus.jmx.JmxCollector.collect(JmxCollector.java:832)
	at io.prometheus.metrics.model.registry.MultiCollector.collect(MultiCollector.java:26)
	at io.prometheus.metrics.model.registry.PrometheusRegistry.scrape(PrometheusRegistry.java:72)
	at io.prometheus.metrics.exporter.common.PrometheusScrapeHandler.scrape(PrometheusScrapeHandler.java:112)
	at io.prometheus.metrics.exporter.common.PrometheusScrapeHandler.handleRequest(PrometheusScrapeHandler.java:53)
	at io.prometheus.metrics.exporter.httpserver.MetricsHandler.handle(MetricsHandler.java:43)
	at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:95)
	at jdk.httpserver/sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:82)
	at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:98)
	at jdk.httpserver/sun.net.httpserver.ServerImpl$Exchange$LinkHandler.handle(ServerImpl.java:851)
[...]

PR #960, opened 2024-05-26
Issue 958 / Handle name collisions

fstab:
Experimenting with a solution to #958.

@dhoard this is work-in-progress, but I'm wondering what you think of the general approach?

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -26,12 +26,8 @@ import io.prometheus.metrics.core.metrics.Counter;
 import io.prometheus.metrics.core.metrics.Gauge;
 import io.prometheus.metrics.model.registry.MultiCollector;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
-import io.prometheus.metrics.model.snapshots.CounterSnapshot;
-import io.prometheus.metrics.model.snapshots.GaugeSnapshot;
-import io.prometheus.metrics.model.snapshots.Labels;
 import io.prometheus.metrics.model.snapshots.MetricSnapshots;
 import io.prometheus.metrics.model.snapshots.Unit;
-import io.prometheus.metrics.model.snapshots.UnknownSnapshot;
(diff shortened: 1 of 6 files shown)


dhoard:
@fstab my thoughts...

---

I tested the code with a few more MBeans that duplicate the colliding name issue along with an MBean that has `_total` MBean attribute name.

```
public MatchedRule withValue(double value) {
```

... requires using the `PrometheusNaming.sanitizeMetricName()`

```
   public MatchedRule withValue(double value) {
        return new MatchedRule(
                PrometheusNaming.sanitizeMetricName(name),
                matchName,
                type,
                help,
                labelNames,
                labelValues,
                value,
                valueFactor);
    }
```

With this change, all the JUnit tests pass.

----

For the extra label, I prefer `_objectname` versus `unique_mbean_name` since it's shorter and has a more concise meaning. (The leading `_`  can be an indicator that it was added by the exporter.) I also don't believe we need the value in the extra label.

This results in metrics like...

```
# HELP kafka_rest_jersey_metrics_v3_topics_partitions_reassignment_list_request The request count using a cumulative counter kafka.rest:name=null,type=jersey-metrics,attribute=v3.topics.partitions-reassignment.list.request-total
# TYPE kafka_rest_jersey_metrics_v3_topics_partitions_reassignment_list_request untyped
kafka_rest_jersey_metrics_v3_topics_partitions_reassignment_list_request{_objectname="kafka.rest<type=jersey-metrics><>v3.topics-partitions-reassignment.list.request-total"} 0.0
kafka_rest_jersey_metrics_v3_topics_partitions_reassignment_list_request{_objectname="kafka.rest<type=jersey-metrics><>v3.topics.partitions-reassignment.list.request-total"} 0.0
```

I have this code working on a real Kafka server. I don't see any scrape performance degradation.

fstab:
Thanks @dhoard, I pushed an update with your suggestions. Feel free to merge (or modify and merge).

dhoard:
@fstab I have added the new test MBeans. I also changed the code so that the value is not part of the extra `_objectname` label.

Currently testing all test container as well as a Kafka server.

PR #897, opened 2023-12-05
Support ObjectName patterns in excludeObjectNameAttributes

KestasG:
Working on local project encountered performance issues with bean scraping. While analyzing found out that there are lot of unused attributes being tested against rules and taking significant amount of time. It is impossible to list all objects in _excludeObjectNameAttributes_ as there are thousands of them, so decided to add pattern support for it. This allowed to reduce scraping time from 3.9 sec to 1.9 sec.

PR #973, opened 2024-06-13
Fix possible memory leak in ObjectNameAttributeFilter

kw4s:
If:
* autoExcludeObjectNameAttributes is set to true (and it is by default)
* application registers and unregisters a lot of MBeans with unique names
* those MBeans contain attribute types that cannot be exported

Then the ObjectNameAttributeFilter will cause a memory leak where excludeObjectNameAttributesMap will keep adding information about new beans to the exclude map but will never remove it, keeping information about bans no longer registered in MBeanServer:

Screenshot from 2024-06-11 12-09-20

I've added similar approach to what is already implemented in JmxMBeanPropertyCache - that is, once we get a new list of beans that we're interested in, remove any old ones from the cache **and dynamic attribute filter**.
I've also split excludeObjectNameAttributesMap into two maps, one for attributes from config (as we don't want to remove them automatically) and the other one for dynamically added attributes (those should be considered for removal).
This approach will also be relevant when resolving conflicts with #897 (as there the property names from config should be treated as patterns but property names added dynamically should be treated as "literal" - compared with equals).


kw4s:
the minimal reproducible example (after adding necessary imports): https://gist.github.com/kw4s/8b4b3f9e076bec01f2a29f3651ce1d06

dhoard:
@kw4s thanks for the PR!!!

Can you sign your PR (DCO requirement)?

Have you ran the integration tests?

kw4s:
@dhoard 
Hi, I've amended the commit to add sign-off, I've also applied spotless required changes (did not run full build before, and I did not notice there is style enforced, sorry about that) and checked all integration tests, all PASS.

dhoard:
@kw4s I wanted to let you know that I haven't forgotten about the PR. We are on a critical path for OpenTelemetry integration. Once addressed, I'll review your PR.

batk0:
Hello! I've stumbled on the same problem. What's the status of this? Can it be merged in near future?

kw4s:
For us, as a workaround, we just disabled `autoExcludeObjectNameAttributes`

dhoard:
@kw4s Thanks for the PR!!!

PR #981, opened 2024-06-22
OpenTelemetry integration

dhoard:
OpenTelemetry integration.


shivanshuraj1333:
Thanks for this!!

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -58,12 +61,15 @@ public static void premain(String agentArgument, Instrumentation instrumentation
             String host = config.host != null ? config.host : DEFAULT_HOST;
 
             httpServer =
-                    new HTTPServerFactory()
+                    HTTPServerFactory.getInstance()
```
fstab, line 64: Should it be possible to use _only_ OpenTelemetry? I think the current implementation will open a Web server with a Prometheus scrape endpoint in any case, even if a user just wants OpenTelemetry.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -39,6 +41,7 @@ public class JavaAgent {
     private static final String DEFAULT_HOST = "0.0.0.0";
 
     private static HTTPServer httpServer;
+    private static OpenTelemetryExporter openTelemetryExporter;
```
fstab, line 44: These variables are not used, they can be removed.

fstab (changes requested):
Hi Doug, sorry for the delay, I finally found the time for a first review. I focused mostly on the production code. Looking good, I left a few comments.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
dhoard: Correct.

The thought was that if there are issues, someone might want to hit the `/healthy/` or `/metrics` endpoint to check if the server is healthy/capturing metrics.

Also, the usage configuration currently requires the port...

```
java -javaagent:./jmx_prometheus_javaagent-1.0.1.jar=12345:config.yaml -jar yourJar.jar
```

```
java -jar jmx_prometheus_httpserver-1.0.1.jar 12345 config.yaml
```

I can see usage confusion if the port is specified but not being used (or the opposite... port not specified but required.) We don't look at the arguments before loading the configuration YAML.

Does it make sense to have a `jmx_prometheus_javaagent_opentelemetry.jar` or `jmx_opentelemetry_javagent.jar` project/artifact?

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
dhoard: I was unsure of the GC behavior if the objects weren't referenced. I'll remove them.

jmx_prometheus_common/src/main/java/io/prometheus/jmx/common/opentelemetry/OpenTelemetryExporterFactory.java
```diff
@@ -0,0 +1,176 @@
...
+                                                            "Invalid configuration for"
+                                                                    + " /openTelemetry/endpoint"
+                                                                    + " must not be blank")))
+                                    .map(
+                                            new ValidateIsURL(
+                                                    ConfigurationException.supplier(
+                                                            "Invalid configuration for"
+                                                                    + " /openTelemetry/endpoint"
+                                                                    + " must be a URL")))
+                                    .orElse("http://localhost:4317");
```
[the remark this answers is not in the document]
dhoard: We can use the defaults, making the configuration optional, but we need to determine how the configuration YAML should look when someone wants OpenTelemetry with default values.

A "dangling" YAML key will work but looks odd.

```
openTelemetry:
rules:
  - pattern: ".*"
```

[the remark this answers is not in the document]
dhoard: I changed the code to use OpenTelemetry defaults, if not specified in the configuration.

A "dangling" `openTelemetry:` YAML key will enable OpenTelemetry.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
fstab: I think it would be better to stick with one single JAR, having different deployment artifacts is confusing.

What do you think of making the port optional? So if you start with `-javaagent:./jmx_prometheus_javaagent-1.0.1.jar` without a port it will not expose a Prometheus endpoint?

Opening a network port may cause security concerns, and I assume there are users who want to avoid this when they switch to pushing OpenTelemetry.

jmx_prometheus_common/src/main/java/io/prometheus/jmx/common/opentelemetry/OpenTelemetryExporterFactory.java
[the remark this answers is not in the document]
fstab: I think it's ok to have a "dangling" `openTelemetry:` key. If you look at [examples of the OpenTelemetry collector config](https://opentelemetry.io/docs/collector/configuration/) there are lots of "dangling" keys, so OpenTelemetry users are used to this.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
dhoard: Sounds like a good approach. I'll make the change and update the documentation.

jmx_prometheus_common/src/main/java/io/prometheus/jmx/common/opentelemetry/OpenTelemetryExporterFactory.java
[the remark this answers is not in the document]
dhoard: My recent push allows the "dangling" `openTelemetry:` key.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -16,105 +16,78 @@
...
-            httpServer =
-                    new HTTPServerFactory()
-                            .createHTTPServer(
-                                    InetAddress.getByName(host),
-                                    config.port,
-                                    PrometheusRegistry.defaultRegistry,
-                                    new File(config.file));
+            switch (arguments.getMode()) {
+                case HTTP:
+                    {
```
fstab, line 57: The curly brackets `{` `}` are not needed here.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
```diff
@@ -16,105 +16,78 @@
...
+                        OpenTelemetryExporterFactory.getInstance()
+                                .createOpenTelemetryExporter(
+                                        PrometheusRegistry.defaultRegistry, file);
+
+                        break;
+                    }
+                default:
+                    {
+                        throw new RuntimeException("Undefined mode");
```
fstab, line 77: This error message isn't very intuitive. What do you think of printing something like this?
```
"Failed to start the Prometheus JMX exporter: Either specify a port for the Prometheus scrape endpoint, or enable OpenTelemetry mode."
```

Also, maybe we should not print the stack trace here. We could just print the error message to stderr and then exit:

```
System.err.println("...");
System.exit(-1);
```

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/Arguments.java
```diff
@@ -0,0 +1,147 @@
...
+
+        Mode mode;
+        String host = null;
+        Integer port = null;
+        String filename;
+
+        if (matcher.matches()) {
+            switch (matcher.groupCount()) {
+                case 2:
+                    {
```
fstab, line 104: Curly braces `{` `}` are not needed for switch statements.

run_sample_httpserver.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
```
fstab, line 1: Rename the script `httpserver` -> `standalone`.

run_sample_httpserver.sh
```diff
@@ -10,5 +10,5 @@ java \
   -Dcom.sun.management.jmxremote.ssl=false \
   -Dcom.sun.management.jmxremote.authenticate=false \
   -Dcom.sun.management.jmxremote.port=5555 \
-  -jar jmx_prometheus_httpserver/target/jmx_prometheus_httpserver-"${VERSION}".jar 5556 \
+  -jar jmx_prometheus_standalone/target/jmx_prometheus_standalone-"${VERSION}".jar 5556 \
   example_configs/httpserver_sample_config.yml
```
fstab, line 14: Rename `httpserver` -> `standalone`

tools/patch-and-run-integration-test-suite.sh
```diff
@@ -16,7 +16,7 @@ function emit_error () {
 }
 
 JAVA_AGENT_JAR="${1}"
-HTTPSERVER_JAR="${2}"
+STANDALONE_JAR="${2}"
 PROJECT_ROOT_DIRECTORY=$(git rev-parse --show-toplevel)
 
 if [ "$#" -ne 2 ];
```
fstab, line 22: In line 24: rename `httpserver` -> `standalone`.

fstab (changes requested):
Hi Doug, it's looking great!

I got a few minor comments regarding error messages or renames. Functionality is looking great.

run_sample_httpserver.sh
dhoard: Fixed.

tools/patch-and-run-integration-test-suite.sh
dhoard: Fixed

run_sample_httpserver.sh
dhoard: Fixed.

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java
dhoard: The braces are the result of using the `spotless-maven-plugin` and configured Google Java Format for code formatting.

Reference: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

dhoard: Since there are only two modes, I refactored the code to use an `if` / `else`.

The use of the `spotless-maven-plugin` along with Google Java Format requires braces even with a single statement.

https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/Arguments.java
```diff
@@ -0,0 +1,150 @@
...
+                                        agentArgument));
+                    }
+            }
+
+            if (host.trim().isEmpty()) {
+                throw new ConfigurationException(
+                        format("Malformed arguments for Standalone HTTP mode [%s]", agentArgument));
+            }
+        } else {
+            mode = Mode.OPEN_TELEMETRY;
```
fstab, line 144: If I understand correctly, if neither a port for the Prometheus HTTP endpoint nor `opentelemetry:` in the config is specified, `jmx_exporter` will silently do nothing.

Should we print an error and exit in that case?

dhoard:
I am closing PR for new/cleaner PR.

fstab:
:+1:  please ping me when the new PR is ready.

Issue #1001, opened 2024-10-03
[Feature] allow javaAgent to load a custom Authenticator


gtully:
The context is apache activemq artemis, where the mbean server can be locked down to authenticated users. 
The RBAC, requires ArtemisPrincipals, and typically artemis generates these through  JAAS logins. An example authenticator that we would like to plug in is at https://github.com/apache/activemq-artemis/blob/main/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/HttpServerAuthenticator.java

see some documentation at: https://activemq.apache.org/components/artemis/documentation/latest/management.html#artemis_rbac_mbean_server_guard
With this feature, the agent can directly authenticate with Artemis. With out some solution, the agent won't be able to scrape any mbeans!

There is another piece of the puzzle covered by https://github.com/prometheus/client_java/issues/1088
To propagate an authenticated jaas subject to the application or MBean calls we need an attribute and a doAs handler.

dhoard:
@gtully I have integrated the code. I made some changes around class names/configuration.

```yaml
httpServer:
  authentication:
    plugin:
      class: io.prometheus.jmx.AuthenticatorPlugin
      subjectAttributeName: io.prometheus.jmx.CustomAuthenticatorSubjectAttribute
```

gtully:
@dhoard thanks for your help on this, and for the ibm java fix. much appreciated.

PR #1002, opened 2024-10-03
add support for loading a customAuthenticator class #1001 - needs cli…

gtully:
…ent_java 1.4

requires https://github.com/prometheus/client_java/issues/1088

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <prometheus.metrics.version>1.3.1</prometheus.metrics.version>
+        <prometheus.metrics.version>1.4.0-SNAPSHOT</prometheus.metrics.version>
     </properties>
 
     <scm>
--- a/docs/README.md
+++ b/docs/README.md
@@ -251,6 +251,28 @@ httpServer:
 
 ---
 
+### Java agent use of a custom com.sun.net.httpserver.Authenticator class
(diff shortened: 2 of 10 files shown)


gtully:
Is there a branch where the snapshot versions are referenced? I have tweaked the pom.xml (which cannot correct) to have this work. thanks!

dhoard:
The main branch build version is generic `1.x.x` (no `-SNAPSHOT` version.)

pom.xml
```diff
@@ -24,7 +24,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <prometheus.metrics.version>1.3.1</prometheus.metrics.version>
+        <prometheus.metrics.version>1.4.0-SNAPSHOT</prometheus.metrics.version>
```
gtully, line 27: @dhoard thanks, for this change, I guess there is no other option, the client_java project uses snapshots.

dhoard: @gtully to clarify, by design, the `jmx_exporter` project only uses `client_java` release artifacts. There is no snapshot dependency between the projects.

It does become cumbersome when there are dependencies, but it only occurs rarely.

gtully: thanks, that makes sense, this will have to wait for https://github.com/prometheus/client_java/issues/1088

dhoard:
@gtully Thanks for the PR!!! I merged the code/made an update to fix the build.

gtully:
@dhoard thanks for taking the time to review and merge, much appreciated.

PR #1098, opened 2024-12-13
Intern MBean names used in cache

yana-ilieva:
Hello,

I am opening this PR as per the discussion in the Prometheus Developers group.

### Finding
The exporter's caching mechanism can use up a lot of heap memory and can cause frequent GC clean ups, leading to performance degradation of the main application. The exact reason is the MatchedRulesCache object which can grow excessively in size. For each rule with caching enabled, the object keeps all MBean name strings it was matched against. All rules are matched against the same set of JMX MBeans and as a result for each rule we have duplication of the same string values.

### Suggested solution
Use string interning when creating the _matchName_ object, which is used in the cache. This way, we utilize JVM's string pool and reduce the memory usage by String objects with identical content.

### Observed improvement
In my local testing, there is improvement in both memory usage and request speed, as seen in the lower-right corner in the table.
 
image


dhoard:
@yana-ilieva Thanks for the PR!

Issue #834, opened 2023-07-03
HTTPS - Add support to TrustStore


VinceGall:
In Version 0.19.0, HTTPS support has been introduced.  However, it is only possible to configure a keystore and not a truststore.

```
httpServer:
  ssl:
    keyStore:
      filename: localhost.jks
      password: changeit
    certificate:
      alias: localhost
```

Without a truststore I get this error:
```
2023-07-03 | 18:07:02.313 | main | INFO | io.prometheus.jmx.WebServer | Running
Jul 03, 2023 6:07:05 PM io.prometheus.jmx.JmxCollector collect
SEVERE: JMX scrape failed: java.rmi.ConnectIOException: error during JRMP connection establishment; nested exception is:
        javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

I'm able to workaround the problem by configuring the HttpServer with:
```
httpServer:
  ssl:
    certificate:
      alias: localhost
```

And providing the system properties like this:
-Djavax.net.ssl.keyStore=<keystore file> -Djavax.net.ssl.keyStorePassword=<keystore password> -Djavax.net.ssl.trustStore=<truststore file> -Djavax.net.ssl.trustStorePassword=<truststore password>

However this is not ideal because the passwords can be scraped via a simple "ps -ef" command

dhoard:
Please provide your full exporter YAML file content.

dhoard:
Looking more closely at the information you provided above, this is working as expected.

---

HTTPS support only requires a keystore. The keystore for HTTPS support can be configured in two ways:

1. Configuring the keystore in the exporter YAML configuration

```
httpServer:
  ssl:
    keyStore:
      filename: localhost.jks
      password: changeit
    certificate:
      alias: localhost
```

... or ... 

2. Configuring the keystore / keystore password as JVM arguments

```
-Djavax.net.ssl.keyStore= 
-Djavax.net.ssl.keyStorePassword=
```

---

JMX SSL scraping support for the standalone Webserver currently requires JVM system properties:

```
-Djavax.net.ssl.keyStore=
-Djavax.net.ssl.keyStorePassword=
-Djavax.net.ssl.trustStore=
-Djavax.net.ssl.trustStorePassword=
```

This functionality/configuration was not changed 0.19.0 and is unrelated to the HTTPS support that was added.

---

[...]

PR #1118, opened 2025-01-14
add trustStore and needClientAuth config to yaml (#834)

gtully:
@dhoard I am making use of the custom authenticator and want to grant permissions based on the clients certificate so I need to be able to configure needClientAuth on the ssl parameters. Also, in kube land, I need to be able to configure the key and trust store types to support PEM key stores. Having these exposed in the yaml config leaves the system properties free for others.

From what I understand, this additional truststore config will also help with #834


gtully:
peeking at that test failure..

gtully:
if it helps, for some context, the yaml for this use case is produced at: https://github.com/gtully/activemq-artemis-operator/blob/_jmx_exporter_agent/controllers/activemqartemis_reconciler.go#L2096

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/ssl/SSLWithTrustStoreAndClientAuth.java
```diff
@@ -0,0 +1,277 @@
...
+    @Verifyica.Order(1)
+    public void testHealthy(ExporterTestEnvironment exporterTestEnvironment) throws IOException {
+
+        String url = exporterTestEnvironment.getUrl(ExporterPath.HEALTHY);
+
+        try {
+            HttpResponse httpResponse = HttpClient.sendRequest(url);
+            throw new AssertionFailedError("expected exception on no client cert");
+        } catch (Exception expectedOnNoClientCert) {
+        }
```
dhoard, line 137: Change this to use Assertj `assertThatExceptionOfType`

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/ssl/SSLWithTrustStoreAndClientAuth.java
```diff
@@ -0,0 +1,277 @@
...
+            km.init(keyStore, password);
+            TrustManagerFactory tm =
+                    TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            tm.init(keyStore);
+
+            sslContext.init(
+                    km.getKeyManagers(), tm.getTrustManagers(), new java.security.SecureRandom());
+
+            return sslContext;
+        } catch (IOException
```
dhoard, line 116: Why are we catching exceptions and logging them?

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/ssl/SSLWithTrustStoreAndClientAuth.java
```diff
@@ -0,0 +1,277 @@
+/*
+ * Copyright (C) 2023-present The Prometheus jmx_exporter Authors
```
dhoard, line 2: Change the copyright year since this is a new file.

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/ssl/SSLWithTrustStoreAndClientAuth.java
gtully: Thanks, done. much nicer.

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/ssl/SSLWithTrustStoreAndClientAuth.java
gtully: should not be, good catch. thanks.

integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/ssl/SSLWithTrustStoreAndClientAuth.java
gtully: done

gtully:
@dhoard thanks for the review, much appreciated.

dhoard:
@gtully Initial comments...

1. Since this enables mutual TLS, I feel we should change the configuration from `needClientAuth` to `mutualTLS`

```
httpServer:
  ssl:
    mutualTLS: true
    keyStore:
      type: PKCS12
      filename: localhost.pkcs12
      password: changeit
    trustStore:
      type: PKCS12
      filename: localhost.pkcs12
      password: changeit
```

2. We can simplify the logic to get the configuration value by creating a new `ConvertToBoolean` class.

```
package io.prometheus.jmx.common.configuration;

import io.prometheus.jmx.common.util.Precondition;
import java.util.function.Function;
import java.util.function.Supplier;

/** Class to implement ConvertToBoolean */
public class ConvertToBoolean implements Function<Object, Boolean> {

    private final Supplier<? extends RuntimeException> supplier;

    /**
     * Constructor
     *
     * @param supplier supplier
     */
    public ConvertToBoolean(Supplier<? extends RuntimeException> supplier) {
        Precondition.notNull(supplier);
        this.supplier = supplier;
    }

    /**
     * Method to apply a function
     *
     * @param value value
     * @return the return value
     */
    @Override
    public Boolean apply(Object value) {
        if (value == null) {
            throw new IllegalArgumentException();
        }

        try {
            if (value instanceof Boolean) {
                return (Boolean) value;
            } else {
                return Boolean.valueOf(value.toString());
            }
        } catch (Throwable t) {
            throw supplier.get();
        }
    }
}

```

Usage

```
                boolean mutualTLS =
                        rootYamlMapAccessor
                                .get("/httpServer/ssl/mutualTLS")
                                .map(
                                        new ConvertToString(
                                                ConfigurationException.supplier(
                                                        "Invalid configuration for"
                                                                + " /httpServer/ssl/needClientAuth"
                                                                + " must be a boolean")))
                                .map(
                                        new ValidateStringIsNotBlank(
                                                ConfigurationException.supplier(
                                                        "Invalid configuration for"
                                                                + " /httpServer/ssl/mutualTLS"
                                                                + " must not be blank")))
                                .map(
                                        new ConvertToBoolean(
                                                ConfigurationException.supplier(
                                                        "Invalid configuration for"
                                                                + " /httpServer/ssl/mutualTLS"
                                                                + " must be a boolean")))
                                .orElse(false);

if (mutualTLS) {

// ... code to set up the truststore variables, configure the Http server, etc.

```

3. The integration test class `SSLWithTrustStoreAndClientAuth` needs to have both positive and negative tests for all the scenarios for full test coverage

```
@Verifyica.Test
public void testHealthy(ExporterTestEnvironment exporterTestEnvironment) throws IOException
```

```
@Verifyica.Test
public void testDefaultTextMetrics(ExporterTestEnvironment exporterTestEnvironment) throws IOException
```

```
@Verifyica.Test
public void testOpenMetricsTextMetrics(ExporterTestEnvironment exporterTestEnvironment) throws IOException
```

```
@Verifyica.Test
public void testPrometheusTextMetrics(ExporterTestEnvironment exporterTestEnvironment) throws IOException
```

```
@Verifyica.Test
public void testPrometheusProtobufMetrics(ExporterTestEnvironment exporterTestEnvironment) throws IOException
```

gtully:
@dhoard sure, mutualTLS works too and is more high level, thanks.
I incorporated your ConvertToBoolean idea and repeated the with and without client cert pattern on each of the test scenarios. I am not sure of the value because the tls layer is doing the real work. But it is more consistent I guess. Thanks for your time on this.

dhoard:
@gtully looks good! Thanks again for the PR!!!

PR #1120, opened 2025-01-16
Update Cassandra example config

russss:
Cassandra has added a lot more metrics recently. The existing example config results in a scrape time of 15-30 seconds and more than doubles Cassandra's CPU usage in our setup.

This changes the example config to use a whitelist approach with a sample whitelist which takes a much more reasonable 200 ms to scrape.

### DIFF
--- a/examples/cassandra.yml
+++ b/examples/cassandra.yml
@@ -1,13 +1,43 @@
 lowercaseOutputLabelNames: true
 lowercaseOutputName: true
-whitelistObjectNames: ["org.apache.cassandra.metrics:*"]
 
-blacklistObjectNames:
-  # ColumnFamily is an alias for Table metrics
-  - "org.apache.cassandra.metrics:type=ColumnFamily,*"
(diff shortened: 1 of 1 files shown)


dhoard:
@russss Thanks for the PR!!! Given the other Cassandra example is valid, I feel we should add this as another example.

Thoughts?

russss:
Personally, I don't think there's any use for the previous config any more. It's definitely too slow to use with a 15 second prometheus scrape interval, it causes Cassandra to use a lot of CPU, and it generates literally thousands of metrics which will blow up the size of the TSDB.

This will presumably only get worse as Cassandra adds even more metrics, so I think using the blacklist approach is a bad idea in general as you risk Cassandra upgrades flooding your TSDB with new metrics (which is what happened to me).

I think the only reasonable way to use jmx_exporter with Cassandra now is to cherry-pick the metrics you want to use, and I think this is a good starting point as it provides a number of example match patterns.

PR #1124, opened 2025-01-28
Support for extra metrics

karina-calma:
This PR has a dependency on [PR](https://github.com/prometheus/jmx_exporter/pull/996). If this functionality is of interest then that PR will be merged first and then this one.

Some of our mbeans only have string attributes, so we won't have any metrics at all, as only numeric or boolean attributes will generate values. We have implemented a way to define a custom value via the configuration file. Here is an example: we want to generate a new metric, such as "isActive" with the value true (1.0)
 
includeObjectNames: ["io.prometheus.jmx:type=stringValue"]
metricCustomizers:
  - mbeanFilter:
      domain: io.prometheus.jmx
      properties:
        type: stringValue
    extraMetrics:
      - name: isActive
        value: true
        description: This is a boolean value indicating if the scenario is still active or is completed.
 
If someone does not explicitly configure metricCustomizers, they should see no impact to performance. When the feature is used, we will go through all the elements in the metricCustomizers block and perform string comparison on the domain and the (optional) properties with every MBean that is matched by the defined rules. For those metrics that match, we will create a new metric with defined name and value. The description is used in the "HELP" section of the logs and indicates what purpose the metric serves.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -77,6 +77,16 @@ public class JmxCollector implements MultiCollector {
         ArrayList<String> labelValues;
     }
 
+    public static class MetricCustomizer {
+        MBeanFilter mbeanFilter;
+        List<String> attributesAsLabels;
+    }
+
+    public static class MBeanFilter {
+        String domain;
+        Map<String, String> properties;
+    }
+
     private static class Config {
         Integer startDelaySeconds = 0;
         String jmxUrl = "";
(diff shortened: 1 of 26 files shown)


dhoard:
@karina-calma I believe this can already be accomplished using existing functionality.

See https://www.robustperception.io/converting-string-states-to-booleans-with-the-jmx-exporter/

karina-calma:
Hi @dhoard, 

We took a look on this approach, but we have some concerns in term of performance, because this solution involves pattern matching for all the mbeans and all the rules.

We have a significant number of mbeans and this approach will put a load on the server and increase response time.
We proposed an alternative solution that is based on string comparison.

dhoard:
@karina-calma I merged https://github.com/prometheus/jmx_exporter/pull/996 . Can you resolved the merge conflicts for this PR?

karina-calma:
Hi @dhoard, we fixed the conflicts.

PR #1126, opened 2025-01-31
Fixed code to use the Java system properties for keystore and truststo…

dhoard:
…re types if not configured in exporter.yaml


dhoard:
@gtully I found some inconsistency in the keystore/truststore type handling in the existing `HTTPServerFactory` code.

The code didn't use the Java system property (if defined). I made some changes to be more consistent.

Can you take a look?

Thanks!

gtully (commented):
that looks sensible to me.

PR #1163, opened 2025-04-17
Redesign cache for better performance

RafalSumislawski:
This PR improves performance of jmx_exporter when caching is enabled while also making the cache smaller. The improvements come primarily from:
* Instead of having and checking separate cache for each rule, we now have one cache that contains the match from the first matching cacheable rule or an unmatched value if none of the cacheable rules match. This means that the performance and cache size won't degrade as the number of rules increases.
* Avoiding construction and manipulation of string when looking up the cache. These operations are delayed until necessary, and avoided completely if a cache entry is found. This way allocation rate is greatly reduces easing out the pressure on GC.
* Avoiding `String#intern` which while hugely beneficial for decreasing the size of the cache under current design (where one key appears many times in the cache), also has a heavy CPU time cost.

Anecdotally: In my standalone setup collecting kafka-streams metrics with 27 cachable rules these changes result in 50% reduction in CPU consumption.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
+++ b/collector/src/main/java/io/prometheus/jmx/JmxCollector.java
@@ -20,6 +20,7 @@ import static java.lang.String.format;
 import static java.util.logging.Level.FINE;
 import static java.util.logging.Level.SEVERE;
 
+import io.prometheus.jmx.MatchedRulesCache.CacheKey;
 import io.prometheus.jmx.logger.Logger;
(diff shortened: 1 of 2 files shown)


dhoard:
@RafalSumislawski Thanks for the PR!!!

Reviewing the PR... do we incur any locking performance when retrieving cached rules with the `ConcurrentHashMap`  in `MatchedRuleCache`?

RafalSumislawski:
@dhoard Done

dhoard:
@RafalSumislawski Thanks for the PR!

PR #1170, opened 2025-05-01
Cache sanitized name and labels

RafalSumislawski:
Currently the rule matches cache stores unsanitized metric name and labels. That means that sanitization has to be done on each scrape. In my tests that sanitization is responsible for 11% of CPU time (+GC) and 39% of allocations. This cost can be easily avoided by performing the sanitization before caching.

CPU profile (with the relevant parts marked in purple):
image

Allocation profile (with the relevant parts marked in purple):
image


dhoard:
@RafalSumislawski thanks for the PR!

PR #1229, opened 2025-05-28
Integration tests performance tweak

dhoard:
Change MetricsParser used for integration tests to remove use of charAt and cleaned up formatting.


dhoard:
Per https://errorprone.info/bugpattern/LoopOverCharArray, `charAt` is more efficient.

Issue #1206, opened 2025-05-19
Upgrade Java Unit Tests to JUnit5


dhoard:
JUnit tests are using JUnit4. These should be upgraded to JUnit5 and use AssertJ.

dhoard:
Closed by PR https://github.com/prometheus/jmx_exporter/pull/1246

Thanks @slender652 !!!

PR #1246, opened 2025-06-08
Upgraded Java Unit Tests to JUnit5 

slender652:
As per #1206, this PR upgrades JUnit dependency to 5.13.0, rewrites current unit tests to be compatible with the new JUnit version and also replaces JUnit assertions with AssertJ.


dhoard:
@slender652 thanks for the PR!!!

PR #1275, opened 2025-07-09
Optionally recognise *_total as counters

RafalSumislawski:
Among many improvements, the `jmx_exporter` version 1.x brings with itself breaking changes in metrics names. The most significant of those changes is the removal of the `_total` suffixes from metric names that used to have that suffix with JMX exporter 0.x. This suffix is for example common in `kafka-streams` metrics. In order to migrate from 0.x to 1.x, we would either need to embrace the change, which translates to adapting hundreds of dashboards and alerts, or work around this issue by defining separate rules for metrics that end with _total and those that don't. This effectively doubles the number of rules which comes with performance and maintenance cost.

In this PR I'm proposing an opt-in feature that would make the 0.x -> 1.x migration easier. It's simple in principle: By setting `inferCounterTypeFromName=true` a user can tell `jmx_exporter`, that metrics with the `_total` suffix are implicitly counters, which effectively preserves the total suffix (it gets removed, but then it gets added because the metric is a counter). While this doesn't make the 1.x metric names 100% compatible with 0.x, it removes the bulk of the differences making the migration relatively easy.

### DIFF
--- a/collector/src/test/java/io/prometheus/jmx/JmxCollectorTest.java
+++ b/collector/src/test/java/io/prometheus/jmx/JmxCollectorTest.java
@@ -58,6 +58,7 @@ public class JmxCollectorTest {
         Camel.registerBean(mbs);
         CustomValue.registerBean(mbs);
         StringValue.registerBean(mbs);
+        KafkaClient.registerBean(mbs);
     }
 
     @BeforeEach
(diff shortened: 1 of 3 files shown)


Gor027 (approved):

dhoard:
@RafalSumislawski, thanks for the PR!!!

I will discuss with the team.

Issue #1114, opened 2025-01-08
Hot reload - cert renewal


lenin-joseph:
Hi, 

We run jmxexporter as a container with the below command and we would like to know whether hot reload of certs is supported?

java -XX:InitialRAMPercentage=50.000000 -XX:MinRAMPercentage=70.000000 -XX:MaxRAMPercentage=50.000000 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=7071 -jar /opt/jmx_exporter/jmx_prometheus_httpserver-0.20.0.jar 9072 /tmp/https_config.yml

Configs:

httpServer:
  ssl:
    keyStore:
      filename: /tmp/keystore.jks
      password: password
    certificate:
      alias: cmx
hostPort: localhost:7072
ssl: false

bash-4.4$ java -version
openjdk version "11.0.25" 2024-10-15
OpenJDK Runtime Environment (build 11.0.25+9-suse-150000.3.119.1-x8664)
OpenJDK 64-Bit Server VM (build 11.0.25+9-suse-150000.3.119.1-x8664, mixed mode)

dhoard:
@lenin-joseph hot reloading of the certificates used by the HTTP server is not supported.

# Questions

1. Are using the standalone JmxExporter and connecting via RMI?
2. If you are connecting via RMI, are the RMI connections using SSL/TLS?
3. If so, how are you managing the RMI SSL certificate rotation since they are handled by the JVM as command-line properties/arguments?

# Details

Implementation would require some refactoring. We would have to move the configuration loading/reloading to the main Java agent/standalone classes and remove it from `JmxExporter` in the `collector` module.

There are a few people that are using the `JmxCollector` directly. Moving the configuration loading/reloading would require them to implement the logic themselves.

Considering our support statement on using the `collector` module...

https://prometheus.github.io/jmx_exporter/1.1.0/collector/

... I'm not opposed to the change.

Hakky54:
I think it should be easy to add the possiblity of ssl reloading without introducing too much code changes. I see the sslcontext is supplied to the server here https://github.com/prometheus/jmx_exporter/blob/47e0f13244c0cca0d69095ac24943368910cc034/jmx_prometheus_common/src/main/java/io/prometheus/jmx/common/HTTPServerFactory.java#L822-L829

[...]

PR #1277, opened 2025-07-14
Add support for reloading ssl at runtime

Hakky54:
PR related to feature request https://github.com/prometheus/jmx_exporter/issues/1114


Hakky54:
@dhoard I added basic/minimal reloading mechanism. Currently it reloads every hour regardless whether the file has changed. This can be enhanced by checking every n amount of time as interval whether the file has been updated and then reload the  ssl. Next to that I refactored the code a bit. I temporally removed the exclusion of slf4j as I didn't know how to properly resolve the failing tests as the library depends on slf4j.

Wdyt of the initial PR, is this something what would work out for you? Or do you prefer a different kind of solution. So should we polish/adjust this PR or close it, what do you think?

dhoard:
@Hakky54 Thanks for the PR!!!

I will discuss with the team whether we want to go with this approach or whether we should implement a full reload at the agent/standalone level.

Hakky54:
> I will discuss with the team whether we want to go with this approach or whether we should implement a full reload at the agent/standalone level.

Sure, thats fine. Looking forward to the outcome. I have adjusted the implementation by the way. It now checks every hour for files changes and updates only if the files have been modified.

Hakky54:
Hi @dhoard did you had a change to discuss with the team regarding this PR, do you want me to adjust something about it?

dhoard:
@Hakky54 Looking at the library, it appears you have a dependency on slf4j.

Is that correct?

Hakky54:
That is great to hear! I am not sure how to include the package for shadowing. Can you point me into the right direction?

Issue #1302, opened 2025-09-02
NPE if JUL Handler does not have a Formatter


rroesch1:
* JUL Handlers can have a Formatter, but that's optional (i.e. `java.util.logging.Handler#getFormatter` may return `null`)
* #1219 introduced code which tries to access the formatter without proper null-check
* the JVM crashes with a NPE if a JUL Handler **without** formatter is in use (e.g. [SLF4JBridgeHandler](https://github.com/qos-ch/slf4j/blob/master/jul-to-slf4j/src/main/java/org/slf4j/bridge/SLF4JBridgeHandler.java))

```
Exception in thread "main" java.lang.ExceptionInInitializerError
	at io.prometheus.jmx.JavaAgent.<clinit>(JavaAgent.java:39)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unknown Source)
	at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.ensureClassInitialized(Unknown Source)
	at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.newMethodAccessor(Unknown Source)
	at java.base/jdk.internal.reflect.ReflectionFactory.newMethodAccessor(Unknown Source)
	at java.base/java.lang.reflect.Method.acquireMethodAccessor(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(Unknown Source)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(Unknown Source)
Caused by: java.lang.NullPointerException: Cannot invoke "Object.getClass()" because the return value of "java.util.logging.Handler.getFormatter()" is null
	at e1723a08afd7bca35570fd31a7656f59.io.prometheus.jmx.logger.LoggerFactory.<clinit>(LoggerFactory.java:37)
	... 10 more
*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" with message Outstanding error when calling method in invokeJavaAgentMainMethod at src/java.instrument/share/native/libinstrument/JPLISAgent.c line: 627
[...]

PR #1303, opened 2025-09-02
Add null-check for JUL Formatters in Handlers (#1302)

rroesch1:
see: #1302


rroesch1:
@dhoard this PR should fix #1302

Issue #1334, opened 2025-10-10
Support for providing cipher suites and enabled TLS protocols


mardlucca:
https://github.com/prometheus/jmx_exporter/blob/b5eb913bfb4ca5a5784f767ac28f4d8bd13917da/jmx_prometheus_common/src/main/java/io/prometheus/jmx/common/HTTPServerFactory.java#L837

Could you please provide support for providing cipher suites and enabled TLS protocols? Something along the lines of:
```
httpServer:
  ssl:
    keyStore:
      filename: localhost.jks
      password: changeit
    certificate:
      alias: localhost
    enabledProtocols: "TLSv1.2,TLSv1.3"
    cipherSuites: "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_DHE_RSA_WITH_AES_256_GCM_SHA384"
```

For protocols, apparently if you use "TLS" for your SSLContext, java will enable all supported cipher suites. Then you can restrict them using SSL parameters...
(Relevant code here: https://github.com/prometheus/jmx_exporter/blob/main/jmx_prometheus_common/src/main/java/io/prometheus/jmx/common/util/SSLContextFactory.java#L146)

Thank you,

Marcio

dankots:
I have prometheus stats being collected on its own port, exposed just for that purpose. I would like to restrict the ciphers greatly for this connection, since I control both ends. However, I can't restrict the ciphers and protocols of the rest of the app, as they are customer-accessed and should support a slightly larger list of commonly supported secure ciphers.

dhoard:
@Hakky54 given the recent merge of your SSL code libary, is this something easily implemented? Do you have an example?

dhoard:
Closed with PR https://github.com/prometheus/jmx_exporter/pull/1357. Thanks @Hakky54 !!!

PR #1357, opened 2025-11-13
Added support for custom ssl protocols and ciphers

Hakky54:
This PR should resolve the following feature request: https://github.com/prometheus/jmx_exporter/issues/1334

### DIFF
--- a/integration_test_suite/integration_tests/pom.xml
+++ b/integration_test_suite/integration_tests/pom.xml
@@ -86,6 +86,11 @@
             <version>1.28.0</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents.client5</groupId>
+            <artifactId>httpclient5</artifactId>
+            <version>5.5.1</version>
+        </dependency>
+        <dependency>
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
             <version>3.27.6</version>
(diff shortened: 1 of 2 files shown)


Hakky54:
@dhoard I did the initial code changes for pulling in the protocols and ciphers from the property/yaml file. Which should work, but I am not sure whether I need to adjust the schema of the yaml or do you think this will just work without other changes.

I will also check whether I am able to add some tests

dhoard:
@Hakky54 looks like a solid approach. Since they are optional, we need to make sure that missing values don't cause an issue. (Code may be fine.)

The other part is an integration test, which can be added later.

Hakky54:
@dhoard I just added the integration test where I test following scenarios:
- Http client is matching the ssl protocol of the server
- Http client is matching the ssl ciphers of the server
- Http client is not matching the ssl protocol of the server
- Http client is not matching the ssl ciphers of the server

Can you please have another look at the PR?

dhoard:
Since it's for integration tests, and Apache 2.0 licensed, it's not an issue. Whichever approach (Apache HTTP Client 2 or URLConnection) you feel is the cleanest will work.

dhoard (approved):

dhoard:
@Hakky54 Looks good. Thanks for the PR!!!

PR #1443, opened 2026-04-23
Enable custom metrics path for HTTP server via configuration file

jiahuili430:
- Add `configureMetricsPath()` to read custom metrics export path
  from the YAML configuration file.
  e.g.:
  ```yaml
  httpServer:
    metrics:
      path: /custom/metrics/path
  ```

- Update relevant tests to ensure the feature functions correctly.

Related PR: https://github.com/prometheus/client_java/pull/1735

### DIFF
--- a/integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/metrics/path/MetricsPathConfigurationTest.java
+++ b/integration_test_suite/integration_tests/src/test/java/io/prometheus/jmx/test/http/metrics/path/MetricsPathConfigurationTest.java
@@ -1,0 +1,239 @@
+/*
+ * Copyright (C) The Prometheus jmx_exporter Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.prometheus.jmx.test.http.metrics.path;
+
+import static io.prometheus.jmx.test.support.http.HttpResponse.assertHealthyResponse;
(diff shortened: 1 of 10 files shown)


jiahuili430:
Hi @dhoard, our project is using `jmx_prometheus_javaagent` to export metrics, but we need to customize the metrics endpoint. I have tested this via workflows and tests passed. Could you please review it? If everything looks good, I hope we can merge it into a new release. Thank you!

dhoard:
@jiahuili430 I agree this is a worth while feature. A few general notes.

- the paths should be configured in the YAML file
- there needs to be an integration test to set up the configuration and validate it

pgj:
>     * the paths should be configured in the YAML file

The port on which the HTTP server runs can already be configured in the command line.  Why is that an exception from this rule?  When discussing this feature with @jiahuili430, including the metrics path seems to be a plausible design choice, given that the port may be specified there.

dhoard:
@pgj the port as an argument has been around for a while. Ideally it would be in the Yaml file. but changing this requirement would break existing installations.

What is the concern with putting in the Yaml file?

pgj:
There is no actual concern on my side, I just wanted to understand why this is.  Now, based on your response, I assume that the port configuration in the command line is just kept around for backward compatibility, but it is deprecated, and new configuration parameters shall go to the YAML configuration only.

dhoard:
Correct - It's still supported for backwards compatibility.

Marking it deprecated is straight forward, but actually removing support will be much more difficult because of the sheer volume of users.

jiahuili430:
Updated based on the comments and rebased to the `main` branch.
Could you please review again?
Thank you, @dhoard

dhoard:
@jiahuili430 I have convert from Verifyica for integration tests to Paramixel (https://github.com/paramixel/paramixel) once you feel your PR is solid, I'll do the conversion and merge.

I also changed the build to run `./run-quick-test.sh` since the free GitHub runners take too long for the smoke test.

jiahuili430:
Refactored the integration tests into `Paramixel`. The CI is much faster now, 11 min vs 1 hour. Thank you for the improvement!

dhoard:
@jiahuili430 Thanks for PR!!!

pgj:
@dhoard thank you very much for your support.  Is there an ETA on the next release that will include this change?

dhoard:
@pgj I found what appears to be a regression in `client_java` 1.6.1. 

https://github.com/prometheus/client_java/issues/2087

It needs to be investigated / resolved before I can perform a release. `client_java` 1.6.1 will break existing JMX Exporter users.

PR #1464, opened 2026-05-13
ci: pin contents: read on build and manual-build

arpitjain099:
Adds an explicit top-level `permissions: contents: read` block to the two CI workflows still inheriting org defaults. Both only run `./run-quick-test.sh` / `./mvnw clean install` against a Java toolchain checkout.


dhoard:
@arpitjain099 thanks for the PR!. Per process requirements can you sign for DCO compliance?

Issue #1000, opened 2024-09-18
Issue in Integrating jmx_prometheus_javaagent with Jboss


hemnath83:
We are running on Jboss EAP 7.2.0 and Java 1.8.
When i tried to integrate jmx_prometheus_javaagent with Jboss facing in loading class file of logger.

Could not load Logmanager "org.jboss.logmanager.LogManager"
java.lang.ClassNotFoundException: org.jboss.logmanager.LogManager
        at java.net.URLClassLoader.findClass(URLClassLoader.java:382)

However when I comment / remove jmx_prometheus_javaagent-0.17.1.jar then jboss is starting without issue and i checked the logs the class of logmanger is getting loaded properly without jmx agent config.

"JAVA_OPTS=%JAVA_OPTS% -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
"JAVA_OPTS=%JAVA_OPTS% -javaagent:C:\\T24\\Area\\Temenos\\prometheus-2.54.1.windows-amd64\\jmx_prometheus_javaagent-0.17.1.jar=8080:C:\\T24\\Area\\Temenos\\prometheus-2.54.1.windows-amd64\\config.yaml"

dhoard:
@hemnath83 This issue is caused because a JMX exporter uses Java logging but doesn't use the application's classpath.

You will need to modify the application classpath so that the jar(s) required to use `org.jboss.logmanager.LogManager` are available to the Java agent classloader.

Reference: https://github.com/prometheus/jmx_exporter/issues/455

hemnath83:
Hi @dhoard ,

Thank you for your response.
Please note i tried to add to the classpath by explicitly adding -Xbootpath in my JVM args but still it did not work.
Im still getting the same error.

Below are my complete JVM arguments in startup for your reference.

set "JAVA_OPTS=-d64 -Xms6G -Xmx8G -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=1G -Xss1024m"
rem "JAVA_OPTS=%JAVA_OPTS% -verbose:class"
set "JAVA_OPTS=%JAVA_OPTS% -Xbootclasspath/p:$JBOSS_HOME/modules/system/layers/base/org/jboss/logmanager/main/jboss-logmanager-2.1.5.Final-redhat-00001.jar"
[...]

Issue #976, opened 2024-06-17
JMX agent interfering with logging format of application


mapsacosta:
We have started monitoring our application with the JMX exporter. Everything works well in terms of metrics, but we are experiencing a strange side effect. Our application log format gets overridden and we get a different logging format that messes with some of our other monitoring.

Our application has a specific logging format 

```
	System.setProperty("java.util.logging.SimpleFormatter.format", "[%4$-8s %1$tF %1$tT %1$tL] %5$s %6$s%n");
```

which translates to:

```
[INFO     2024-06-15 15:45:09 627] Logging level set to CONFIG
[INFO     2024-06-15 15:45:09 658] Caching nodes
[INFO     2024-06-15 15:45:09 694] DB connections to XYZ servers will be direct
```

However, when we start the application with JMX exporter java agent, our logs become unreadable. The same startup log lines look like: 

```
Jun 17, 2024 2:20:30 PM <redacted classpath>
INFO: Logging level set to CONFIG
Jun 17, 2024 2:20:30 PM <redacted classpath>
INFO: Caching nodes
Jun 17, 2024 2:20:30 PM <redacted classpath>
INFO: DB connections to XYZ servers will be direct
```

Has anyone else encountered this? Is there a way to override jmx exporter logging format to fit our original log structure? Why is the exporter overriding application-specific settings?  

Thanks

dhoard:
@mapsacosta The exporter doesn't set any system properties. The exporter Logger is a wrapper around ` java.util.logging.Logger`

https://github.com/dhoard/jmx_exporter/blob/main/collector/src/main/java/io/prometheus/jmx/logger/Logger.java

I _suspect_ what is happening is that the exporter Logger code is initializing the `java.util.logging.Logger` system with the default format.

Because...

 `System.setProperty("java.util.logging.SimpleFormatter.format", "[%4$-8s %1$tF %1$tT %1$tL] %5$s %6$s%n");`

... is executed in application code, after the initial calls to ` java.util.logging.Logger.getLogger()` the format is most likely not changing the format.

Can you test passing passing the format as JVM argument? i.e. `-Djava.util.logging.SimpleFormatter.format`

zalmane:
Just encountered the same issue.
My solution was to add this CLI flag to JAVA_OPTS:
```
-Dava.util.logging.manager=org.jboss.logmanager.LogManager
```

mapsacosta:
[...]

Issue #455, opened 2019-12-29
Cannot run JMX Exporter in Wildfly 17 running on Java 11


straurob:
Currently we're running Wildfly (version 17.0.1.Final) in a self-created Docker image based on Java 8. For monitoring our application via Prometheus we installed jmx_exporter and configured it as shown below.

Now we'd like to switch to the official Docker image j`boss/wildfly:17.0.1.Final` which however uses Java 11. When using our current configuration the JVM cannot be created because `-Xbootclasspath/p` is not supported anymore.

Just replacing `-Xbootclasspath/p` with -`Xbootclasspath/a` does not work and gives the following error message upon startup:

```
Could not load Logmanager "org.jboss.logmanager.LogManager"
java.lang.ClassNotFoundException: org.jboss.logmanager.LogManager
    at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
    at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
    at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
    at java.logging/java.util.logging.LogManager$1.run(LogManager.java:239)
    at java.logging/java.util.logging.LogManager$1.run(LogManager.java:223)
    at java.base/java.security.AccessController.doPrivileged(Native Method)
    at java.logging/java.util.logging.LogManager.<clinit>(LogManager.java:223)
    at java.logging/java.util.logging.Logger.demandLogger(Logger.java:648)
    at java.logging/java.util.logging.Logger.getLogger(Logger.java:717)
    at java.logging/java.util.logging.Logger.getLogger(Logger.java:701)
    at io.prometheus.jmx.shaded.io.prometheus.jmx.JmxCollector.<clinit>(JmxCollector.java:38)
    at io.prometheus.jmx.shaded.io.prometheus.jmx.JavaAgent.premain(JavaAgent.java:29)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[...]

PR #1528, opened 2026-07-18
Add configurable agent logging backend

### LINKED ISSUE #1127: Java agent causes Java logging (JUL) issues on various platforms
# Issues

The JMX Exporter Java agent's use of standard Java logging (JUL). This causes various issues due to initialization/classloader issues because a Java agent uses a different classloader/classpath than an application.

## Related Issues

JMX Exporter Java agent issues...

- https://github.com/prometheus/jmx_exporter/issues/455
- https://github.com/prometheus/jmx_exporter/issues/976
- https://github.com/prometheus/jmx_exporter/issues/1000

... and GlashFish issues...

- https://github.com/eclipse-ee4j/glassfish/issues/25133
- https://github.com/eclipse-ee4j/glassfish/pull/25183

## Investigate using a native logging implementation

The JMX Exporter already has a native logging implementation that is used for development/developer debugging.

https://github.com/prometheus/jmx_exporter/tree/main/collector/src/main/java/io/prometheus/jmx/logger

## Pros

Migrating all code using standard Java logging (JUL) to the native implementation will resolve these issues since the code is packaged in the Java agent.

This makes integration into applications that use a custom JUL `LogManager`/logging implementation a non-issue.

## Cons

The native logging solution, which is currently only used for development/developer debugging, is very limited.

- all or nothing
- only logs to `System.out`
- doesn't log to a file
- doesn't have per class configuration
- only a single logging level
- is not dynamic

In most scenarios, JMX Exporter logging is typically only used when there are issues with the exporter. We may be able to solve some of the issue.

arnabnandy7:
## Summary

Adds configurable logging backends for the JMX Exporter Java agents to prevent exporter logging from prematurely initializing or modifying the application’s Java Util Logging (JUL) configuration.

The Java agent and isolator Java agent now use a JUL-independent native backend by default. Users who rely on the existing JUL integration can explicitly select it through a system property or environment variable.

Fixes #1127.

## Background

The Java agent previously initialized JUL while its classes were loaded during JVM startup. This could happen before the target application or application server had configured its logging system.

Because JUL’s `LogManager` is global to the JVM, early initialization could interfere with applications that use custom logging managers, including WildFly, JBoss EAP, and GlassFish. Reported symptoms included:

- Failure to load a custom `LogManager` because its classes were unavailable to the agent classloader.
- Application log formatting being initialized before the application configured it.
- Exporter code replacing formatters on shared JUL handlers.
- Application-server-specific classpath and startup workarounds being required.

Related reports include #455, #976, and #1000.

### Configurable logging backends

Introduces a logging backend abstraction behind the existing `Logger` and `LoggerFactory` API.

Two backends are available:

- `native`: writes exporter log messages directly to standard error without obtaining a JUL logger.
- `jul`: retains the existing Java Util Logging integration and formatter behavior.

Existing exporter classes continue to use the same logging facade:

```
private static final Logger LOGGER = LoggerFactory.getLogger(Example.class);
```

This keeps backend selection out of collector and exporter business logic.

### Safe Java agent defaults

The following artifacts select `native` as their default before creating their first logger:

- `jmx_prometheus_javaagent`
- `jmx_prometheus_isolator_javaagent`

This prevents exporter-owned logger initialization from accessing JUL during normal Java agent startup.

The collector retains `jul` as its general default, preserving existing behavior when it is used outside these agent entry points, including standalone operation.

### Configuration

The logging backend can be selected with the system property:

```
-Djmx.prometheus.exporter.logging.backend=native
```

or:

```
-Djmx.prometheus.exporter.logging.backend=jul
```

It can also be selected with the environment variable:

```
JMX_PROMETHEUS_EXPORTER_LOGGING_BACKEND=native
```

Supported values are:

- `native`
- `jul`

The system property takes precedence over the environment variable. Explicit configuration overrides the artifact-specific default.

Invalid values fail with a descriptive error instead of silently falling back to another backend.

### Lazy JUL initialization

JUL-specific behavior is contained in `JulLoggerBackend`. Selecting the native backend does not call:

```
java.util.logging.Logger.getLogger(...)
```

The existing JUL formatter setup is also performed only after JUL has been selected. This retains the existing JUL behavior without initializing the global logging system in native mode.

### HTTP server logging

`HTTPServerFactory` previously created a JUL logger directly, bypassing the project’s logging facade.

It now uses `LoggerFactory`, ensuring that native mode remains effective when HTTP server configuration warnings are logged.

The affected warning messages were converted from JUL `MessageFormat` placeholders to the format-string convention already used by the project’s logger facade.

### Isolator agent

The isolator agent’s embedded logging implementation now supports the same backend configuration and defaults to native logging.

This is necessary because isolating the nested exporter with a separate classloader does not isolate the JVM’s JUL `LogManager`.

### Documentation

The Java agent and isolator agent deployment documentation now covers:

- Default backend behavior
- Native output destination and enabled levels
- System property and environment variable configuration
- Configuration precedence
- When backend selection occurs
- Risks associated with explicitly selecting process-wide JUL during agent startup

Versioned documentation was not changed because this functionality has not yet been released.

## Compatibility

The existing JUL implementation has not been removed.

Users who require integration with application JUL handlers, levels, files, or formatting can select:

```
-Djmx.prometheus.exporter.logging.backend=jul
```

Selecting JUL preserves the existing root and named-logger formatter configuration. It also retains the previous application-server risks because JUL and its `LogManager` remain process-wide.

Native logging currently enables `INFO`, `WARN`, and `ERROR` messages and writes them to standard error. `TRACE` remains disabled unless developer-debug behavior is enabled separately.

## Limitations

This change prevents JMX Exporter’s own logging facade and `HTTPServerFactory` from directly initializing JUL in native mode.

It cannot guarantee that every JDK component or third-party dependency loaded by the agent will never access JUL. JUL remains global JVM state and cannot be isolated through shading or a custom classloader.

The change nevertheless removes the known early initialization and formatter mutation performed directly by JMX Exporter.

## Testing

The following modules were built and tested:

- `collector`
- `jmx_prometheus_common`
- `jmx_prometheus_javaagent`
- `jmx_prometheus_isolator_javaagent`

Validation included:

- Existing JUL logging tests
- JUL level mapping tests
- Native backend output tests
- Explicit JUL backend selection
- Invalid backend validation
- HTTP server tests
- Java agent argument tests
- Isolator agent tests
- Spotless formatting checks
- `git diff --check`
- Docusaurus production documentation build

All selected module tests and documentation builds passed.

### DIFF
--- a/collector/src/main/java/io/prometheus/jmx/logger/JulLoggerBackend.java
+++ b/collector/src/main/java/io/prometheus/jmx/logger/JulLoggerBackend.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) The Prometheus jmx_exporter Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.prometheus.jmx.logger;
+
+import java.util.logging.Formatter;
+import java.util.logging.Handler;
+import java.util.logging.SimpleFormatter;
+
+/** Logging backend that delegates to Java Util Logging (JUL). */
+final class JulLoggerBackend implements LoggerBackend {
+
+    private static final String ROOT_LOGGER = "";
+
+    private final java.util.logging.Logger logger;
+
+    static {
+        configureSimpleFormatters(java.util.logging.Logger.getLogger(ROOT_LOGGER));
+    }
+
+    /**
+     * Constructs a JUL backend for the specified logger name.
+     *
+     * @param loggerName logger name
+     */
+    JulLoggerBackend(String loggerName) {
+        logger = java.util.logging.Logger.getLogger(loggerName);
+        configureSimpleFormatters(logger);
+    }
+
+    private static void configureSimpleFormatters(java.util.logging.Logger logger) {
+        for (Handler handler : logger.getHandlers()) {
+            Formatter formatter = handler.getFormatter();
+            if (formatter != null && formatter.getClass().getName().endsWith(SimpleFormatter.class.getName())) {
+                handler.setFormatter(new LoggerFormatter());
+            }
+        }
+    }
+
+    @Override
+    public boolean isEnabled(Level level) {
+        return logger.isLoggable(toJulLevel(level));
+    }
+
+    @Override
+    public void log(Level level, String message) {
+        logger.log(toJulLevel(level), message);
+    }
+
+    static java.util.logging.Level toJulLevel(Level level) {
+        switch (level) {
+            case TRACE:
+                return java.util.logging.Level.FINEST;
+            case INFO:
+                return java.util.logging.Level.INFO;
+            case WARN:
+                return java.util.logging.Level.WARNING;
+            case ERROR:
+                return java.util.logging.Level.SEVERE;
+            default:
+                throw new IllegalArgumentException("Unknown logging level: " + level);
+        }
+    }
+}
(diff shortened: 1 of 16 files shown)


dhoard:
@arnabnandy7 Thanks for the PR.

In general, the idea seems sound, but the code changes the default behavior which would cause unexpected behavior.

Thoughts on changing the default behavior if the system property is not used?

arnabnandy7:
> Thoughts on changing the default behavior if the system property is not used?

@dhoard thanks for the review. I agree that changing the default could surprise existing users, especially if they rely on JUL for log routing, formatting, or level configuration.

I kept JUL as the default when neither the system property nor the environment variable is set, and make native logging opt-in using `-Djmx.prometheus.exporter.logging.backend=native`.

Updated both agents, the tests, and the documentation accordingly. This should provide a workaround for #1127 without changing the existing behavior.

PR #1529, opened 2026-07-18
fix: check for null when MBeanInfo is incomplete

BZValoche:
If the MBeanInfo is incomplete / invalid, a null pointer exception can occur, aborting the entire loop.

@dhoard 
@fstab

### DIFF
--- a/.github/workflows/approve-workflows.yml
+++ b/.github/workflows/approve-workflows.yml
@@ -15,13 +15,13 @@ jobs:
     if: >-
       github.event.issue.pull_request &&
       github.event.comment.body == '/workflow-approve' &&
-      github.repository_owner == 'prometheus'
+      (github.repository_owner == 'prometheus' || github.repository_owner == 'prometheus-community')
     runs-on: ubuntu-latest
     permissions:
       actions: write
       contents: read
       pull-requests: write
     steps:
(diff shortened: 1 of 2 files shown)


dhoard:
@BZValoche please run ./mvw spotless:apply and push update.

BZValoche:
> @BZValoche please run ./mvw spotless:apply and push update.

Done!

dhoard:
@dhoard please sign for DCO and push. Thank you.

BZValoche:
Should be ok now.

dhoard (approved):

dhoard:
@BZValoche Thanks for the PR!

BZValoche:
Thanks for the merge.
I just discussed with someone who has access to the faulty mbean source code. They simply ignore the `attributes` parameter of `getAttributes(...)` and return all the attributes every time.
During the first call to `getMbeanInfo()` a certain attribute is included, and the subsequent calls do not include it. But `getAttributes(...)` still returns this attribute, hence the NPE.

BZValoche:
>During the first call to `getMbeanInfo()` a certain attribute is included, and the subsequent calls do not include it. But `getAttributes(...)` still returns this attribute, hence the NPE.

In fact, this attribute is excluded automatically (because it is a `List<String>`), that's why the error happens.

